PR #13, opened 2023-05-12
Update Prompted Generator - Add Box Prompts

anwai98:
- The prompt generator now gives box prompts on top of point prompts too!

Question:
(There are two ways I thought of approaching the prompt generator class, I'll mention both below)
- One could be two separate classes (one for `PointPromptGenerator` and another for `PointAndBoxPromptGenerator`) (the differences are super slim, but might be an open utility for all purposes)
- Other could be (current PR) one class (`PointAndBoxPromptGenerator`) which returns box prompts subjected to `get_box_prompts` argument else stays consistent to just prompts (returning "None" for boxes - see [here](https://github.com/computational-cell-analytics/micro-sam/commit/3abcbd04e95aac7eee3c7c2ad977763c77824993#diff-6fe0fbcb1f8365acc6bd99b3d6125aa2ab6a1eafb20786a2b40aa1eabf4a2d90R80-R83))

Which one do you recommend?

### DIFF
--- a/micro_sam/prompt_generators.py
+++ b/micro_sam/prompt_generators.py
@@ -4,12 +4,15 @@ from scipy.ndimage import binary_dilation
 
 class PointAndBoxPromptGenerator:
     def __init__(self, n_positive_points, n_negative_points, dilation_strength,
-                 get_point_prompts=False, get_box_prompts=False):
+                 get_point_prompts=True, get_box_prompts=False):
         self.n_positive_points = n_positive_points
         self.n_negative_points = n_negative_points
         self.dilation_strength = dilation_strength
         self.get_box_prompts = get_box_prompts
         self.get_point_prompts = get_point_prompts
+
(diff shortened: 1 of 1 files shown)


constantinpape:
Your current solution is fine. But I would also add `get_point_prompts=True` to the function signature, so we can only get the boxes if needed.

constantinpape:
I have also started to write tests for the point prompts in #16; we should first fix and merge the tests before working on this PR.

anwai98:
> I have also started to write tests for the point prompts in #16; we should first fix and merge the tests before working on this PR.

Nice, okay. So currently I first updated the prompt generator class to utilise the possibility of all kinds of prompting. Now the accessibility would increase (I'll look at the referred PR as well now). Thanks!

anwai98:
Spotted a bug (small yet vital) in the prompt generator (mentioned [here](https://github.com/computational-cell-analytics/micro-sam/pull/16#issuecomment-1546685341)). It's fixed here now as well.

PR #16, opened 2023-05-13
Add test for the point prompt generator

constantinpape:


constantinpape:
@anwai98 I have added a test for the prompt generator here. It's currently failing and I think I know why. It's a bit to complex to discuss here we can have a look together on Monday.

anwai98:
Okay, I spotted something minor which has been creating a misinterpretation for the background points (in short, I was dilating the negative pixels, which interacts with the "true" cell lines and hence "might" mark negations inside the cell). Fixed it in the PR above. (I made the requisite checks on LiveCELL, will update the test case above as well)

I also observed something: (vigra would be a wise choice for the blob-structures as regionprops doesn't essentially mark the "cell centers" as desired - as discussed the behaviour before too). So I will check it with vigra and the current prompter updates. Thanks.

anwai98:
There were some minor elements in the test (the major ones are mentioned above already). Looks like it's working now.

test/test_prompt_generators.py
```diff
@@ -17,18 +19,22 @@ def test_point_prompt_generator(self):
...
             generator = PointPromptGenerator(n_pos, n_neg, dilation_strength=4)
-            for label_id in label_ids:
-                center, box = centers[label_id], boxes[label_id]
+            for label_id in label_ids[:-1]:  # we dodge the last element due to indexing issues for last val
+                center, box = centers[label_id], boxes[label_id]  # we start @id0 hence first mask prompts are accessed
                 coords, point_labels, _, _ = generator(labels, label_id, center, box)
                 coords_ = (np.array([int(coo[0]) for coo in coords]),
                            np.array([int(coo[1]) for coo in coords]))
-                mask = labels == label_id
+                mask = labels == label_id + 1  # here we start @id0 hence adding 1 (to ignore background)
```
constantinpape, line 37: This is not a  good idea. We need to make sure that the `label_id` indexing matches the actual label ids.

test/test_prompt_generators.py
```diff
@@ -4,6 +4,8 @@
 from skimage.data import binary_blobs
 from skimage.measure import label
 
+COLOR_CYCLE = ["#00FF00", "#FF0000"]
```
constantinpape, line 7: This shouldn't be added here.

test/test_prompt_generators.py
anwai98: Ah, I made this change to adapt to the exact mask's configuration to access the `label_id` (because `label_ids` have 0 {background} included right now, and I access `label_id + 1`'s  mask because that's the first expected mask (cross-checked it w.r.t. the centers and bboxes too)).

test/test_prompt_generators.py
anwai98: Ah okay. I would just pass it directly to the function then rather globally as now. Thanks.

anwai98:
Rechecked the `PointPromptGenerator` class compared to the `box-prompter` branch (this should work as it is with the current test workflow (by default, we would get the points as the test case expects))

test/test_prompt_generators.py
```diff
@@ -0,0 +1,63 @@
...
+        from micro_sam.prompt_generators import PointPromptGenerator
+        from micro_sam.util import get_cell_center_coordinates
+
+        labels = self._get_test_data()
+        label_ids = np.unique(labels)
+
+        mode = "v"  # this gets center points using vigra - regionprops throws error for asymmetrical blobs
+        centers, boxes = get_cell_center_coordinates(labels, mode=mode)
+        if mode == "v":
+            centers.pop(0)   # to avoid the background element
```
constantinpape, line 25: This should be handled in `get_cell_center_coordinates` instead of the test (or other code that uses this).
The format of the output should be the same for all modes, so that it can be used interchangeably.

constantinpape: And we should also set the default mode to `v` since it gives more correct results.

test/test_prompt_generators.py
```diff
@@ -0,0 +1,63 @@
...
+        data = binary_blobs(length=256)
+        labels = label(data)
+        return labels
+
+    def test_point_prompt_generator(self):
+        from micro_sam.prompt_generators import PointPromptGenerator
+        from micro_sam.util import get_cell_center_coordinates
+
+        labels = self._get_test_data()
+        label_ids = np.unique(labels)
```
constantinpape, line 20:
```suggestion
        label_ids = np.unique(labels)[1:]
```

test/test_prompt_generators.py
constantinpape: It's more readable to change the `label_ids` instead. (See my suggestion on top).

test/test_prompt_generators.py
```diff
@@ -0,0 +1,63 @@
...
+
+        mode = "v"  # this gets center points using vigra - regionprops throws error for asymmetrical blobs
+        centers, boxes = get_cell_center_coordinates(labels, mode=mode)
+        if mode == "v":
+            centers.pop(0)   # to avoid the background element
+
+        test_point_pairs = [(1, 0), (1, 1), (2, 4), (3, 9)]
+        for (n_pos, n_neg) in test_point_pairs:
+            generator = PointPromptGenerator(n_pos, n_neg, dilation_strength=4)
+            for label_id in label_ids[:-1]:  # we dodge the last element due to indexing issues for last val
```
constantinpape, line 30:
```suggestion
            for label_id in label_ids:
```

constantinpape (commented):
A few more changes, then I think this is good to go.

test/test_prompt_generators.py
anwai98: Yes, that keeps the functionality alike. Updated it now, thanks!

PR #26, opened 2023-05-23
Minor documentation changes

ajinkya-kulkarni:
Used global variables for file paths and added comments for functions

### DIFF
--- a/development/instance_segmentation.py
+++ b/development/instance_segmentation.py
@@ -1,14 +1,17 @@
-import micro_sam.util as util
 import napari
-
 from elf.io import open_file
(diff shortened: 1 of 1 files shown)


constantinpape (approved):
Thanks! One comment though: the scripts in `development` are for developing new functionality. For example, the `instance_segmentation.py` for trying different automatic instance segmentation approaches (that are not part of the napari apps yet because they don't work quite as well yet).
So these files may change quite fast and the results you get may not be so good (yet).

The examples for running the annotation apps are here: https://github.com/computational-cell-analytics/micro-sam/tree/master/examples

PR #29, opened 2023-05-29
Update Random Prompting for Point Prompt Generator

anwai98:
- Updated the point prompting (either to take deterministic values or list of (range of) values

### DIFF
--- a/micro_sam/prompt_generators.py
+++ b/micro_sam/prompt_generators.py
@@ -22,6 +22,21 @@ class PointAndBoxPromptGenerator:
             center_coordinates: Coordinates for the centroid seed of the cell
             bbox_coordinates: Bounding box coordinates around the cell
         """
+        print(self.n_positive_points, self.n_negative_points)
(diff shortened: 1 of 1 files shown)


micro_sam/prompt_generators.py
```diff
@@ -5,15 +5,24 @@
...
         self.get_point_prompts = get_point_prompts
 
         if self.get_point_prompts is False and self.get_box_prompts is False:
             raise ValueError("You need to request for box/point prompts or both")
 
+        # pass an integer to allow deterministic prompting or list to randomise it
+        if isinstance(n_positive_points, int):
+            self.n_positive_points = n_positive_points
+        else:
+            self.n_positive_points = np.random.randint(n_positive_points[0], n_positive_points[1] + 1)
```
constantinpape, line 19: You should do this in `__call__` and not in `__init__`!
For what we have in mind here we want to draw a random number of points from the range **for each object**.
With your current changes you set this to some fixed number for the whole generator class.

constantinpape (changes requested):

Issue #31, opened 2023-05-31
Difficult to see whether first object is commited properly


GenevieveBuckley:
It is difficult to see whether the first object is commited or not, because the colour does not change. It might be helpful for users if there was an obvious change when the commit happens.

Possible solution: Use the "shuffle colormap" one one of the labels layers, so they won't match. I think that could be done in the setup, using something like `viewer.layers['committed_objects'].new_colormap()`

Context: I shared some first impressions using micro-sam with Constantin. He asked me to write them down here, so we don't forget what was said.

PR #33, opened 2023-05-31
Randomize colourmap to make it easier to see when objects are committed

GenevieveBuckley:
Closes https://github.com/computational-cell-analytics/micro-sam/issues/31

I can't see how to pass a specific random seed value to the new_colormap function, so we really do get a totally new and random color mapping each time.

Because I made this change in three places, maybe this section of the code should be refactored into a new function in micro_sam/sam_annotator/util.py?


constantinpape (approved):

GenevieveBuckley:
Makes sense, I think that's a good plan. Feel free to tag me in the [napari zulip](https://napari.zulipchat.com/login/) discussion (I can't read all the message threads, there's too many, but I should notifications if I'm tagged)

constantinpape:
> Feel free to tag me in the [napari zulip](https://napari.zulipchat.com/login/) discussion (I can't read all the message threads, there's too many, but I should notifications if I'm tagged)

Thanks and will do. (I feel the same about the napari zulip ;))

PR #35, opened 2023-06-02
Update example scripts to include data download

GenevieveBuckley:
Right now it is a little difficult for new users to run the example scripts, because the example data is not included.

This PR updates the scripts so that the example data is automatically downloaded if it doesn't already exist. 

I've used [pooch](https://www.fatiando.org/pooch/latest/) to manage the example data download. Pooch has several nice features, including checking whether the data already exists before downloading, progress bars, checking the MD5 file hash for security, unzipping only the exact files we need from a zip bundle, etc. I like it a lot.

Some points:
* The sam_annotator_2d.py example uses a single tiff file from the [LiveCell dataset](https://sartorius-research.github.io/LIVECell/). I cannot figure out how to download just that file, is this possible? 
    * It is impractical to download the entire 1.3 GB images.zip so we can extract only one image. The LiveCell dataset license is Creative Commons by attribution - I think this means it should be possible to host the single file we need somewhere on the internet. That would be much easier. 
* Pooch doesn't currently allow you to delete the zip archives it downloads (you can, but this means it will automatically re-download them, even if the unpacked files still exist). [Relevant issue here](https://github.com/fatiando/pooch/issues/158)

### DIFF
--- a/examples/sam_annotator_tracking.py
+++ b/examples/sam_annotator_tracking.py
@@ -1,21 +1,46 @@
+import os
+from pathlib import Path
+
 from elf.io import open_file
+import pooch
 from micro_sam.sam_annotator import annotator_tracking
 
 
-# This runs the interactive tracking annotator for data from the cell tracking challenge:
-# It uses the training data for the HeLA dataset. You can download the data from
-# http://data.celltrackingchallenge.net/training-datasets/DIC-C2DH-HeLa.zip
 def track_ctc_data():
-    path = "./data/DIC-C2DH-HeLa/train/01"
-    with open_file(path, mode="r") as f:
+    example_data_directory = "./data"
+    with open_file(fetch_example_data(example_data_directory), mode="r") as f:
         timeseries = f["*.tif"][:50]
     annotator_tracking(timeseries, embedding_path="./embeddings/embeddings-ctc.zarr")
 
 
-def main():
-    # run interactive tracking for data from the cell tracking challenge
-    track_ctc_data()
+def fetch_example_data(save_directory):
+    """Cell tracking challenge dataset DIC-C2DH-HeLa.
+
+    Cell tracking challenge webpage: http://data.celltrackingchallenge.net
+    HeLa cells on a flat glass
+    Dr. G. van Cappellen. Erasmus Medical Center, Rotterdam, The Netherlands
+    Training dataset: http://data.celltrackingchallenge.net/training-datasets/DIC-C2DH-HeLa.zip (37 MB)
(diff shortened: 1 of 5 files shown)


examples/sam_annotator_tracking.py
```diff
@@ -1,21 +1,46 @@
...
 
-# This runs the interactive tracking annotator for data from the cell tracking challenge:
-# It uses the training data for the HeLA dataset. You can download the data from
-# http://data.celltrackingchallenge.net/training-datasets/DIC-C2DH-HeLa.zip
 def track_ctc_data():
-    path = "./data/DIC-C2DH-HeLa/train/01"
-    with open_file(path, mode="r") as f:
+    example_data_directory = "./data"
+    with open_file(fetch_example_data(example_data_directory), mode="r") as f:
         timeseries = f["*.tif"][:50]
```
GenevieveBuckley, line 12: Why is the cell tracking timeseries truncated to 50 files? There are 84 tiff files in the training "01" folder.
Should we:
1. Not unzip the full 84 files, but only unzip the 50 files that will be used in this demo?
2. Or, should we remove the `[:50]` from this line, and use the whole series in this demo?

GenevieveBuckley:
This is ready for review. 

There is still one more example script `sam_annotator_2d.py`, but it is silly to try and download a giant `images.zip` bundle to unzip just one image file. I think that should be discussed, and any work happen in a separate PR.

constantinpape:
Thanks @GenevieveBuckley! I didn't here of `pooch` before and it looks very useful. I will have a closer look at this later, just a few co 

> There is still one more example script `sam_annotator_2d.py`, but it is silly to try and download a giant `images.zip` bundle to unzip just one image file. I think that should be discussed, and any work happen in a separate PR.

Yes, I agree. I think I can host the example images separately; but will need to double check with the data licenses if this is ok.

constantinpape:
We could also think about using `pooch` for downloading the model: https://github.com/computational-cell-analytics/micro-sam/blob/master/micro_sam/util.py#L39-L63
I will think about this when reviewing this more closely. (And should then ofc be done in a separate PR.)

examples/sam_annotator_tracking.py
GenevieveBuckley, line 12:
```suggestion
        timeseries = f["*.tif"]
```

GenevieveBuckley: Great, done!

GenevieveBuckley:
Ok, I think this one is ready now.

I've removed the `[:50]` as discussed, and have made issues for the other two action items that are not part of this PR.

PR #52, opened 2023-06-21
modified image series annotator

koopa31:
As discussed, I changed a lot of things to:
- Load images of different sizes.
- Easily change brush size using up and down arrows to manually change segmented object.
- Addition of a button to manually modify a not-satisfying result.
Currently, RGB images are computed as grayscale, as I didn't manage to make it work. This modification is dedicated to making batch processing smoother.

### DIFF
--- a/examples/image_series_annotator_app.py
+++ b/examples/image_series_annotator_app.py
@@ -16,6 +16,7 @@ from magicgui import magicgui
 from micro_sam.segment_from_prompts import segment_from_points
 from micro_sam.sam_annotator.util import create_prompt_menu, prompt_layer_to_points, toggle_label
 from napari import Viewer
+from qtpy.QtWidgets import QPushButton
 
 
 @magicgui(call_button="Segment Object [S]")
@@ -42,6 +43,10 @@ def decrease_brush_size(v):
         current_size = v.layers["segmented_object"].brush_size
         if current_size > 1:
             v.layers["segmented_object"].brush_size -= 1
+
+
+def activate_segmented_object():
+    v.layers.selection.active = v.layers["segmented_object"]
 
 
 def image_series_annotator(embedding_save_path, output_folder):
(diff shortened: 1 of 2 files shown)


examples/image_series_annotator_app.py
```diff
@@ -23,25 +27,53 @@ def segment_wigdet(v: Viewer):
...
 
     # get the sam predictor and precompute the image embeddings
     PREDICTOR = util.get_sam_model()
-    images = np.stack([imageio.imread(p) for p in image_paths])
-    image_embeddings = util.precompute_image_embeddings(PREDICTOR, images, save_path=embedding_save_path)
-    util.set_precomputed(PREDICTOR, image_embeddings, i=0)
+    image = imread(image_paths[0])
+
+    if len(image.shape) == 3:
+        image_embeddings = util.precompute_image_embeddings(PREDICTOR, rgb2gray(image))
```
constantinpape, line 62: You need to pass the argument `ndim=2`  to the precompute function, then it also works with rgb data.
This tells `precompute_image_embeddings` that the input has 2 spatial dimension and it will interpret the last axis as RGB then.

constantinpape (commented):
Hi @koopa31,
first of all thanks for looking into this.
A few things:
- you need to specifiy `ndim=2` in the `precompute_image_embeddings`, then it will work with RGB (see comment in code)
- I think the idea with the brush mode is good, but currently would break the other annotation tools. It would make sense to have this in a separate contribution.
- I have thought about how to implement the image annotator in a better way, I will open an issue or PR on this and ping you there.

constantinpape:
Hi @koopa31 
I realized how to implement the image annotator in a better and more re-usable way, see #56. So I will go ahead and close this.

Regarding these two contributions:
> - Easily change brush size using up and down arrows to manually change segmented object.
> - Addition of a button to manually modify a not-satisfying result.

I think this makes sense, and I would be happy to see it as a contribution! However, it would be better if you can first create an issue about it where you describe the intended functionality, then I can give you some pointers on where you need to change code for it. (Because the current implementation would break a lot of things.)

PR #61, opened 2023-07-10
Update Prompt Generator - First Point as Any FG Point

anwai98:
Updating the prompt generator to use the 1 point objective as an option to any point in the foreground region instead of the center coordinate


constantinpape (approved):
That's exactly what we discussed!
Will merge once the tests pass.

PR #62, opened 2023-07-10
Tunable instance seg

constantinpape:


constantinpape:
This PR:
- Refactors the file names from `segment_instances` -> instance_segmentation` and `segment_from_prompts` -> `prompt-based-segmentation`.
- Introduces new classes for automatic mask generation: `AutomaticMaskGenerator`, `EmbeddingMaskGenerator`, `TiledEmbeddingMaskGenerator`.
    - They each have an `initialize` and `generate` method, which precompute mask data and then allow to regenerate the segmentation with different parameters.
- Uses the new mask generators in the 2d annotator to enable immediate segmentation when changing `iou_pred_thresh` or `stability_score_thresh`.

PR #67, opened 2023-07-12
issue 40 add input data signature to embeddings file.

### LINKED ISSUE #40: Better caching of embeddings
Currently the caching of embeddings is not aware of changing inputs. So if the image or any relevant parameters (e.g. tiling) change the napari annotator is not aware and things just break because the shapes etc. don't match anymore.

The best solution would probably be to hash all relevant things (the input image and the tiling params) and then warn or throw an error if the hash doesn't agree.

cc @GenevieveBuckley

paulhfu:
fixes #40

### DIFF
--- a/micro_sam/util.py
+++ b/micro_sam/util.py
@@ -1,6 +1,7 @@
 import hashlib
 import os
-from shutil import copyfileobj
+import warnings
+from shutil import copyfileobj, rmtree
 
 import numpy as np
 import requests
@@ -8,6 +9,7 @@ import torch
 import vigra
 import zarr
 
+from PyQt5 import QtCore, QtWidgets
 from elf.io import open_file
 from nifty.tools import blocking
 from skimage.measure import regionprops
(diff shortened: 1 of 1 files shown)


constantinpape (commented):
Thanks! And this is exactly the right idea.
Two more points:
- We should check for a bit larger input how long it takes to compute the hash and whether the overhead would ever become noticable. Can you check how long it takes for the 3d data from https://github.com/computational-cell-analytics/micro-sam/blob/master/examples/sam_annotator_3d.py
- I am not sure if a warning is enough here. (Users usually just overlook them...). The ideal solution would be to have a small menu that pops up (if we have a data signature mismatch) and where the user has to confirm that they want to continue.

paulhfu:
Ok, I will do that.
Should there also be an option in that dialog to open a file explorer and create/select a new file?

constantinpape:
> Should there also be an option in that dialog to open a file explorer and create/select a new file?

Yes, that would be good! (And n case of a new file it should then trigger the embedding computation.)

paulhfu:
The suggested changes have been implemented.
The time of the 3dim input hash calc was:

timeit.timeit(lambda: hashlib.sha1(input_.tobytes()).hexdigest(), number=1)
0.14425896300053864

which is in seconds.

constantinpape:
> timeit.timeit(lambda: hashlib.sha1(input_.tobytes()).hexdigest(), number=1)
> 0.14425896300053864

Ok, that's absolutely negligible. We can stick with the straightforward hashing then.

micro_sam/util.py
```diff
@@ -314,6 +316,41 @@ def _precompute_3d(input_, predictor, save_path, lazy_loading, tile_shape=None,
...
+        string or os.path: path to a file (new or old) depending on user decision
+    """
+    msgbox = QtWidgets.QMessageBox()
+    msgbox.setWindowFlags(QtCore.Qt.CustomizeWindowHint | QtCore.Qt.WindowTitleHint)
+    msgbox.setWindowTitle("Warning")
+    msgbox.setText('The input data does not match the embeddings file.')
+    ignore_btn = msgbox.addButton("Ignore" , QtWidgets.QMessageBox.RejectRole)
+    overwrite_btn = msgbox.addButton("Overwrite file" , QtWidgets.QMessageBox.DestructiveRole)
+    select_btn = msgbox.addButton("Select different file" ,QtWidgets.QMessageBox.AcceptRole)
+    create_btn = msgbox.addButton("Create new file" ,QtWidgets.QMessageBox.AcceptRole)
```
constantinpape, line 339: Can we do proper file creation here, rather than having to first create the folder and then select it?
(This is not quite obvious, so it would be better if the user could just enter a new filename in the dialogue and then press `Create` or so).

If there is an issue that you actually need to create a folder for zarr: I suggest you just create a dummy file via the dialogue and then remove it again.

constantinpape (commented):
Thanks and this is great!
The only improvement left from my side is the file creation menu.

micro_sam/util.py
paulhfu: Yes there was the issue, that a dialog to create a directory does apparently not exist.
I used a "create file" dialog and tweaked it a bit so the user can just type in filename.zarr which is then created once the dialog is closed.

constantinpape (approved):
Thanks for fixing the file creation. This is much better now!

PR #78, opened 2023-07-17
Refactor the wrong file dialog gui

constantinpape:
To avoid pulling in PyQt on the top-level so that `micro_sam` can be used as proper library without GUI elements.

@anwai98 this should fix the error you reported.

fyi @paulhfu : from now on let's make sure all GUI elements go into `micro_sam/sam_annotator` ;) .


paulhfu:
@constantinpape, yes makes sense.

PR #87, opened 2023-07-23
Quit QApplication after entry gui is terminated

### LINKED ISSUE #85: Running the annotator raises warning
For me using the GUI entrypoint from `annotator` raises a warning:
```
/home/pape/Work/my_projects/micro-sam/micro_sam/sam_annotator/annotator_2d.py:243: UserWarning: A QApplication is already running with 1 event loop. To enter *another* event loop, use `run(max_loop_level=2)`
  napari.run()
```
This seems to be the case for using any of the 4 annotators with it (so far I have checked for `annotator_2d` and `image_series_annotator`).

paulhfu:
Fixes #85

Quit QApplication after entry gui is terminated.
Call correct annotator in tracking workflow.

### DIFF
--- a/micro_sam/sam_annotator/annotator.py
+++ b/micro_sam/sam_annotator/annotator.py
@@ -2,12 +2,14 @@ import os
 import magicgui
 import numpy as np
 from magicgui.widgets import Container, Label, LineEdit, SpinBox, ComboBox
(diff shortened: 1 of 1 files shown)


constantinpape (approved):
Thanks! I checked and it also fixes the warning for me.

PR #91, opened 2023-07-24
Add installer script

### LINKED ISSUE #68: Provide installation script for micro-sam
So that users can install it without having to install conda themselves.

The best option I am aware of for this is conda-constructor: https://github.com/conda/constructor

An example for where it is used is here: https://github.com/hci-unihd/covid-if-annotations/tree/master/dev/deployment (might be a bit outdated though).

paulhfu:
Fixes #68
Provide installation script and accompanying files for conda constructor installer build.


paulhfu:
@constantinpape,
This is not ready for review. 
I am waiting for micro-sam to be released as public conda package to finish this.
I am certain you can create a nicer image for thumbnails and backgrounds.

deployment/construct.yaml
```diff
@@ -0,0 +1,28 @@
...
+  - python =3.11
+  - napari
+  - pooch
+  - python-elf
+  - pytorch
+  - pytorch-cuda>=11.7  # you may need to update the cuda version to match your system
+  - torchvision
+  - tqdm
+
+# conclusion_text: short quickstart guide
```
paulhfu, line 23: Do you have suggestions for conclusion_text, welcome_text and company?

constantinpape:
> I am waiting for micro-sam to be released as public conda package to finish this.

The PR is already merged, see https://github.com/conda-forge/staged-recipes/pull/23374.
So it should be available by tomorrow. I will also work on a new release then, so that we have all the recent changes available.

> I am certain you can create a nicer image for thumbnails and backgrounds.

I created a (first version of a) logo yesterday. You can use that:
https://github.com/computational-cell-analytics/micro-sam/blob/master/doc/images/micro-sam-logo.png

constantinpape:
Hi @paulhfu ,
micro_sam is on conda-forge now: https://github.com/conda-forge/micro_sam-feedstock
I will work on a new release to include all the recent changes in the evening.

paulhfu:
Hi @constantinpape,

This is somehow not working yet. Solving the environment is taking forever as soon as I am adding a package besides the micro_sam package. I tried on ubuntu and windows.
I will have to look a bit deeper. Also, I think setting mamba as the conda solver is not working. Maybe doing the detour over an env.yaml file is a better solution.

paulhfu:
It is working when I create the env first manually with mamba and then build the installer from that.

deployment/construct.yaml
```diff
@@ -0,0 +1,22 @@
+version: 0.0.2
```
constantinpape, line 1: It would be nice to not hard-code this, but instead get it from here: https://github.com/computational-cell-analytics/micro-sam/blob/master/micro_sam/__version__.py, which is supposed to be "the single source of truth" regarding the version string. (Or alternatively you could also use the github tag, which should always be the same).

I am not sure about the best strategy for this, maybe a python script that takes this file as a template and then changes it to the correct version and resaves.

deployment/construct.yaml
constantinpape: I will think about it. For now, can you just put some dummy text like `Lorem ipsum`? So that I see where exactly this text shows up before writing it.

constantinpape:
> It is working when I create the env first manually with mamba and then build the installer from that.

Ok, that sounds like a viable solution.

constantinpape:
Hi @paulhfu ,
the CI issues are fixed now, so the tests here pass again. Let me know if this is ready to fully review.

paulhfu:
Hi @constantinpape,
There are no artifacts produced and I also did not test the CI script yet.
Hope to finish with it tonight. I will let you know.

paulhfu:
Hi @constantinpape,

There are still issues with creating a conda env in the workflow script. It seems that the windows workflow gets a bit further than the others. I will continue later with that.

(unrelated to this PR) but one thing I noticed, is that the file .github/workflows/doc_env.yaml is treated as workflow file.

constantinpape:
Hi @paulhfu ,

> There are still issues with creating a conda env in the workflow script. It seems that the windows workflow gets a bit further than the others. I will continue later with that.

Ok! Let me know if you are stuck somehow, or if I can help in any way.

> (unrelated to this PR) but one thing I noticed, is that the file .github/workflows/doc_env.yaml is treated as workflow file.

Good point, I will fix it!

constantinpape:
> (unrelated to this PR) but one thing I noticed, is that the file .github/workflows/doc_env.yaml is treated as workflow file.

I fixed that in #105

paulhfu:
Hi @constantinpape,
The Windows and Linux CI installers work fine.
I tried for some time now to get the macos CI installer to work.
It seems that I cannot install mamba via conda because libarchive is reluctant to install properly.
I did a detour using micromamba now which I can install with brew. But conda constructor has its issues with that.
Envs do not seem to be at the location where they are expected. 

I will continue fixing this last issue.

constantinpape:
I fixed the conda issue now. With release `0.1.1.post1` all the annotator commands work.
So the conda package should be ready for building the installer from it now.

> I will continue fixing this last issue.

:+1:

paulhfu:
Hi @constantinpape, this is ready for review now. The build of all installers is working fine.
Some notes:
- I could not make the windows installer include the startmenu shortcut on Windows. I activated appending the Path variable, so the CLI should be available.
- Macos installer looks a bit weird, but it is the only way that worked. Mamba installation over conda was not possible. Micromamba worked if installed with brew but was not accepted by constructor as a solver. Therefore mamba is installed over micromamba and constructor is called with mamba.
- I could only test the windows and the ubuntu installer.

.github/workflows/build_installers.yaml
```diff
@@ -0,0 +1,89 @@
...
+          micromamba install -y -c conda-forge python=3.10
+          micromamba install -y -c conda-forge constructor
+          micromamba install -y -c conda-forge ruamel.yaml
+          micromamba install -y -c conda-forge mamba
+          mamba env create --file=env.yaml
+          python version_getter.py
+          mkdir ./${{ matrix.os }}_x86_64
+          constructor --conda-exe=$(which mamba) --output-dir ./${{ matrix.os }}_x86_64 .
+
+      - name: upload installer
```
constantinpape, line 84: If I understand this correctly the artifact will not be committed to the repo, but be associated to the actions.
And it will currently be retained for only 5 days due to the `retention-days` setting.
Do I understand this correct?
(If yes, I think this makes sense for now).

constantinpape (approved):
Thanks! Can you answer the one comment I left in the code?
If I understand the current approach correctly then I would go ahead and merge it.

I see a few things to address as follow-up:
- Attach the file to the release directly instead of using the current artifact upload solution. I found two potential actions for this (only did a quick google search, maybe there are better solutons):
  - https://github.com/marketplace/actions/upload-to-github-release
  - https://github.com/marketplace/actions/upload-files-to-a-github-release
- Build the release once the actual package is on conda (which will not be sync with the release). I can think of two potential solutions rn (1st is more elegant, but 2nd much simpler):
   - Run action as a cron job that checks if the latest release has binaries uploaded, if not check if it's on conda forge, and if yes runs the actual workflow
   - Only trigger this on a post-release, that I would only publish after the full release is on conda-forge

But let's go ahead with the current implementation for now, I want to test this before we think about the next steps.

constantinpape:
> * I could not make the windows installer include the startmenu shortcut on Windows. I activated appending the Path variable, so the CLI should be available.

Ok, having the start menu would be fairly important so that this can be used by "non-computational" users. But let's tackle this in a follow-up after I (and students) had a chance to tackle the current version.

> * Macos installer looks a bit weird, but it is the only way that worked. Mamba installation over conda was not possible. Micromamba worked if installed with brew but was not accepted by constructor as a solver. Therefore mamba is installed over micromamba and constructor is called with mamba.

Whatever works :).

> * I could only test the windows and the ubuntu installer.

Ok! I will test all three installers with students once it's merged and then see if there are any other issues or required changes to the installers themselve.

paulhfu:
I was able to generate the startmenu on the machine that builds the installer. But constructor does not seem to pick up on it and it does not include it into the installer. I left the files which are necessary for the start menu in here for future work on it.

Documentation on menuinst is a bit tentative and it is also only supported for windows but it is mentioned, that it might also work on Macos and Ubuntu. I think that would also be worth a quick test.

.github/workflows/build_installers.yaml
paulhfu: Exactly, it will be available for 5 days on the workflow page.
https://github.com/computational-cell-analytics/micro-sam/actions/runs/5706862079

It would definitely quite convenient to directly zip a release.

paulhfu: Exactly, it will be available for 5 days on the workflow page.
https://github.com/computational-cell-analytics/micro-sam/actions/runs/5706862079

It would definitely quite convenient to directly zip a release.

constantinpape:
> I was able to generate the startmenu on the machine that builds the installer. But constructor does not seem to pick up on it and it does not include it into the installer. I left the files which are necessary for the start menu in here for future work on it.
> 
> Documentation on menuinst is a bit tentative and it is also only supported for windows but it is mentioned, that it might also work on Macos and Ubuntu. I think that would also be worth a quick test.

Ok, maybe we could open an issue on this in constructor to ask for help.

For now I will go ahead and merge it, so that we can test it here.

I will give you feedback as soon as we had a chance to test it! (Or let you know if any other issues arise.)

PR #98, opened 2023-07-26
Add preliminary custom models

constantinpape:


constantinpape:
Looks like something in pip is broken, check again tomorrow.

PR #108, opened 2023-07-30
Update documentation for the annotation tools

constantinpape:
cc @paulhfu : this updates some of the function names for things you have added, so that they are excluded from the documentation.

I don't think that interferes with any of your changes in #91, but still wanted to ping you here as a heads up.

### DIFF
--- a/build_doc.py
+++ b/build_doc.py
@@ -1,0 +1,17 @@
+import argparse
+from subprocess import run
+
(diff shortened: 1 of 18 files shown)


paulhfu:
Thanks for the heads up @constantinpape. #91 is almost finished and there are no conflicts. Good to see this refactoring in any case. From now on I pay more attention to privacy markers and type hints.

PR #124, opened 2023-08-06
109_fix_macos_installer

### LINKED ISSUE #109: Feedback installers
Hi @paulhfu ,
this issue is for collecting feedback on the installers. 

For now I only have positive things to report:
- The CI runs through without an issue and the artifacts are uploaded to https://github.com/computational-cell-analytics/micro-sam/actions/runs/5707871215
- I tested the linux installer and it works.

@sushmita-nair and @Sagnik700: could you please test the windows installer?

@Marei33: could you please test the mac installer? 

For downloading the installers please follow the link above, and scroll to the bottom, then you can download the respective installer:

image

paulhfu:
Fixes #109

- do not add to system path during installation

### DIFF
--- a/deployment/construct.yaml
+++ b/deployment/construct.yaml
@@ -12,4 +12,4 @@ welcome_text: Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do ei
   tempor incididunt ut labore et dolore magna aliqua.
 conclusion_text: Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris
   nisi ut aliquip ex ea commodo consequat.
-initialize_by_default: true
+initialize_by_default: false


paulhfu:
Hi @constantinpape,

shall I proceed to add the signing workflow with a "Developer ID Installer certificate" for MacOs?
I created a free developer account but certificates seem to be behind a paywall.
Information about signing and the certificates can be found [here](https://help.apple.com/xcode/mac/current/#/dev033e997ca) and [here](https://developer.apple.com/help/account/create-certificates/create-developer-id-certificates/).

constantinpape:
Hi @paulhfu ,
is the signing necessary for the installation to work, or will it only throw a warning if it's not signed?

constantinpape:
Ah, I found it:
```
On macOS, if your app isn’t downloaded from the Mac App Store or signed with a Developer ID certificate, it won’t launch unless the user completely disables [Gatekeeper](https://help.apple.com/xcode/mac/current/#/dev190d53ccb). Users have the option of enabling or disabling identified developers in System Preferences.
```

deployment/construct.yaml
```diff
@@ -12,4 +12,4 @@ welcome_text: Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do ei
   tempor incididunt ut labore et dolore magna aliqua.
 conclusion_text: Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris
   nisi ut aliquip ex ea commodo consequat.
-initialize_by_default: true
\ No newline at end of file
+initialize_by_default: false
```
constantinpape, line 15: What does this change mean?

constantinpape:
Ok, so from what I understand, for signing you need a developer account, which is 100$ a year. So let's not do that.
Either we:
- Use the bundled app solution (I will follow up in #109 on that)
- Or we don't provide a Mac installer

paulhfu:
Yes, this is also how I understood it.

deployment/construct.yaml
paulhfu: This will disable the default activation of the "append to system path" - checkbox.

constantinpape (approved):

constantinpape:
I will go ahead and merge it, but reopen #109 so we can discuss there if there is anything else we can do for mac.

PR #125, opened 2023-08-06
Add AMG

anwai98:
- Added the Automatic Mask Generation using Grid Search for FT / Vanilla Models

### DIFF
--- a/micro_sam/evaluation/automatic_mask_generation.py
+++ b/micro_sam/evaluation/automatic_mask_generation.py
@@ -8,25 +8,11 @@ from glob import glob
 import imageio.v2 as imageio
 from elf.evaluation import mean_segmentation_accuracy
 
+from micro_sam.instance_segmentation import mask_data_to_segmentation
 from micro_sam.util import precompute_image_embeddings
 from micro_sam.instance_segmentation import AutomaticMaskGenerator
 
 from .inference import get_predictor
-
-
-def remove_bg_as_biggest_element(image, masks):
-    """This function removes the background by sorting the masks based on area,
-    and ignores the object with biggest area
-    """
-    instance_labels = np.zeros((image.shape[0], image.shape[1]), dtype=int)
-
-    sorted_anns = sorted(masks, key=(lambda x: x['area']), reverse=True)
-    for j, ann in enumerate(sorted_anns):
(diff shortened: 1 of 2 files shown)


micro_sam/evaluation/automatic_mask_generation.py
```diff
@@ -0,0 +1,180 @@
...
+import imageio.v2 as imageio
+from elf.evaluation import mean_segmentation_accuracy
+
+from micro_sam.util import precompute_image_embeddings
+from micro_sam.instance_segmentation import AutomaticMaskGenerator
+
+from .inference import get_predictor
+
+
+def remove_bg_as_biggest_element(image, masks):
```
constantinpape, line 17: Please remove this and use https://github.com/computational-cell-analytics/micro-sam/blob/master/micro_sam/instance_segmentation.py#L53 instead.

micro_sam/evaluation/automatic_mask_generation.py
```diff
@@ -0,0 +1,180 @@
...
+                "SA75": sas[5]
+            }
+            tmp_df = pd.DataFrame([result_dict])
+            net_list.append(tmp_df)
+
+    img_gs_df = pd.concat(net_list)
+    img_gs_df.to_csv(img_save_path, index=False)
+
+
+def per_image_amg(args):
```
constantinpape, line 67: This should all be independent of LiveCELL. Please have a look how I separated out things in
- https://github.com/computational-cell-analytics/micro-sam/blob/dev/micro_sam/evaluation/inference.py#L306 and
- https://github.com/computational-cell-analytics/micro-sam/blob/dev/micro_sam/evaluation/livecell.py#L38

The functionality here should be generic and just take `image_paths` and `gt_paths`. It also shouldn't use `argparse`, but provide a clean functional interface instead.
And you can implement livecell specific functions using the functionality from here in `livecell.py`.

anwai98: Ahha yes, makes total sense to generalise it already! Cleaned up the functionality as per your previous snippets' expectations!

constantinpape:
Thanks for the changes @anwai98. I will go ahead and merge this and then do some more clean up myself.

PR #131, opened 2023-08-08
Fix Multimasking

anwai98:
- Updated the `n_prompts` (number of objects) to now accept only fixed values (however, still haven't updated the variable name as also mentioned in the discussion)
- Updated the multimasking to False while we accept box-only prompts!


constantinpape (approved):
Thanks!

Issue #129, opened 2023-08-08
Reset track_state to track after a commit


Sagnik700:
If the `track_state` was kept in `division` before clicking on Commit, the new annotation started with the track_state being in `division` rather than being in `track`. If it is not noticed then it leads to multiple errors and wrong segmentations, so after committing every track, the `track_state` should be reset to `track`

constantinpape:
Hi Sagnik,
thanks for reporting. I gave this a try in #132 ,  but unfortunately it doesn't work as expected.
I will need to dig deeper, but probably won't have time before I go on holidays.

constantinpape:
After reviewing the design, this unfortunately is not possible: the track state always refers to the last point that was placed, so updating it would undo the division label.

PR #132, opened 2023-08-08
Reset tracking menu state to track

constantinpape:
See also #129


constantinpape:
This change does not make sense.

PR #133, opened 2023-08-09
130_fix_file_dialogs

paulhfu:
Tested on Windows. Yet to test on Linux.


constantinpape (approved):
Thanks! Will merge once the test passes (seems to be a random error)

PR #135, opened 2023-08-10
Iterative Prompting for Inference

anwai98:
- Added Iterative Prompt Generator and Inference Scheme (haven't tested it, however have used the same logic from our trainer to iteratively update the prompts n times)


constantinpape (approved):
Thanks! Looks good, and I will go ahead and fix up the rest.

PR #164, opened 2023-08-26
109 macos installer

paulhfu:


paulhfu:
For lack of hardware I could not locally debug the MacOs application build. Here are instructions to do that if someone with appropriate hardware wants to continue.
The application is build via [Pyinstaller](https://pyinstaller.org/en/stable/). How to setup the build env can be found in this branch in:
`.github/workflows/build_installers.yaml`
Basically, Pyinstaller has to be included in the micro_sam conda env. Then the build script
`deployment/build_pyinstaller_app.sh`
can be executed.
After the build, the app can be found under `deployment/dist/annotator/annotator`.
Pyinstaller does not do the best job when it comes to collecting the correct files needed for the application runtime.
Therefore the app should be tested thoroughly. If an error comes, e.g. a missing package, it can be included with the `--hidden-import` flag. This can be added in the build script.
If modules are missing data, it can be included with the `--add-data` flag. This flag just copies the referenced resource, so it can also be used to copy whole python modules if the `--hidden-import` does not work.
Sometimes it is a bit tricky to find the missing resource because the application only complains about missing libs which can not always be directly linked to python modules in the env.
If everything works, we can consider to use the `--onefile` option which will produce one single executable. However, the app and its whole env will be packed into this exe and during runtime it will unpack which leads to longer startup times.

constantinpape:
Thanks @paulhfu! We will find someone with a MAC who can take this over.

PR #166, opened 2023-08-31
Fix YAML formatting errors so Github actions will run

GenevieveBuckley:
I noticed that the `build_docs.yaml` GitHub action is failing to run, because there is an error in the YAML formatting for that workflow.

https://github.com/computational-cell-analytics/micro-sam/actions/runs/6034379146

>  Invalid workflow file: .github/workflows/build_docs.yaml#L22
> You have an error in your yaml syntax on line 22

This PR fixes the YAML formatting by replacing tab indentation with spaces, and additionally removes some comment lines (I'm not sure if they were also causing formatting errors, but it makes it more difficult to understand whether those lines should actually be included and run or not)

### DIFF
--- a/.github/workflows/build_docs.yaml
+++ b/.github/workflows/build_docs.yaml
@@ -19,14 +19,7 @@ permissions:
 jobs:
   # Build the documentation and upload the static HTML files as an artifact.
   build:
-	runs-on: ubuntu-latest
-  # runs-on: ${{ matrix.platform }}
-  # strategy:
-  #   matrix:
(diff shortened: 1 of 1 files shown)


GenevieveBuckley:
Here are the errors yaml lint showed for the build_docs.yaml file on the current master branch 
https://www.yamllint.com/

Tabs are not allowed as indentation at line 22, column 1
All mapping items must start at the same column at line 29, column 1
All mapping items must start at the same column at line 58, column 1

constantinpape (approved):

Issue #167, opened 2023-08-31
Providing micro-sam as napari plugin widgets


GenevieveBuckley:
I think we need a thread to discuss implementing napari plugin widgets for micro-sam.
This is follow up work to https://github.com/computational-cell-analytics/micro-sam/pull/165

GenevieveBuckley:
Also including an error Constantin ran into (I have not seen the same error, but will keep an eye out for it now)
https://github.com/computational-cell-analytics/micro-sam/pull/165#issuecomment-1697518398
> When trying to start the 2d annotator plugin I get an error:
>
> ``` RuntimeError: Failed to import command at 'micro_sam.sam_annotator.annotator_2d:annotator_2d_widget': /lib/x86_64-linux-gnu/libtinfo.so.6: version `NCURSES6_TINFO_6.2.20211010' not found (required by /home/pape/software/conda/miniconda3/envs/sam-plugin/lib/python3.10/site-packages/vigra/../../../././libncurses.so.6) ```
>
> (Just fyi, just importing vigra works).

constantinpape:
Hi @GenevieveBuckley,
thanks for opening the issue and working on this! Here are my initial thoughts on how to change the architecture to enable the plugins.

## 1. Computing Image Embeddings and "raw" image layer

We add a new widget for the embedding computation. The user can just select an image layer that is used to precompute the embeddings (and optionally to specify a path where the embeddings are cached). This will also solve #161 (cause we get a progress bar "for free") and #17 as the name "raw" is not hard-coded anymore.
As far as I understand we can't rely on global variables any more when we switch to the plugin, so we need to store the image embeddings somewhere  else than in `IMAGE_EMBEDDINGS`, e.g. as a property of the viewer or the corresponding image layer. (Same is true for the other global state we currently have)

## 2. Other hard-coded layers

For the other hard-coded layers (`prompts`, `box_prompts`, `current_object`, `commited_objects`, etc.) it would be best to keep them hard-coded, and if possible to just have the plugin create them when called. From my understanding this should be possible by having type annotated return values for these layers.

## 3. Type annotations and magicfactory

[...]

PR #177, opened 2023-09-05
Prototype napari plugin widget for annotator_2d

GenevieveBuckley:
Relates to https://github.com/computational-cell-analytics/micro-sam/issues/167


micro_sam/sam_annotator/annotator_2d.py
```diff
@@ -191,6 +217,41 @@ def annotator_2d(
...
+    raw: ImageData,
+    v: napari.Viewer,
+    embedding_path: os.PathLike = "./embeddings",
+    show_embeddings: bool = False,
+    segmentation_result: Optional[LabelsData] = None,
+    precompute_amg_state: bool = False,
+    model_type: AVAILABLE_MODELS = AVAILABLE_MODELS[_DEFAULT_MODEL],
+    tile_shape: Tuple[int, int] = (None, None),
+    halo: Tuple[int, int] = (None, None),
+    predictor: Optional[SamPredictor] = None,  # not accessible via magicgui widget, not a recognised type
```
constantinpape, line 254: Regarding the comment: that's ok. This doesn't make sense to pass via napari anyways. The predictor is currently exposed for 2 reasons:
1. To pass the already initialized predictor in the `image_series_annotator`, where the logic for image embeddings is done externally.
2. To enable passing predictors that were loaded from a custom checkpoint.

As far as I understand your comment, it doesn't hurt to leave it in for now. **After we have merged this PR** I can rethink a bit how this is handled because: re 1. we need to change the `image_series_annotator` logic anyways and re 2.: it would be good to enable loading of custom models also from the napari plugin. (Here: custom model means that we are using a url or filepath to specifiy the checkpoint rather than one of the names from `AVAILABLE_MODELS`).

micro_sam/sam_annotator/annotator_2d.py
```diff
@@ -224,39 +285,51 @@ def annotator_2d(
...
+        v.add_image(embedding_vis, name="embeddings", scale=scale)
+
     # we set the pre-computed image embeddings if we don't use tiling
     # (if we use tiling we cannot directly set it because the tile will be chosen dynamically)
     if tile_shape is None:
+        print("tiling precompute")
         util.set_precomputed(PREDICTOR, IMAGE_EMBEDDINGS)
 
-    # viewer is freshly initialized
-    if v is None:
```
constantinpape, line 245: Just fyi: this condition is for the `image_series_annotator`: in that case we pass the napari viewer object via the `v` argument. Hence it is not `None` and `_update_viewer` is called to replace all relevant layers.
I just wanted to clarify because it relates to my comment on `_update_viewer`. We can go ahead and remove the condition since the `image_series_annotator` needs to be redone after finalizing the 2d annotator.

constantinpape (commented):
Thanks for starting to work on this @GenevieveBuckley! I checked the high level structure and this perfectly matches what we discussed in #167. I haven't tried running it myself yet, since I am not sure if it's ready for that yet. Let me know if/once it is and then I will test it!
Besides that I just left a few minor comments with clarifications (and one minor change request).
Let me know if you need any input from my side to continue!

constantinpape:
Hi @GenevieveBuckley ,
I have started with a developer documentation [here](https://github.com/computational-cell-analytics/micro-sam/blob/dev/doc/development.md) to give a bit more high level overview of the intended design for the plugin, including a flow diagram that exemplifies it for the 2d annotation tool. I hope this is helpful and please let me know if you have any more questions about this.

For this PR I suggest you only focus on the 2d annotation plugin, i.e. that `annotator_2d_plugin` and the corresponding napari plugins work as intended. We can then take care of the CLI function (`annotator_2d`) and the other plugins in follow-up PRs. 
From what I can see the following things are still missing for that:
- The `embedding widget`, including how we handle that the name of the Image layer is not hard-coded to `raw` any more.
- How we deal with the global variables? Can we move the corresponding state into attributes of the viewer and / or layers?

constantinpape:
Superseeded by recent changes.

PR #178, opened 2023-09-06
Supporting Apple Silicon devices

GenevieveBuckley:
As [discussed here](https://github.com/computational-cell-analytics/micro-sam/pull/176#issuecomment-1707524862), the mps torch backend does not support 64 bit tensors. This PR updates @constantinpape's `mps` branch of micro-sam to help fix that problem.

Some notes:
- Apparently the torch nightly version does support 64 bit tensors for the mps backend (Apple Silicon). I haven't tried it yet. I guess you *could* make the nightly torch version a requirement, but this seems like a hassle.
- We don't just have problems in micro-sam, there are also problems with this in the segment-anything repository. There are three options for dealing with this:
   1. **My preferred option:** vendor the function. We can keep an edited copy of the `batched_mask_to_box` function from segment-anything (needed for pre-computing AMG) in the micro-sam repository. Later on, if the bug gets fixed in the upstream repositories, we can remove this from micro-sam.
   2. Ask users to install the nightly version of torch, or wait for that to become available more generally (not ideal, it involves hassle for users, or waiting for a later pytorch release)
   3. Wait for [a PR to be merged](https://github.com/facebookresearch/segment-anything/pull/122) in segment-anything so it can support older versions of torch with the mps backend (also not ideal, it involves a lot of waiting for this to become available to users. Also, I don't think the facebook team are very motivated to merge PRs - because [there is an existing PR with the changes we want](https://github.com/facebookresearch/segment-anything/pull/122) but there has been no activity since June. For this reason, it is my least preferred option.)

### DIFF
--- a/micro_sam/__init__.py
+++ b/micro_sam/__init__.py
@@ -5,5 +5,8 @@
 .. include:: ../doc/python_library.md
 .. include:: ../doc/finetuned_models.md
 """
+import os
 
 from .__version__ import __version__
+
+os.environ["PYTORCH_ENABLE_MPS_FALLBACK"] = "1"
--- a/micro_sam/instance_segmentation.py
+++ b/micro_sam/instance_segmentation.py
@@ -189,7 +189,7 @@ class AMGBase(ABC):
             mask, changed = amg_utils.remove_small_regions(mask, min_area, mode="islands")
             unchanged = unchanged and not changed
 
-            new_masks.append(torch.as_tensor(mask).unsqueeze(0))
+            new_masks.append(torch.as_tensor(mask, dtype=torch.int).unsqueeze(0))
             # give score=0 to changed masks and score=1 to unchanged masks
             # so NMS will prefer ones that didn't need postprocessing
             scores.append(float(unchanged))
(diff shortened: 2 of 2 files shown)


GenevieveBuckley:
Also, can you upvote [this comment](https://github.com/pytorch/pytorch/issues/77764#issuecomment-1687703864) please? It affects everyone who uses this repository, so it would be good to have that reflected in the number of upvotes.

It's the reason for the failing CI test `test_automatic_mask_generator`, since `torchvision.rms` is not implemented for MPS backends.

I think it's sensible to use the suggestion to set the environment variable `PYTORCH_ENABLE_MPS_FALLBACK=1`. Unless one of us decides to work on it, I don't expect it will be implemented in the near future.

Here's the error message from miro-sam `test_automatic_mask_generator`:
> The operator 'torchvision::nms' is not currently implemented for the MPS device. If you want this op to be added in priority during the prototype phase of this feature, please comment on https://github.com/pytorch/pytorch/issues/77764. As a temporary fix, you can set the environment variable `PYTORCH_ENABLE_MPS_FALLBACK=1` to use the CPU as a fallback for this op. WARNING: this will be slower than running natively on MPS.

constantinpape (approved):
Hi @GenevieveBuckley,
thanks for working on this, and the changes here look good!
I will merge it now and then comment more generally in #176.

constantinpape:
For some reason the changes to the dtypes affect the quality of the test outcomes. I will go ahead and merge in spite of this and then have a look at this myself. (This likely shows that these tests are too brittle and should be fixed ...)

PR #179, opened 2023-09-06
Clear torch memory cache in tests

GenevieveBuckley:
I've been running out of memory when running the test suite locally. If I clear the torch cache before the two worst memory hogs (the tiling tests) this helps the situation a lot.


constantinpape:
The relevant functionality is now on dev so I will merge.

PR #180, opened 2023-09-07
Add MPS Pytorch support for batched_mask_to_box function

GenevieveBuckley:
This PR adds an edited copy of the `batched_mask_to_box()` function from `segment_anything/utils/amg.py` to the micro-sam codebase. The function has been edited to solve two problems:
1. It makes this function compatible with the MPS PyTorch backend for apple silicon, and
2. We add a check that the input mask dtype is boolean (since apparently there is some bug that means masks with integer dtypes return completely different results 👀 )

The license for segment-anything is Apache 2.0, which is compatible with the MIT licence micro-sam uses.

Eventually, if this gets fixed upstream we can remove it from the micro-sam code and go back to using the `batched_mask_to_box()` function directly from segment-anything. But that seems unlikely to happen soon.
- Currently there is a PR open for adding MPS support, but there has been no activity since June 2023
https://github.com/facebookresearch/segment-anything/pull/122
- As far as I know, no one has reported the bug involving boolean vs integer mask input to `batched_mask_to_box()`. Creating an issue would be helpful here (maybe I'll try and do that this evening).

Extra context: [discussed here](https://github.com/computational-cell-analytics/micro-sam/pull/176#discussion_r1318187348) previously on the main PR for adding .MPS support to micro-sam.

### DIFF
--- a/micro_sam/_vendored.py
+++ b/micro_sam/_vendored.py
@@ -1,0 +1,58 @@
+"""
+Functions from other third party libraries.
+
+We can remove these functions once the bug affecting our code is fixed upstream.
+
+The license type of the thrid party software project must be compatible with
+the software license the micro-sam project is distributed under.
+"""
+import torch
+
+
+# segment_anything/util/amg.py
+# https://github.com/facebookresearch/segment-anything
+def batched_mask_to_box(masks: torch.Tensor) -> torch.Tensor:
+    """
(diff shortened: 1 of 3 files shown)


GenevieveBuckley:
Update: I did make an issue https://github.com/facebookresearch/segment-anything/issues/552

constantinpape (approved):

Issue #181, opened 2023-09-07
Status of MPS Support


constantinpape:
Keeping track of the MPS related things @GenevieveBuckley and me discovered while working on MPS support:
- There is a PR in SAM to enable it: https://github.com/facebookresearch/segment-anything/pull/122. The necessary changes are all in `AMG`, and in code that we duplicate in order to make using the mask generation interactively more convenient. So whether this PR gets merged or not does not make a real difference for us, but it's still good to keep track of it.
- We discovered a bug in the bounding box generation, that leads to different results for integer versus boolean masks. For now, Gen has fixed this in #180 by copying the relevant function and fixing it. She also created an issue in SAM: https://github.com/facebookresearch/segment-anything/issues/552. If they ever fix it we could remove the vendored function and just depend on SAM again.
- It's currently unclear if MPS speeds things up or slows them down. I will create some benchmark scripts we can use to check. Depending on this we can decide whether to use MPS by default if it is available or not.

constantinpape:
I implemented the benchmarks in #182 now. They are implemented in https://github.com/computational-cell-analytics/micro-sam/blob/dev/development/benchmark.py.
The script accepts the arguments `-m` to set the model type (either `vit_h, vit_l or vit_m`) and `-d` to set the device (either `cpu`, `cuda` or `mps`, latter two if available).

@GenevieveBuckley @Marei33 could you run these benchmarks on your laptops with the following four settings please:
`vit_b, cpu`, `vit_b, mps`, `vit_h, cpu`, `vit_h, mps`.

And then paste the results here. To run them you will need to check out the latest `dev` branch. Let me know if anything is unclear.

For reference, here are the results for `vit_b` on my (linux) laptop:

| model   | device   | benchmark             |    runtime |
|:--------|:---------|:----------------------|-----------:|
| vit_b   | cpu      | embeddings            |  5.72605   |
| vit_b   | cpu      | prompt-p1n0           |  0.0527573 |
| vit_b   | cpu      | prompt-p2n4           |  0.0473237 |
| vit_b   | cpu      | prompt-box            |  0.0519929 |
| vit_b   | cpu      | prompt-box-and-points |  0.0487959 |
| vit_b   | cpu      | amg                   | 70.4746    |

GenevieveBuckley:
[...]

PR #183, opened 2023-09-08
Trying to improve performance for MPS

constantinpape:
This PR introduces two changes to try and improve the AMG performance for MPS backend:
- Implement a numpy based function for `mask_to_rle_pytorch` in `_vendored`.
- Set `points_per_batch` to `None` in the constructor and choose 16 if the backend is MPS. 

See #181 for motivation.

For me there is no difference in runtime when using the version from SAM compared to the numpy version in vendored. I tried both with my laptop (`cpu`) and a A100 GPU on the cluster (`cuda`).
Note that I tried the RLE implementation from https://www.kaggle.com/code/stainsby/fast-tested-rle/notebook but it gave wrong results. I used the RLE implementation from SO instead.

AMG Benchmark with `utils.amg.mask_to_rle_pytorch`:

| model   | device   | benchmark   |   runtime |
|:--------|:---------|:------------|----------:|
| vit_b   | cpu      | amg         |   77.9112 |
| vit_b   | cuda     | amg         |   7.02548 |

AMG Benchmark with `vendored.mask_to_rle_pytorch`:

| model   | device   | benchmark   |   runtime |
|:--------|:---------|:------------|----------:|
| vit_b   | cpu      | amg         |   77.0576 |
| vit_b   | cuda     | amg         |    7.0637 |

### DIFF
--- a/micro_sam/_vendored.py
+++ b/micro_sam/_vendored.py
@@ -1,11 +1,14 @@
 """
 Functions from other third party libraries.
 
-We can remove these functions once the bug affecting our code is fixed upstream.
+We can remove these functions once the bugs affecting our code is fixed upstream.
 
 The license type of the thrid party software project must be compatible with
 the software license the micro-sam project is distributed under.
 """
+from typing import Any, Dict, List
+
+import numpy as np
 import torch
 
 
@@ -58,4 +61,29 @@ def batched_mask_to_box(masks: torch.Tensor) -> torch.Tensor:
     else:
         out = out[0]
 
+    return out
+
+
+# segment_anything/util/amg.py
+# https://github.com/facebookresearch/segment-anything
+def mask_to_rle_pytorch(tensor: torch.Tensor) -> List[Dict[str, Any]]:
+    """Calculates the runlength encoding of binary input masks.
+
+    Implementation based on
+    https://stackoverflow.com/questions/1066758/find-length-of-sequences-of-identical-values-in-a-numpy-array-run-length-encodi
+    """
+    # Put in fortran order and flatten h, w
+    b, h, w = tensor.shape
+    tensor = tensor.permute(0, 2, 1).flatten(1)
+    tensor = tensor.detach().cpu().numpy()
+
+    n = tensor.shape[1]
+
+    # encode the rle for the individual masks
+    out = []
+    for mask in tensor:
+        diffs = mask[1:] != mask[:-1]  # pairwise unequal (string safe)
+        indices = np.append(np.where(diffs), n - 1)  # must include last element posi
+        counts = np.diff(np.append(-1, indices))  # run lengths
+        out.append({"size": [h, w], "counts": counts.tolist()})
     return out
--- a/test/test_vendored.py
+++ b/test/test_vendored.py
@@ -3,41 +3,82 @@ import unittest
 import numpy as np
 import torch
 
+from segment_anything.utils.amg import mask_to_rle_pytorch as mask_to_rle_pytorch_sam
+from skimage.draw import random_shapes
+
 
 class TestVendored(unittest.TestCase):
-    def setUp(self):
-        mask_numpy = np.zeros((10,10)).astype(bool)
+    def _get_mask_to_box_data():
+        mask_numpy = np.zeros((10, 10)).astype(bool)
         mask_numpy[7:9, 3:5] = True
-        self.mask = mask_numpy
-        self.expected_result = [3, 7, 4, 8]
+        expected_result = [3, 7, 4, 8]
+        return mask_numpy, expected_result
 
-    def test_cpu_batched_mask_to_box(self):
+    def _test_batched_mask_to_box(self, device):
         from micro_sam._vendored import batched_mask_to_box
 
-        device = "cpu"
+        mask, expected_result = self._get_mask_to_box_data()
         mask = torch.as_tensor(self.mask, dtype=torch.bool, device=device)
         expected_result = torch.as_tensor(self.expected_result, dtype=torch.int, device=device)
         result = batched_mask_to_box(mask)
         assert all(result == expected_result)
+
+    def test_cpu_batched_mask_to_box(self):
+        self._test_batched_mask_to_box(device="cpu")
 
     @unittest.skipIf(not torch.cuda.is_available(),
                      "CUDA Pytorch backend is not available")
     def test_cuda_batched_mask_to_box(self):
-        from micro_sam._vendored import batched_mask_to_box
-
-        device = "cuda"
-        mask = torch.as_tensor(self.mask, dtype=torch.bool, device=device)
-        expected_result = torch.as_tensor(self.expected_result, dtype=torch.int, device=device)
-        result = batched_mask_to_box(mask)
-        assert all(result == expected_result)
+        self._test_batched_mask_to_box(device="cuda")
 
     @unittest.skipIf(not (torch.backends.mps.is_available() and torch.backends.mps.is_built()),
                      "MPS Pytorch backend is not available")
     def test_mps_batched_mask_to_box(self):
-        from micro_sam._vendored import batched_mask_to_box
+        self._test_batched_mask_to_box(device="mps")
 
-        device = "mps"
-        mask = torch.as_tensor(self.mask, dtype=torch.bool, device=device)
-        expected_result = torch.as_tensor(self.expected_result, dtype=torch.int, device=device)
-        result = batched_mask_to_box(mask)
-        assert all(result == expected_result)
+    def _get_mask_to_rle_pytorch_data(self):
+        shape = (128, 256)
+
+        # randm shapes for 6 masks
+        n_masks = 6
+        masks, _ = random_shapes(shape, min_shapes=n_masks, max_shapes=n_masks)
+        masks = masks.astype("uint32").sum(axis=-1)
+
+        bg_id = 765  # bg val is 3 * 255 = 765
+        mask_ids = np.setdiff1d(np.unique(masks), bg_id)
+
+        one_hot = np.zeros((len(mask_ids),) + shape, dtype=bool)
+        for i, idx in enumerate(mask_ids):
+            one_hot[i, masks == idx] = 1
+        one_hot = torch.from_numpy(one_hot)
+
+        # make sure that all corner pixels are zero
+        one_hot[:, 0, 0] = 0
+        one_hot[:, -1, -1] = 0
+        one_hot[:, 0, -1] = 0
+        one_hot[:, -1, 0] = 0
+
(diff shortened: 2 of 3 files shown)


constantinpape:
Hi @GenevieveBuckley,
I hope this fixes the performance issues with MPS now. Could you give it a try and check if it actually helps?
Thanks!

GenevieveBuckley:
> I used the RLE implementation from SO instead

Could you link to the Stack Overflow thread? Thanks!

GenevieveBuckley:
Summary: these changes make the amg benchmark with `vit_b` and `vit_l` roughly equal on CPU vs MPS. However, for the `vit_h` model, the amg benchmark is still a lot slower on MPS.  

Here are the benchmark timings from the `rle` branch (this PR):

### vit_h
`python benchmark.py --model_type vit_h --device cpu` / `mps`
| model   | device   | benchmark             |    runtime |
|:--------|:---------|:----------------------|-----------:|
| vit_h   | cpu      | embeddings            | 18.9906    |
| vit_h   | cpu      | prompt-p1n0           |  0.036999  |
| vit_h   | cpu      | prompt-p2n4           |  0.0369489 |
| vit_h   | cpu      | prompt-box            |  0.0359869 |
| vit_h   | cpu      | prompt-box-and-points |  0.0386569 |
| vit_h   | cpu      | amg                   | 47.025     |
|---------|----------|-----------------------|-------------|
| vit_h   | mps      | embeddings            |  13.2879    |
| vit_h   | mps      | prompt-p1n0           |   0.031723  |
| vit_h   | mps      | prompt-p2n4           |   0.02931   |
| vit_h   | mps      | prompt-box            |   0.0253892 |
| vit_h   | mps      | prompt-box-and-points |   0.028873  |
| vit_h   | mps      | amg                   | 109.321     |
| vit_h   | mps      | amg                   | 62.9395     |
| vit_h   | mps      | amg                   | 120.484     |

* I ran the amg benchmark with `vit_h` and `mps` several times, because the time taken varied a lot. 

### vit_b
`python benchmark.py --model_type vit_b --device cpu` / `mps`
| model   | device   | benchmark             |    runtime |
|:--------|:---------|:----------------------|-----------:|
| vit_b   | cpu      | embeddings            |  3.7418    |
| vit_b   | cpu      | prompt-p1n0           |  0.0377681 |
| vit_b   | cpu      | prompt-p2n4           |  0.036375  |
| vit_b   | cpu      | prompt-box            |  0.035975  |
| vit_b   | cpu      | prompt-box-and-points |  0.0358782 |
| vit_b   | cpu      | amg                   | 39.9603    |
|---------|----------|-----------------------|------------|
| vit_b   | mps      | embeddings            |  2.10778   |
| vit_b   | mps      | prompt-p1n0           |  0.032361  |
| vit_b   | mps      | prompt-p2n4           |  0.0296309 |
| vit_b   | mps      | prompt-box            |  0.0258861 |
| vit_b   | mps      | prompt-box-and-points |  0.029561  |
| vit_b   | mps      | amg                   | 41.3489    |

### vit_l
`python benchmark.py --model_type vit_l --device cpu` / `mps`
| model   | device   | benchmark             |    runtime |
|:--------|:---------|:----------------------|-----------:|
| vit_l   | cpu      | embeddings            |  9.38109   |
| vit_l   | cpu      | prompt-p1n0           |  0.0396922 |
| vit_l   | cpu      | prompt-p2n4           |  0.03947   |
| vit_l   | cpu      | prompt-box            |  0.0396121 |
| vit_l   | cpu      | prompt-box-and-points |  0.0398631 |
| vit_l   | cpu      | amg                   | 40.0773    |
|---------|----------|-----------------------|------------|
| vit_l   | mps      | embeddings            |  4.78144   |
| vit_l   | mps      | prompt-p1n0           |  0.031987  |
| vit_l   | mps      | prompt-p2n4           |  0.029345  |
| vit_l   | mps      | prompt-box            |  0.0258131 |
| vit_l   | mps      | prompt-box-and-points |  0.0297689 |
| vit_l   | mps      | amg                   | 39.9337    |

test/test_vendored.py
```diff
@@ -3,41 +3,75 @@
...
+    def _get_mask_to_box_data(self):
+        mask_numpy = np.zeros((10, 10)).astype(bool)
         mask_numpy[7:9, 3:5] = True
-        self.mask = mask_numpy
-        self.expected_result = [3, 7, 4, 8]
+        expected_result = [3, 7, 4, 8]
+        return mask_numpy, expected_result
 
-    def test_cpu_batched_mask_to_box(self):
+    def _test_batched_mask_to_box(self, device):
```
GenevieveBuckley, line 17: I like this refactoring.

micro_sam/_vendored.py
```diff
@@ -59,3 +62,31 @@ def batched_mask_to_box(masks: torch.Tensor) -> torch.Tensor:
         out = out[0]
 
     return out
+
+
+# segment_anything/util/amg.py
+# https://github.com/facebookresearch/segment-anything
```
GenevieveBuckley, line 68: We should really say more about:
1. Why we felt we needed to put these functions here, and 
2. What condition needs to be met to remove it and go back to a more official implementation. How will someone know in future when it is possible or desirable to remove these workarounds?

And yes, I didn't write either of those things down when I added a function here 😆 Oops, I think that was a mistake and we should fix that now too.

GenevieveBuckley:
Overall, this is an excellent result! I'm very pleased, and think this PR should be merged, ideally in time to make it into the next release.

Discussion:
It is likely that the poor performance we see on vit_h is due to the model size exceeding the amount of memory MPS has available. From the [segment-anything paper](https://arxiv.org/abs/2304.02643):
- `vit_b` (base) has 91M parameters
- `vit_l` (large) has 308M parameters
- `vit_h`(huge) has 636M parameters

I don't know if we should consider making the default model type `vit_l` (instead of `vit_h`) in micro-sam.
Pros:
- We avoid this slowdown for users with MPS
- It's a smaller model, so it's a bit faster overall for all users (regardless of whether they use cpu/cuda/mps)
Cons:
- Currently there is no fine-tuned `vit_l` model for EM or LM.
- I don't know if the performance is worse or not. (Did you or one of your colleagues assess this, as part of the work before you started work on fine-tuning models? You might already know 😄 )

Possible alternatives: 
- We could warn the user if they have both MPS and `vit_h`/`vit_h_em`/`vit_h_lm` selected.

GenevieveBuckley:
Ok, to make a much fairer benchmark comparison, I've just re-started my computer and closed everything except python.
I have a first generation M1 macbook pro with 16GB of RAM (Ventura 13.5.1 (22G90)).

If nothing else is running simultaneously, then I don't have the same memory problem with `vit_h`. I'll open a separate issue where we can discuss whether to warn users if they might be in danger of running out of memory with MPS.

### rle branch
| model   | device   | benchmark             |    runtime |
|:--------|:---------|:----------------------|-----------:|
| vit_h   | mps      | embeddings            |  7.54463   |
| vit_h   | mps      | prompt-p1n0           |  0.035074  |
| vit_h   | mps      | prompt-p2n4           |  0.0320771 |
| vit_h   | mps      | prompt-box            |  0.0284541 |
| vit_h   | mps      | prompt-box-and-points |  0.03319   |
| vit_h   | mps      | amg                   | 33.2734    |
|---------|----------|-----------------------|------------|
| vit_h   | cpu      | embeddings            | 13.037     |
| vit_h   | cpu      | prompt-p1n0           |  0.0410733 |
| vit_h   | cpu      | prompt-p2n4           |  0.0386727 |
| vit_h   | cpu      | prompt-box            |  0.0387678 |
| vit_h   | cpu      | prompt-box-and-points |  0.040426  |
| vit_h   | cpu      | amg                   | 35.8723    |

### current dev branch

| model   | device   | benchmark             |    runtime |
|:--------|:---------|:----------------------|-----------:|
| vit_h   | mps      | embeddings            |  7.44625   |
| vit_h   | mps      | prompt-p1n0           |  0.035424  |
| vit_h   | mps      | prompt-p2n4           |  0.0320592 |
| vit_h   | mps      | prompt-box            |  0.028363  |
| vit_h   | mps      | prompt-box-and-points |  0.0332661 |
| vit_h   | mps      | amg                   | 76.1841    |
|---------|----------|-----------------------|------------|
| vit_h   | cpu      | embeddings            | 13.0751    |
| vit_h   | cpu      | prompt-p1n0           |  0.0397439 |
| vit_h   | cpu      | prompt-p2n4           |  0.0428631 |
| vit_h   | cpu      | prompt-box            |  0.0406017 |
| vit_h   | cpu      | prompt-box-and-points |  0.0418942 |
| vit_h   | cpu      | amg                   | 37.8754    |

### Summary
* Your new rle (run length encoding) implementation makes a huge difference, this is fantastic 🎉 
* The embeddings are a fair bit faster with mps (vs cpu). The 2D image examples use a reasonable amount of memory, and I haven't encountered any massive slowdowns due to exceeding the available memory. It seems clear that users should prefer to use MPS instead of cpu for this wherever possible.
* The amg (automatic mask generation) uses a LOT of memory with the mps backend, which sometimes causes huge slowdowns (performance 3.5 times slower than cpu). When you do have enough memory available, the mps backend is only about 2 seconds faster than the cpu computation. The cost/benefit here is not so clear - you have the possibility of a small speedup, but more likely you might accidentally run out of memory and cause a massive slowdown instead. I would probably prefer to use cpu instead of MPS in these cases.
* Using smaller models does not drop the amount of memory used by the amg benchmark by as much as I expected. They are all quite big memory hogs (although I only ever saw the huge slowdowns for the `vit_h` model).
* Using pytorch vs pytorch-nightly makes no difference to the benchmark timings.

constantinpape:
Thanks for looking so thoroughly into this @GenevieveBuckley!
A few comments:
- A reason for the large memory requirement in the AMG are that many masks need to be processed (64 x 64 x 3), where each mask corresponds to a (binary) array that has the image shape. To improve this we could:
  - Profile the memory usage separately and see if there is some other function we could vendor to improve it.
  - Discard unlikely masks early.
  - (I will create an issue about this later; I would handle this independent of the current PR as it is not directly related to MPS)
- Regarding `vit_l`: we have not benchmarked it yet, to keep the experimental overhead a bit limited we only did `vit_b` and `vit_h` so far. But I want to extend the benchmarks to `vit_l` as well once we also compare to the new smaller models (see #186). For now I would leave `vit_h` as the default, and we can revisit this decision later once more data is available.
 - Using different backends for AMG versus the rest: this would complicate the design a bit. For now, I would just leave things as they are and use MPS as default as available, and see if there is an easy fix to the memory problem (see also the point above).
 
 In summary: I will go ahead and merge this :).

micro_sam/_vendored.py
constantinpape: I didn't see that before merging; I will go ahead and fix it up on `dev` directly ;).

PR #190, opened 2023-09-13
Implement RLE computation with numba

constantinpape:
Unfortunately the current attempt is slower than the previous implementation:

**numba implementation:**

| model   | device   | benchmark   |   runtime |                                                                                                        
|:--------|:---------|:------------|----------:|                                                                                                        
| vit_h   | cpu      | amg         |   95.0089 | 
                                                                                                       
**previous implementation:**
     
| model   | device   | benchmark   |   runtime |                                                                                             
|:--------|:---------|:------------|----------:|                                                                                                        
| vit_h   | cpu      | amg         |   80.4364 |

### DIFF
--- a/development/benchmark.py
+++ b/development/benchmark.py
@@ -51,8 +51,9 @@ def _add_result(benchmark_results, model_type, device, name, runtimes):
     return benchmark_results
 
 
-def benchmark_embeddings(image, predictor, n=3):
+def benchmark_embeddings(image, predictor, n):
     print("Running benchmark_embeddings ...")
+    n = 3 if n is None else n
     times = []
     for _ in range(n):
         t0 = time.time()
@@ -62,8 +63,9 @@ def benchmark_embeddings(image, predictor, n=3):
     return ["embeddings"], [runtime]
 
 
-def benchmark_prompts(image, predictor, n=10):
+def benchmark_prompts(image, predictor, n):
     print("Running benchmark_prompts ...")
+    n = 10 if n is None else n
     names, runtimes = [], []
 
     embeddings = util.precompute_image_embeddings(predictor, image)
@@ -143,8 +145,9 @@ def benchmark_prompts(image, predictor, n=10):
     return names, runtimes
 
 
-def benchmark_amg(image, predictor, n=1):
+def benchmark_amg(image, predictor, n):
     print("Running benchmark_amg ...")
+    n = 1 if n is None else n
     embeddings = util.precompute_image_embeddings(predictor, image)
     amg = instance_seg.AutomaticMaskGenerator(predictor)
     times = []
@@ -164,6 +167,7 @@ def main():
     parser.add_argument("--benchmark_embeddings", "-e", action="store_false")
     parser.add_argument("--benchmark_prompts", "-p", action="store_false")
     parser.add_argument("--benchmark_amg", "-a", action="store_false")
+    parser.add_argument("-n", "--n", type=int, default=None)
 
     args = parser.parse_args()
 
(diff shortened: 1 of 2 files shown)


constantinpape:
I have reimplemented this in C++ and this removes the performance drawback:

**C++ implementation:**
                                                                                                            
| model   | device   | benchmark   |   runtime |                                                                                                        
|:--------|:---------|:------------|----------:|                                                                                                        
| vit_b   | cpu      | amg         |   78.9481 |

GenevieveBuckley:
Should this PR be opened on the master branch, or on dev?

constantinpape:
> Should this PR be opened on the master branch, or on dev?

Yes, this should go on `dev`. I opened it on `master` as a mistake. I changed the base branch to `dev`. (TIL that this is actually possible for an active PR and that it's not necessary to close and open a new one :) )

GenevieveBuckley:
> Unfortunately the current attempt is slower than the previous implementation:

That's interesting, I don't see the same results.

On my machine, the rle-numba branch is slightly faster than the current dev branch. 
But the prefilter amg branch with the vit_h model did exceed the available 16GB memory, which is why I think we see a slowdown with the mps backend there. I imagine if more memory was available it wouldn't be slower.

3 iterations of the amg benchmark function, average runtime reported:
| model   | device   | branch         | benchmark   |   runtime |
|:--------|:---------|:---------------|:------------|----------:|
| vit_h   | cpu      | dev            | amg         |   37.7422 |
| vit_h   | mps      | dev            | amg         |   33.8933 |
| vit_h   | cpu      | rle-numba      | amg         |   35.9172 |
| vit_h   | mps      | rle-numba      | amg         |   33.4212 |
| vit_h   | cpu      | prefilter-amg  | amg         |   33.2118 |
| vit_h   | mps      | prefilter-amg  | amg         |   43.8554 | 
| vit_h   | cpu      | numba+prefilter| amg         |   34.6391 |
| vit_h   | mps      | numba+prefilter| amg         |   42.9275 |

constantinpape:
> > Unfortunately the current attempt is slower than the previous implementation:
> 
> That's interesting, I don't see the same results.
> 
> On my machine, the rle-numba branch is slightly faster than the current dev branch. But the prefilter amg branch with the vit_h model did exceed the available 16GB memory, which is why I think we see a slowdown with the mps backend there. I imagine if more memory was available it wouldn't be slower.

Thanks for looking into it so thoroughly! Indeed some of this is quite surprising, especially that the memory requirements would**increase** with prefiltering. Re advantage of numba: maybe this depends on the CPU architecture because numba makes use of vector instructions?!

GenevieveBuckley:
> Re advantage of numba: maybe this depends on the CPU architecture because numba makes use of vector instructions?!

Yeah, maybe! Can you check again if you still see speed differences between `rle-numba`, `dev` and `master`? Maybe wishful thinking, but I'm kind of hoping the original 15 second difference you saw was a fluke, not what might happen every time.

constantinpape:
Hi @GenevieveBuckley,

I tried again, and I think you're right `numba` was only slower for me because of a fluke. (This is now for n=5, and also includes additional parallelization for nifty, but as you can see that does not really help).  

| model   | device   | benchmark   | implementation |   runtime |
|:----------|:------------|:-----------------|:---------------------|-------------:|
| vit_b    | cpu         | amg             | numba               |  73.2059 |
| vit_b    | cpu         | amg             | numpy               |  74.4526 |
| vit_b    | cpu         | amg             | nifty                   |  77.6681 |
| vit_b    | cpu         | amg             | nifty-parallel      |  75.3288 |

Based on this I will go ahead and merge with numba as the default. And I think we have exceeded all the more or less easy optimizations now!

PR #191, opened 2023-09-13
Filter unlikely masks in AMG

constantinpape:
To reduce the memory footprint. It should not affect the runtime.


constantinpape:
This is outdated.

PR #192, opened 2023-09-14
Refactoring Iterative Training Scheme

anwai98:
WIP : Refactor the training scheme in the `SamTrainer`, and update the iterative prompted inference

### DIFF
--- a/micro_sam/prompt_generators.py
+++ b/micro_sam/prompt_generators.py
@@ -197,7 +197,7 @@ class IterativePromptGenerator:
     """
     def _get_positive_points(self, pos_region, overlap_region):
         positive_locations = [torch.where(pos_reg) for pos_reg in pos_region]
-        # we may have objects withput a positive region (= missing true foreground)
+        # we may have objects without a positive region (= missing true foreground)
         # in this case we just sample a point where the model was already correct
         positive_locations = [
             torch.where(ovlp_reg) if len(pos_loc[0]) == 0 else pos_loc
@@ -258,16 +258,12 @@ class IterativePromptGenerator:
         self,
         gt: torch.Tensor,
         object_mask: torch.Tensor,
-        current_points: torch.Tensor,
-        current_labels: torch.Tensor
     ) -> Tuple[torch.Tensor, torch.Tensor]:
         """Generate the prompts for each object iteratively in the segmentation.
 
         Args:
             The groundtruth segmentation.
             The predicted objects.
-            The current points.
-            Thr current labels.
 
         Returns:
             The updated point prompt coordinates.
@@ -278,7 +274,7 @@ class IterativePromptGenerator:
 
         true_object = gt.to(device)
         expected_diff = (object_mask - true_object)
-        neg_region = (expected_diff == 1).to(torch.float)
+        neg_region = (expected_diff == 1).to(torch.float32)
         pos_region = (expected_diff == -1)
         overlap_region = torch.logical_and(object_mask == 1, true_object == 1).to(torch.float32)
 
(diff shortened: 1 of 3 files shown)


anwai98:
With the recent updates (post necessary checks while training the livecell model), looks like the iterative prompting is working as expected.

micro_sam/prompt_generators.py
```diff
@@ -230,13 +229,13 @@ def _get_negative_points(self, negative_region_batched, true_object_batched, gt_
...
+                custom_df = 3  # custom dilation factor to perform dilation by expanding the pixels of bbox
+                bbox_mask[max(bbox[0] - custom_df, 0): min(bbox[2] + custom_df, gt.shape[-1]),
+                          max(bbox[1] - custom_df, 0): min(bbox[3] + custom_df, gt.shape[-2])] = 1
                 bbox_mask = bbox_mask[None].to(device)
 
-                # NOTE: FIX: here we add dilation to the bbox because in some case we couldn't find objects at all
-                # TODO: just expand the pixels of bbox
-                dilated_bbox_mask = dilation(bbox_mask[None], torch.ones(3, 3).to(device)).squeeze(0)
-                background_mask = abs(dilated_bbox_mask - true_object)
+                background_mask = abs(bbox_mask - true_object)
```
constantinpape, line 238: Since these are torch tensors I would use https://pytorch.org/docs/stable/generated/torch.abs.html here.

micro_sam/training/sam_trainer.py
```diff
@@ -244,57 +238,14 @@ def _update_masks(self, batched_inputs, y, sampled_binary_y, sampled_ids, num_su
...
-                net_coords.append([pos_coordinates, neg_coordinates])
-                net_labels.append([pos_labels, neg_labels])
-
-            if "point_labels" in _inp.keys():
-                updated_point_coords = torch.cat([_inp["point_coords"], torch.tensor(net_coords)], dim=1)
-                updated_point_labels = torch.cat([_inp["point_labels"], torch.tensor(net_labels)], dim=1)
-            else:
-                updated_point_coords = torch.tensor(net_coords)
-                updated_point_labels = torch.tensor(net_labels)
+            iterative_prompter = IterativePromptGenerator()
```
constantinpape, line 242: This should be defined on the class level rather than here, so that we enable different training schemes by exposing this.
So something like:
```python
class SamTrainer:
  def __init__(self, ..., prompt_generator=IterativePromptGenerator(), ...):
    self.prompt_generator = prompt_generator
    
  def _get_updated_points_per_mask_per_subiter(self, ...):
    ...
    net_coords, net_label = self.prompt_generator(x2, x1)
    ...
```

constantinpape:
Something else I noticed while going through this: We do not need a `torch.unique` here:
https://github.com/computational-cell-analytics/micro-sam/blob/master/micro_sam/training/sam_trainer.py#L308

It is sufficient to use `torch.amax(y, dim=(1, 2, 3))` (assuming we have a 4d tensor here, it's probably better to determine this from the data, e.g. `dim=tuple(range(1, y.ndim + 1))`).
Note that we can assume that the groundtruth is labeled consecutively from 0 to max_id. (The training won't work otherwise.)

constantinpape:
@anwai98 besides the minor points I raised this looks good to go.

> With the recent updates (post necessary checks while training the livecell model), looks like the iterative prompting is working as expected.

Does that mean you checked the training already and that it works?

micro_sam/training/sam_trainer.py
anwai98: Ahha yes. I see your point. We still need to check out the intersection of `n_sub_iterations` and different prompting strategy (iterative or classical point and box), but maybe we do this in the next step(s).

constantinpape: You're right about `n_sub_iterations`, but let's not worry about this for now.

anwai98:
Thanks Constantin for the feedback!

> Does that mean you checked the training already and that it works?

Yes, I checked it before the current version. I will recheck once again before I request for review.

constantinpape (approved):

Issue #185, opened 2023-09-08
Implement more annotation layers for complex shapes?


constantinpape:
It can be difficult to annotate objects with complex shapes and/or non axis aligned orientation with the box layer (which yields better results than point prompts in most cases).

We could potentially implement better annotation functionality for these cases by using [mask prompts](https://github.com/computational-cell-analytics/micro-sam/blob/master/micro_sam/prompt_based_segmentation.py#L303).
There are different options for how to do this:
- Use a `label layer` and derive the mask prompt from the user annotation in that layer.
- Use a `shape layer` and derive the mask from a polygon (or other) annotation.
- (Another alternative would be to use polygons and train a new prompt encoder for this. This is interesting, but would need much more work).

Implementing prototypes for the first two options is fairly straight forward, but designing the user interface and testing it would be some work. 

cc @GenevieveBuckley

constantinpape:
Hi @GenevieveBuckley, 
I have merged #196 now and it's working for all annotation tools. The only thing missing is to disable the 'Line' and 'Path' buttons in the shape layer (I checked https://forum.image.sc/t/disable-buttons-on-shapes-layer/55243/10 and this seems to be quite involved).
Let me know what you find for your 3d data with complex shapes!

constantinpape:
Just pinging you again here @GenevieveBuckley (since you were not aware that it works for 3d etc. already earlier).
I have merged this on dev, so this is ready to be tested in any setting.
I have not checked how well this works for complex shapes myself yet, so would be great to get some feedback from you, and see if this could be further improved.

PR #196, opened 2023-09-19
Implement prototype for segmentation of complex shapes

constantinpape:
I started to implement a prototype for the segmentation of complex shapes, see #185.

I thought about this, and the easiest way to implement this would just be to use the `box_prompt` layer and enable segmentation from `polygon` and `ellipse` shapes in addition to `rectangle`. I think this is the best solution both in terms of implementation (it's fairly simple) and also UI (we don't need to add any new layers).
See how it works:

https://github.com/computational-cell-analytics/micro-sam/assets/4263537/8a7067e4-c381-4651-a336-9022c654d7d3

The current status: it's implemented only for the 2d annotation tool, but would be easy to enable for the volumetric segmentation and tracking tool as well.

@GenevieveBuckley would be great if you can give this a try at some point and see if this works for your data with complex shapes.

If we decide to go ahead with this I still need to:
- enable this for 3d and tracking annotator
- check that tiled segmentation also works with this (I also had to do one change that might affect this with box prompts)
- maybe: disable the `path` and `line` options in the `box_prompt` shape layer (because it cannot be used for prompts, but I am not quite sure if it's simple to do this with the napari shape layer)
- maybe: rename the "box_prompts" layer (because we now also support the other shapes)

### DIFF
--- a/micro_sam/prompt_based_segmentation.py
+++ b/micro_sam/prompt_based_segmentation.py
@@ -27,24 +27,8 @@ def _compute_box_from_mask(mask, original_size=None, box_extension=0):
     coords = np.where(mask == 1)
     min_y, min_x = coords[0].min(), coords[1].min()
     max_y, max_x = coords[0].max(), coords[1].max()
-
-    if box_extension == 0:  # no extension
-        extension_y, extension_x = 0, 0
-    elif box_extension >= 1:  # extension by a fixed factor
-        extension_y, extension_x = box_extension, box_extension
-    else:  # extension by fraction of the box len
-        len_y, len_x = max_y + 1 - min_y, max_x + 1 - min_x
-        extension_y = int(np.round(box_extension * len_y))
-        extension_x = int(np.round(box_extension * len_x))
-
-    box = np.array([
-        max(min_x - extension_x, 0), max(min_y - extension_y, 0),
-        min(max_x + 1 + extension_x, mask.shape[1]), min(max_y + 1 + extension_y, mask.shape[0]),
-    ])
-    if original_size is not None:
-        trafo = ResizeLongestSide(max(original_size))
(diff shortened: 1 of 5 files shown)


GenevieveBuckley:
First impressions: it works great on `examples/annotator_2d.py`. 
I tried out the polygon tool, that other polygon tool where you click to add the points, the triangle, and the ellipse shapes.

I've found it helps to get your line close to the edge of the object you want to segment. Too far inside or outside give you poorer results (understandable, and I didn't find it too annoying - there is a little bit of flexibility & you don't have to be too precise)

GenevieveBuckley:
I agree with the next steps you outline. I personally care most about 3D volumes right now (I have a great dataset I'm working on where I can absolutely see this saving me a lot of time)

Re prompt layer names: we could have "prompts" and "point_prompts"?

constantinpape:
Thanks for testing @GenevieveBuckley. I will go ahead and finish this later, and then you can properly test it for your 3d data.

> I've found it helps to get your line close to the edge of the object you want to segment. Too far inside or outside give you poorer results (understandable, and I didn't find it too annoying - there is a little bit of flexibility & you don't have to be too precise)

In these cases it might help to increase the `box_extension` parameter.
Briefly, we use both the mask and bounding box derived from the polygon or ellipse annotation tool as prompts to the model. This bounding box will usually be fairly tight, and SAM works better when bounding boxes are a bit larger; so the extension factor enlarges it by the given fraction of the object lengths.

> Re prompt layer names: we could have "prompts" and "point_prompts"?

Yes, I think these are the most straight forwards names. I will change it!

constantinpape:
Regarding disabling the unwanted parts of the shape layer: https://forum.image.sc/t/disable-buttons-on-shapes-layer/55243/10

Issue #195, opened 2023-09-18
Remove clashes with napari built-in keybindings


GenevieveBuckley:
I've noticed that some of the keybindings in micro-sam overlap with the built-in keybindings in napari. This makes micro-sam inconvenient to use. We should re-work the keybindings to avoid clashes wherever possible.

There is a list of (most? all?) napari keybindings in the preferences menu: `napari` > `Preferences` > `Shortcuts`
You should also be able to open the keybindings preference menu using the shortcut: `Control` + `C` (or `Command` + `C`)

The specific example I found was that the `V` keybinding, which usually toggles visibility of the selected layer(s) in napari, is overwritten by micro-sam to segment volumes with the 3D annotator. I often want to toggle the label layer off and on so I can see how well I think the segmentation matches the original image data, and not being able to use the `V` shortcut for this makes things much more inefficient.

constantinpape:
Thanks for bringing this up, that's a very good point. We should fix this while working on #167.

I will have a closer look at the details tomorrow.

constantinpape:
I checked the keybindings, and I think `V` is the only one that clashes with an existing one.
We should def. change that (and given the updated plans we discuss can do that before #167).

We have two options:
- We don't provide a key binding, since volume segmentation / tracking takes longer anyways, so having a key binding doesn't change that much time.
- Switch to some other key binding. I would suggest `Ctrl + V` or `Shift + V`.

What do you think @GenevieveBuckley?

GenevieveBuckley:
True, I guess we could just not have a keybinding for it. Still, it might be nicer if we could think of a good keybinding.

I don't think we should use `Ctrl + V`, since that's already associated closely with copy-paste commands.
Maybe `Shift + V`, or `Shift + S` (since we already have `S` for segmenting one slice, this might be a good extension to segmenting all the slices). I'll have a think.

constantinpape:
This is now resolved!

PR #207, opened 2023-09-25
Change keybinding for volume segmentation (remove clash with napari layer visibility keybinding)

GenevieveBuckley:
Closes https://github.com/computational-cell-analytics/micro-sam/issues/195

Now:
* `S` <- segment a single slice (same as before)
* `Shift-S` <- segment all slices (previously `V` for volume segmentation)

I like `S` / `Shift-S` for segment one slice / segment all slices. 
I feel this pattern matches keybinding pattern we see in other places like the points layer, which has `A` / `Shift-A` for select all points in this slice / select all points in the whole volume.

Other alternatives we considered:
* `V` <- our current situation, is no good because this clashes with 
* `Control-V` <- also clashes with napari built in keybindings
* `Shift-V`, `Alt-V`, etc. <- might have worked, but seemed less easy to remember than the `S` / `Shift-S` combination.

Note: something weird has happened with the napari built in keybinding. At some point it looks like the `V` keybinding was accidentally changed to `G`. I don't think this affects actual released versions of napari, but I thought I'd mention it in case you do see it. https://github.com/napari/napari/pull/6261

### DIFF
--- a/micro_sam/sam_annotator/annotator_3d.py
+++ b/micro_sam/sam_annotator/annotator_3d.py
@@ -104,7 +104,7 @@ def _segment_volume_for_auto_segmentation(
 
 
 @magicgui(
-    call_button="Segment Volume [Shift-S]",
+    call_button="Segment All Slices [Shift-S]",
     layer={"choices": ["current_object", "auto_segmentation"]},
     projection={"choices": ["default", "bounding_box", "mask", "points"]},
 )
--- a/micro_sam/sam_annotator/annotator_3d.py
+++ b/micro_sam/sam_annotator/annotator_3d.py
@@ -104,7 +104,7 @@ def _segment_volume_for_auto_segmentation(
 
 
 @magicgui(
-    call_button="Segment Volume [V]",
+    call_button="Segment Volume [Shift-S]",
     layer={"choices": ["current_object", "auto_segmentation"]},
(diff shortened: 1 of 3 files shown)


micro_sam/sam_annotator/annotator_3d.py
```diff
@@ -104,7 +104,7 @@ def _segment_volume_for_auto_segmentation(
 
 
 @magicgui(
-    call_button="Segment Volume [V]",
+    call_button="Segment All Slices [Shift-S]",
```
GenevieveBuckley, line 107: I've changed the language here from "Segment Volume" to "Segment All Slices" because I felt that might make it easier to remember the keybindings. I hope it is unlikely to cause confusion for users, what do you think?

constantinpape (commented):
I agree that `s` / `shift-s` makes a lot of sense to have as keybindings here!
Could you also change it in `annotator_tracking` for `_track_object_widget`?
Thanks!

GenevieveBuckley:
Done!
Keybinding updated to `Shift-S` for volume and tracking annotation. I have also updated the corresponding markdown documentation for this.

micro_sam/sam_annotator/annotator_3d.py
constantinpape: Yes, I think that's fine.

constantinpape (approved):

PR #211, opened 2023-09-25
Fix prompt gens

constantinpape:


anwai98:
Re: shape of `__call__` for the `IterativePromptGenerator` - 
Yes, for both the "segmentation" and "prediction", the expected shapes are `num_objects` X 1 X H X W, where `num_objects` is the number of objects chosen per sample per batch.

constantinpape:
Thanks for the clarification @anwai98 !

micro_sam/prompt_generators.py
```diff
@@ -138,12 +138,14 @@ def _sample_negative_points(self, object_mask, bbox_coordinates, coord_list, lab
         # getting the negative points
         # for this we do the opposite and we set the mask to the bounding box - the object mask
         # we need to dilate the object mask before doing this: we use scipy.ndimage.binary_dilation for this
-        # TODO upldate this to use dilation compatible with pytorch (either use dilation or distance trafo from kornia)
-        dilated_object = binary_dilation(object_mask, iterations=self.dilation_strength)
+        dilated_object = morphology.dilation(object_mask[None, None],
+                                             torch.ones(self.dilation_strength, self.dilation_strength))
+        dilated_object = dilated_object.squeeze().numpy()
```
constantinpape, line 143: I wouldn't cast back to numpy here. That's potentially unnecessary data moving between GPU and CPU.
If you also do
```
background_mask = torch.zeros(object_mask.shape, device=object_mask.device)
```
then everything below should more or less still work. (With some other minor changes, like the `dtype` casts in line 150).

anwai98: Yes, this makes senses (keeping the data in one device and moving ahead with tensors). Took care of this. Thanks for mentioning, Constantin!

micro_sam/prompt_generators.py
```diff
@@ -137,21 +137,20 @@ def _sample_negative_points(self, object_mask, bbox_coordinates, coord_list, lab
...
         # getting the negative points
         # for this we do the opposite and we set the mask to the bounding box - the object mask
-        # we need to dilate the object mask before doing this: we use scipy.ndimage.binary_dilation for this
-        dilated_object = morphology.dilation(object_mask[None, None],
-                                             torch.ones(self.dilation_strength, self.dilation_strength))
-        dilated_object = dilated_object.squeeze().numpy()
+        # we need to dilate the object mask before doing this: we use kornia.morphology.dilation for this
+        dilated_object = object_mask[None, None]
+        for _ in range(self.dilation_strength):
+            dilated_object = morphology.dilation(dilated_object, torch.ones(3, 3))
```
constantinpape, line 143: A final thing: Can you please benchmark how long this takes:
- With the default engine ("unfold").
- And `engine="convolution"`

And also see if these results are significantly different from each other. The kornia documentation claims that `"convolution"` is faster and less memory hungry, so it may be a good idea to use it.

anwai98: Yes, the claims in the documentation are correct.

The `engine="convolution` is twice as fast (per object) as compared to the default engine (`"unfold"`). I have updated it in the latest commit.

PR #221, opened 2023-10-04
Fix Box Prompting in Inference

anwai98:
- Update box-prompting in inference (by fixing the inputs of `use_boxes` condition to `input_boxes`)


constantinpape (approved):

constantinpape:
The test failure looks random, so I will merge anyways. We should add this to the regression tests, I will do that later.

PR #222, opened 2023-10-04
More comprehensive .gitignore file at repository root

GenevieveBuckley:
This PR:
1. Adds the Github [python gitignore template](https://github.com/github/gitignore/blob/main/Python.gitignore) contents to our .gitignore file
2. Adds a few extra files to the gitignore:
    * generated files related to code coverage measurements
    * .DS_Store files (I often work on a Mac)
    * any embeddings folders

I have not removed anything from the .gitignore (but I have made sure there are no duplicate entries, eg: `__pycache__`, etc.)
I also see that there are .gitignore files in subfolders of the repository, and I have not touched these at all either.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,175 @@
-__pycache__/
-*.egg-info/
+# Bulky files
 *.pth
 *.tif
 examples/data/*
 *.out
+*embeddings/*
+
+# Code coverage files
+coverage.xml
+.coverage*
+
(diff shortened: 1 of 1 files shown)


constantinpape (approved):
Thanks! And I don't think we need to touch the `.gitignore` files in the subfolders, they are mostly there to not add the data that is automatically downloaded or created by the example scripts.

Issue #205, opened 2023-09-22
Enable GUI tests in CI


constantinpape:
It would help with all the refactoring tasks to enable tests for the GUI. Currently this is not possible, becaue napari cannot be imported in the Github CLI due to missing graphical libraries (QT, but might lack even more things like OpenGL).

The best strategy here would probably look into what napari does for the CI. (I tried this myself at some point, but gave up because the napari test suite is quite extensive and it was unclear what the best starting point is).

GenevieveBuckley:
## Testing 

* For reference, link to the documentation page on testing napari plugins: https://napari.org/stable/plugins/test_deploy.html
* We will need to switch to pytest (instead of the current unittest). We need both `pytest` and `pytest-qt` to handle tests involving the GUI. Fortunately, pytest can already run the current unit tests with no modification, so there is no extra work needed there (the existing tests are already compatible with pytest).

## Github actions CI
The complications around GUI testing in the GitHub actions CI can be solved by adding two actions to the workflow yaml file:
* [tlambert03/setup-qt-libs](https://github.com/tlambert03/setup-qt-libs), and
* [aganders3/headless-gui](https://github.com/aganders3/headless-gui)

## Setup Qt libraries
First, to set up the Qt graphical libraries on the remote CI machine, we can use Talley Lambert's [tlambert03/setup-qt-libs](https://github.com/tlambert03/setup-qt-libs) action.

[Click here to see the exact line where it is used in the napari workflow file](https://github.com/napari/napari/blob/bdd5cb453df497529531ed0277d477dd8437ecef/.github/workflows/test_pull_requests.yml#L118)
```yaml
- uses: tlambert03/setup-qt-libs@v1
```

## Run headless GUI during testing
Then, in the step where the tests are run, we can use Ashley Anderson's [aganders3/headless-gui](https://github.com/aganders3/headless-gui) action. This gives us a virtual display for the tests.

[Click here to see the exact line where it is used in the napari workflow file](https://github.com/napari/napari/blob/bdd5cb453df497529531ed0277d477dd8437ecef/.github/workflows/test_pull_requests.yml#L144)
```yaml
- uses: aganders3/headless-gui@v1
```

### Workflow YAML file
We'll use [this section of the napari yaml workflow](https://github.com/napari/napari/blob/bdd5cb453df497529531ed0277d477dd8437ecef/.github/workflows/test_pull_requests.yml#L56-L163) to base ours on. It can likely be simplified a lot for our use.

[...]

PR #214, opened 2023-09-27
CI testing - automatic code coverage reports

GenevieveBuckley:
This PR investigates adding code coverage reports to the CI.

Included in this PR:
* Switch to pytest (previously unittest was used)
* Generate code coverage reports (using pytest-cov)
* Uploads code coverage to codecov.io
* Reduces duplication of CI runs, by adjusting the triggers for the github actions workflow
* Switches to micromamba for the CI (instead of miniconda combined with mamba). This creates our test environment in about half the time that miniconda did. It saves about a quarter of the overall CI job time (there aren't a lot of tests right now, so the environment setup is a large proportion of the CI runtime)

When this PR is ready to merge:
1. We'll need to add a secret key to the repository, so the upload to codecov.io can happen automatically, and
2. We should absolutely consider squash merging this PR, the commit history is very messy

PR #223, opened 2023-10-04
Simple GUI test for CI

GenevieveBuckley:
Closes https://github.com/computational-cell-analytics/micro-sam/issues/205

This PR:
1. Updates the github test.yaml workflow file to enable GUI testing on CI (as [described here](https://github.com/computational-cell-analytics/micro-sam/issues/205#issuecomment-1736683856)), and 
2. Adds a very simple GUI test which opens the 2D annotator dock widgets in napari. We don't interact with it or do any actual testing of the annotation functions. This GUI test just creates the image embedding and then opens the dock widgets in the napari viewer, as if it was ready to start.

That brings our [code coverage up to 23%](https://app.codecov.io/github/GenevieveBuckley/micro-sam/commit/58aeaa7978da616d6240ed42b222680f0361145c/tree). I did do some extra work (not in this branch) to see if I could mimic user interaction of annotation to this test... but that only increased the test coverage by about 2%. My guess is that these functions might already be covered by unit tests, so it's not helpful to make a giant integration test anyway.

**Careful**: this PR is just added straight on top of the many, many git commits involved in branch https://github.com/computational-cell-analytics/micro-sam/pull/214. We should do, uh, something with git to fix any problems from that. (Worst case scenario, I can close and and reopen a new PR for this after https://github.com/computational-cell-analytics/micro-sam/pull/214 is merged.)


GenevieveBuckley:
I really do not like how the CI job takes <5 minutes on the `codecov-report` PR branch, but on this branch it is taking over half an hour.
Trying to figure out why it's so slow:
* I don't think it's installing pyqt and napari, all the steps before we run pytest finish in a reasonable amount of time
* The `pytest` step is very, very slow!
    * It's not because my new GUI test is slow - I commented that out in one of the commits, but still found the CI job took an unreasonably long time to run
    * ...I think it might be the use of `aganders3/headless-gui`? We could test this theory by replacing `pytest` with `echo hello` and then look at how long each CI step took to run. I'm not sure what to do if this is the reason the CI is slow, I think we need it for GUI tests.

GenevieveBuckley:
Update: I think it's the `test/test_training.py` file that was making the CI so slow. I accidentally opened this PR against the master branch, not dev (not that it helps much, because everything needs to work in master eventually)

A possible workaround would be to mark the super slow tests as `pytest.mark.slow` and skip them in the CI jobs.

constantinpape:
> Update: I think it's the `test/test_training.py` file that was making the CI so slow.

Yes, unfortunately `test_training.py` is quite slow.

> A possible workaround would be to mark the super slow tests as `pytest.mark.slow` and skip them in the CI jobs.

Yes, that's one possibility. Before that I want to check if I can improve this test a bit to speed it up on CI and so that we can just run it with the normal tests.
(I developed the test on my notebook, and it was fast enough, but I guess the CPU in the CI is weaker, so that it takes quite long.)

For now: since you now have the PR on dev it should not be a problem here. Can you go ahead and fix the merge conflicts (which I think are because I squashed and merged #214) and, if you have any more changes you want to make, add them too?

I will then:
- squash and merge this PR
- merge `dev` into `master` and also update `dev`, so that we have all the tests up-to-date and on the same branches
- look into speeding up `test_training.py` (or mark it as `pytest.mark.slow` if I can't speed it up without making the test worse)
- look into adding more GUI tests

.github/workflows/test.yaml
```diff
@@ -15,7 +15,7 @@ jobs:
   test:
     name: ${{ matrix.os }} ${{ matrix.python-version }}
     runs-on: ${{ matrix.os }}
-
+    timeout-minutes: 60
```
GenevieveBuckley, line 18: Right now, the CI job finishes in around six-ish minutes. On the master branch, including the `test_training.py` script, it takes over half an hour. Google says the default github actions timeout is 360 minutes (6 hours)... but we were seeing CI jobs killed with an error much sooner than that on the master branch (maybe they ran out of memory?)

I'm not sure whether we should keep this line, remove it, or change it to a different number. I'm leaning towards keeping it in here, and maybe changing the timeout number if that makes sense.

GenevieveBuckley:
I've fixed the merge conflicts, and opened a separate issue for discussion about `test_training.py`.

I'm very happy that this CI job runs in about 6 minutes - the gui test is not slow at all!

.github/workflows/test.yaml
constantinpape: Let's keep it for now. I will look into the test.

PR #228, opened 2023-10-13
Add LM Generalist Finetuning

anwai98:
- The training scripts for lm generalist datasets (still need to test it and add necessary docstrings)


anwai98:
This is good to go (tested it for a few epochs, works fine - and much faster).
The above PR also expects this [PR](https://github.com/constantinpape/torch-em/pull/161) from torch-em to be merged simultaneously (to add necessary info about NeurIPS CellSeg dataset)

constantinpape (approved):

Issue #203, opened 2023-09-22
Implement embedding widget


constantinpape:
so that a user can select an image layer in napari for which to compute the embeddings.
Note that this also relates to
- #201  (the initialized model, embeddings and image shape need to be passed into the global state)
- #202 (the model loading should be performed in a thread worker)

```python
from pathlib import Path
from typing import Optional

import napari
from napari.layers import Image
from magicgui import magicgui


@magicgui(call_button="Compute embeddings")
def embedding_widget(
    image: Image,
    model_type: str = "vit_h",  # we should use micro_sam.util._DEFAULT_MODEL, and provide a drop-down menu initialized with micro_sam.util.get_model_names
    device: Optional[str] = None,  # it would be good to determine the available devices beforehand, maybe part of the global state?
    model_source: Optional[str] = None,  # A filepath or URL to custom model weights.
    cache_path: Optional[Path] = None,  # where embeddings for this image are cached (optional)
    ndim: Optional[int] = None,  # to indicate whether the image data is 2d or 3d. Maybe we don't actually need this and can just derive this from whether the image is RGB or not.
):
    # The function needs to do the following things:
    # - Initialize the model (micro_sam.util.get_sam_model)
    # - Compute the embeddings for the image data (micro_sam.util.precompute_image_embeddings)
    #    - This already takes care of caching.
    # - Save the model, embeddings and shape of the image data in the global state
    pass


v = napari.Viewer()
v.window.add_dock_widget(embedding_widget)
napari.run()
```

GenevieveBuckley:
Ok, I had a strangely difficult time combining a thread worker with an indeterminate progress bar in the napari plugin widget. Here's what it should look like:

```python
from time import sleep
from magicgui import magic_factory
from magicgui.tqdm import tqdm


@magic_factory(
    threshold={"widget_type": "FloatSlider", "max": 20},
    call_button="Calculate!",
)
def example_magic_widget(
    img_layer: "napari.layers.Image",
    threshold: "float"
) -> "napari.types.LabelsData":
    with tqdm() as pbar:
        @thread_worker(connect={"finished": lambda: pbar.progressbar.hide()})
        def long_running():
            sleep(5)

        long_running()
```

constantinpape:
This has been done in #235

Issue #201, opened 2023-09-22
Refactor global state


constantinpape:
Currently, the global state is just kept in global variables, and each annotation tool has it's own set of global variables.
We should improve this design to enable building the plugins.

We have the following global state:
- The predictor / model
- The image embeddings
- State for the automatic mask generation
- State for the tracking tool (<- I want to refactor this, so we should not worry about this one for now)

And we probably need to add state for:
- The shape of the image layer (because we don't want to hard-code its name to `"raw"` any more)
- The available devices

We have discussed these options to represent this state:
- A singleton classes
- Adding it to the container widget (#200)
- Having it as part of the class that represents the plugin / tool (we did not discuss this directly, but if we were to choose a class based design it would make sense) (will open an issue on this later)

I currently think that the singleton would be the best solution, since it decouples this from the rest of the design.

PR #235, opened 2023-10-19
Embedding widget for napari plugin

GenevieveBuckley:
Closes https://github.com/computational-cell-analytics/micro-sam/issues/203
A simple widget to compute image embeddings via the micro-sam napari plugin.

I've put the widget stuff in a separate file for now. Not sure if that's the right long term choice or not, but we can always change it later.


micro_sam/sam_annotator/_widgets.py
```diff
@@ -0,0 +1,56 @@
...
+Model = Enum("Model", _MODEL_URLS)
+
+
+@magic_factory(call_button="Compute image embeddings",
+               device = {"choices": ["auto", 'cuda', 'mps', 'cpu']})
+def embedding_widget(
+    image: "napari.layers.Image",
+    model: Model = Model.__getitem__(_DEFAULT_MODEL),
+    device = "auto",
+    save_path: Optional[Path] = None,  # where embeddings for this image are cached (optional)
```
GenevieveBuckley, line 29: Are you sure you want "None" to be the default save path?
I could use something like `pooch.os_cache_dir("micro-sam")/"embeddings"` and use the `image.name` parameter to name the zarr files.

micro_sam/sam_annotator/_widgets.py
```diff
@@ -0,0 +1,56 @@
...
+        ndim = image.data.ndim
+    else:
+        # assumes RGB channels are the last dimension
+        ndim = image.data.ndim
+
+    # Compute the embeddings for the image data
+    with tqdm() as pbar:
+        @thread_worker(connect={"finished": lambda: pbar.progressbar.hide()})
+        def _compute_image_embedding(PREDICTOR, image, save_path, ndim=None):
+            IMAGE_EMBEDDINGS = precompute_image_embeddings(
```
GenevieveBuckley, line 49: I am not completely confident that this saves the image embeddings to the global variable.
I don't work with global variables much, perhaps someone else can help me check this?

micro_sam/sam_annotator/_widgets.py
```diff
@@ -0,0 +1,56 @@
...
+    global PREDICTOR, IMAGE_EMBEDDINGS
+    # Initialize the model
+    PREDICTOR = get_sam_model(model_type=model.name)
+
+    # Get image dimensions
+    if not image.rgb:
+        ndim = image.data.ndim
+    else:
+        # assumes RGB channels are the last dimension
+        ndim = image.data.ndim
```
GenevieveBuckley, line 43: Constantin said [here](https://github.com/computational-cell-analytics/micro-sam/issues/203#issue-1908560489) that the (spatial) image dimensions (`ndim`) should be saved to the global state. Is that true? 

I thought it seemed fine to just pass `ndim` into the function computing the embeddings, do we need it later on as well? It's possible I missed where that global variable is defined, can you clarify?

micro_sam/sam_annotator/_widgets.py
GenevieveBuckley: Specifically, it's the fact this is happening in a different thread that I'm worried about.

micro_sam/util.py
```diff
@@ -145,6 +142,36 @@ def _get_device(device):
     return device
 
 
+def _get_device(device=None):
```
GenevieveBuckley, line 145: I wanted better error handling, if users selected a device that wasn't available on their system.
Then I thought I'd go one step better, and only display the available devices int he embedding widget dropdown menu (using the new `_get_available_devices()` function).

micro_sam/visualization.py
```diff
@@ -3,8 +3,6 @@
 """
 from typing import Tuple
 
-from typing import Tuple
```
GenevieveBuckley, line 6: I don't know how this line was inserted here twice. I'm removing the duplicate.

micro_sam/sam_annotator/_widgets.py
```diff
@@ -0,0 +1,58 @@
...
+        @thread_worker(connect={"finished": lambda: pbar.progressbar.hide()})
+        def _compute_image_embedding(PREDICTOR, image, save_path, ndim=None):
+            IMAGE_EMBEDDINGS = precompute_image_embeddings(
+                predictor = PREDICTOR,
+                input_ = image,
+                save_path = str(save_path),
+                ndim=ndim,
+            )
+
+        _compute_image_embedding(PREDICTOR, image.data, save_path, ndim=ndim)
```
GenevieveBuckley, line 58: Something else to check - does this produce sensible results if I pass it an RGB image data array, and ndim = image.data.ndim - 1? I don't need to do anything fancy to the actual image data array, do I?

micro_sam/sam_annotator/_widgets.py
GenevieveBuckley: Also, this widget is giving us a file picker - I think we want a directory picker here instead.

micro_sam/sam_annotator/_widgets.py
GenevieveBuckley: When I give it an RGB image, like the astronaut example image, it seems to take a really long time (I've been closing the window after about a minute, maybe it never completes?)

In contrast, a grayscale image of the same size, like the camera example image, has the embedding computation finished in a few seconds.

micro_sam/sam_annotator/_widgets.py
constantinpape: This is potentially related to the issue you see with RGB. I will test this branch and follow up.

constantinpape: I think you can ignore [the comment](https://github.com/computational-cell-analytics/micro-sam/issues/203#issue-1908560489).
However, the code you have written here will give the wrong result (and I think explains the long runtime you observe with RGB data you observe below).
The `ndim` variable must be equal to the **number of spatial dimensions**. So in the case of an RGB image it must be two, but your code yields three.
I would suggest to change this code to:
```python
ndim = image.data.ndim
if image.rgb:
   ndim -= 1
```
(This would potentially even work for 3D RGB data. Which is something I have not tried, so not sure if it will work in the code that computes embeddings, but let's not worry about that now).

micro_sam/sam_annotator/_widgets.py
constantinpape: I am also not so sure about this. I would for sure add 
```python
global IMAGE_EMBEDDINGS
```
below the function signature, from my understanding of global variables that should ensure that the global variable is updated.

micro_sam/util.py
constantinpape: Looks good!

micro_sam/sam_annotator/_widgets.py
constantinpape: Yes, for RGB data it should be enough to set `ndim = image.data.ndim - 1`. See also the comment below. Once you fix that it should compute the embeddings for RGB images in a few seconds as well.
(Currently it interprets it as a 3d volume, and that's why it takes so long.)

test/test_sam_annotator/test_widgets.py
```diff
@@ -0,0 +1,31 @@
...
+
+from micro_sam.sam_annotator._widgets import embedding_widget, Model
+from micro_sam.util import _compute_data_signature
+
+
+# make_napari_viewer is a pytest fixture that returns a napari viewer object
+# you don't need to import it, as long as napari is installed
+# in your testing environment.
+# tmp_path is a regular pytest fixture.
+def test_embedding_widget(make_napari_viewer, tmp_path):
```
constantinpape, line 14: If possible we should also check that the `IMAGE_EMBEDDINGS` variable is updated here.

constantinpape (commented):
I had a closer look now and also tried to run the widget.

Two things I noticed:
- The `ndim` computation for RGB data is wrong.
- We should check that the global `IMAGE_EMBEDDINGS` variable is updated.

I also checked out the branch, reinstalled (via `pip install -e`) and tried to run the widget, but starting the widget from the plugins menu fails with:
```
RuntimeError: Failed to import command at 'micro_sam.sam_annotator._widgets:embedding_widget': /lib/x86_64-linux-gnu/libti
nfo.so.6: version `NCURSES6_TINFO_6.2.20211010' not found (required by /home/pape/software/conda/miniconda3/envs/sam/lib/p
ython3.10/site-packages/vigra/../../../././libncurses.so.6) 
```
@GenevieveBuckley : should this work already?

> I've put the widget stuff in a separate file for now. Not sure if that's the right long term choice or not, but we can always change it later.

I think that's good and makes more sense than `micro_sam.sam_annotator.util`, where I put these things before.

micro_sam/sam_annotator/_widgets.py
GenevieveBuckley: >> Also, this widget is giving us a file picker - I think we want a directory picker here instead.
>
> Yep, that would be better. Do you know how to do that or should we ping Talley?

Yes, I have made it a directory picker. Just had to add `save_path={"mode": "d"},  # choose a directory` to the `magic_factory` decorator, and it's working now.

GenevieveBuckley: >> Are you sure you want "None" to be the default save path?
>> I could use something like pooch.os_cache_dir("micro-sam")/"embeddings" and use the image.name parameter to name the zarr files.
>
> I am not so sure about this. I am a bit afraid we will take up quite a lot of disc space. I will check how big the embeddings actually get and we can revisit this before merging.

That is a reasonable concern. 
I think I agree with you about leaving it `None`. But that does make it extra-important that the EMBEDDINGS global variable is being set correctly from the thread worker - I'm not totally sure how to check that. Any ideas?

micro_sam/sam_annotator/_widgets.py
GenevieveBuckley: ... damn, I can't believe I wrote this whole if/else statement out and FORGOT THE MINUS ONE. I'll fix it now 😆

micro_sam/sam_annotator/_widgets.py
GenevieveBuckley: Ok, I'm going to move `global IMAGE_EMBEDDINGS` out from the `embedding_widget` function, and inside the `_compute_image_embedding` thread worker function. I think it must make more sense to have it in one place, not two. I'll see if there's a way I can check if this is working well.

GenevieveBuckley:
> @GenevieveBuckley : should this work already?

Almost. Our biggest problem is that the global IMAGE_EMBEDDINGS variable is not accessible outside our thread worker.

1. The error [you describe here](https://github.com/computational-cell-analytics/micro-sam/pull/235#pullrequestreview-1688359812) is odd, I've never seen that before. It almost looks like there is something wrong with your vigra installation? I can't reproduce this.
2. Yes, the progress bar works now. (Later, I thought we might change it to [match this example](https://github.com/napari/napari/blob/main/examples/mgui_with_threadworker_.py), but that is NOT working right now, despite me [using the development version of magicgui](https://github.com/napari/napari/issues/6354#issuecomment-1772781913), which should have fixed the bug. )
3. Yes, writing the output embeddings to file looks like it's working well. I will fix the test case by sorting lists before asserting they are the same, so it understands `os.listdir` output of `['.zattrs', '.zgroup', 'features']` is the same as `['.zgroup', '.zattrs', 'features']`
4. What is NOT working well is the global IMAGE_EMBEDDINGS variable. According to python, it doesn't exist when I try to access it from outside the thread worker that sets it (eg: in the main embedding_widget function, or in the test case).

How to fix the problem described by 4, above?
* Option 1: We could try to understand how global variables (this would be ideal, but I don't understand this well and am not confident)
* Option 2: Just remove the thread worker, and have the embedding computation block interaction with napari while it runs. (A good short term solution, especially if we can't figure out option 1 quickly)
* Option 3: We could refactor the code to not use global variables. (This would be a very large undertaking, and I'd prefer we picked a quicker solution for this, even if refactoring might be in the long term plan)

GenevieveBuckley:
I think we have one more problem, actually.

When you click the "Compute Embeddings" button a second time, nothing happens. That's ok, it checks the zarr and finds stuff already there, so it doesn't re-compute anything. That's good, even if it is a little disconcerting that there is no user feedback that the operation was sucessful (maybe we should have a "Finished" dialog box pop up or something).

BUT, when you change the input image layer, but forget to update the output zarr filepath... it doesn't seem to check whether the embedding contents in the zarr directory match the input image or not. I thought PR https://github.com/computational-cell-analytics/micro-sam/pull/67 fixed this. I don't see an error or warning message anywhere, and it is returning without computing anything, just like before (except now, this is not the behaviour we want).

constantinpape:
Hi @GenevieveBuckley,

there's quite a lot to follow up on, so I will structure it a bit:

## Running the plugin

I am not sure where the error with vigra comes from, but it indeed seems to be some error in my environment(s) (tested in two now), that happens when it's imported multiple times. I can fix it for now by commenting out a few imports that we currently don't need. Let's ignore this for now, and we can look into this more if the issue persists once we want to merge this (or even later once we have a more advanced state on dev). You can see  what I had to change in the proof of concept singleton PR (see below).

## Global state

I played with this a bit now, and you're right this is tricky. I don't think that this is a big issue going forward, since we want to get rid of the global state anyways! I wrote a quick proof of concept for replacing it with a singleton class and that works!
See https://github.com/GenevieveBuckley/micro-sam/pull/3. For testing that it works I have used this script: https://github.com/computational-cell-analytics/micro-sam/blob/wip-singleton/check_embed_widget.py and when pressing `p` I can check that the `image_embeddings` are computed (and it indeed works!).
So I suggest as next steps:
- I will implement the global state cleanly tomorrow with a singleton pattern (see also #201). I talked to Paul and he's too busy for this right now, but this is actually not that much work, and makes most sense for me to do it anyways because I know what state should go in, so I can do it tomorrow.
- Once that is done you can update the PR here using the singleton instead.

## Cached embeddings 

Indeed this should be addressed by #67 and you should at least see a warning when the wrong file is used. But maybe the warning is suppressed because it happens in the thread worker? 
You can also pass the [wrong_file_callback](https://github.com/computational-cell-analytics/micro-sam/blob/master/micro_sam/util.py#L514) that is called when this happens. You could use that to investigate this further, e.g. crash in that function to make sure the code is triggered.
(The idea behind this callback is to give the user the option to provide a new path, e.g. by opening a GUI where it can be entered, see https://github.com/computational-cell-analytics/micro-sam/blob/master/micro_sam/sam_annotator/gui_utils.py. I am not sure if this fits with the napari plugin logic, we can also change this a bit if needed.)

## Test error

That is weird. Unfortunately I can't run the test locally right now to check it because of
```
$ pytest -k test_annotator_2d
ERROR: usage: pytest [options] [file_or_dir] [file_or_dir] [...]
pytest: error: unrecognized arguments: --cov=micro_sam --cov-report
  inifile: /home/pape/Work/my_projects/micro-sam/pyproject.toml
  rootdir: /home/pape/Work/my_projects/micro-sam
```
Could you documented what's needed to run the tests now? Thanks!

constantinpape:
I implemented the singleton state in #240. This already works and if you want you can merge it here or rebase onto that PR @GenevieveBuckley. I only need to update it tomorrow for the 3d and tracking annotator, but that should not interfere with the changes here.

GenevieveBuckley:
> Could you documented what's needed to run the tests now? Thanks!
`pytest: error: unrecognized arguments: --cov=micro_sam --cov-report`

This looks like you don't have `pytest-cov` installed in your environment? Can you check that?

We install [a couple of pytest plugins (pytest-cov, pytest-qt)](https://github.com/computational-cell-analytics/micro-sam/blob/4fd9ce21dc0710985fddebccc81d80e549691a96/.github/workflows/test.yaml#L45) into our CI test environment.

You're right, documenting this in a contributing guide is a good idea.

constantinpape:
Hi @GenevieveBuckley,
a few updates from my side:
- I have finished the state refactoring, see #240 for details. If you agree with that solution you can merge that PR and use the `AnnotatorState` here. See my prototype implementation in https://github.com/GenevieveBuckley/micro-sam/pull/3 for how it can be used and tested here.
- I installed `pytest-cov` and can run the tests now locally, thanks!
- I created #242 for the developer documentation.
- The test fails because of https://github.com/computational-cell-analytics/micro-sam/blob/master/test/test_training.py#L80 : instead of using `device = "cpu"` we use `device = torch.device("cpu")` there, so the comparisons all fail. I can reproduce this locally, not sure why it works for you locally. We should make sure to also support `torch.device` in these functions and add unit-tests specifically for this as well.
    - We can either do this in the current PR.
    - Or do it in a follow-up PR, and for now just change to `device = "cpu"` in `test_training.py` as well. (I could take care of this, so no problem if you'd rather not do it now to focus on the other parts of this PR. Then please just create an issue about this and assign it to me. Description can be vague, I know what to do, just need an issue to keep track of it...)

GenevieveBuckley:
I think this is ready now - I'm very happy with how it has turned out!

* I've used your singleton AnnotatorState in the embedding widget. It's beautiful, and seems to work really well (🚀 https://github.com/computational-cell-analytics/micro-sam/pull/240)
* I've added an extra section to the test, so now it also checks the AnnotatorState values in-memory, as well as the results that are saved to disk.
* I've made some small changes so that `test_training.py` passes, so we don't have to worry about that test failing. Later, we can make a proper fix for https://github.com/computational-cell-analytics/micro-sam/issues/246, but for now we don't have to worry about it.
* The tests pass locally on my machine, our remote CI test job has passed, and I can open napari and manually interact with the embedding widget - everything looks like it makes sense.

GenevieveBuckley:
Extra notes:

1. I haven't tried to use the 'custom weights' option (mostly because I don't have any custom weights available). So that is untested. If somebody else could take a look at that, it would be helpful. This could be a follow up issue, I don't think it has to block this PR.

2. I've put in a bit of code to make sure the `save_path` directory exists and is empty. It prints an error message to the terminal if the directory is not empty.
A follow up piece of work for this might be trying to figure out why we don't see the hash comparison working (hint: we think it's because that happens in the thread worker), and/or trying to make a pop up dialog box to make the error more obvious than just printing to a terminal the user might not see.

3. For future work, we might like to investigate pop up dialog boxes more generally. It could be good for both error messages, and for informing the user when the embedding computation has completed (good for user feedback in both very short, and very long running computations). Possibly low priority, though.

micro_sam/sam_annotator/_widgets.py
```diff
@@ -0,0 +1,76 @@
...
+    model: Model = Model.__getitem__(_DEFAULT_MODEL),
+    device = "auto",
+    save_path: Optional[Path] = None,  # where embeddings for this image are cached (optional)
+    optional_custom_weights: Optional[Path] = None,  # A filepath or URL to custom model weights.
+) -> ImageEmbeddings:
+    """Image embedding widget."""
+    # Make sure save directory exists and is an empty directory
+    if save_path is not None:
+        if not save_path.exists():
+            os.makedirs(save_path)
```
constantinpape, line 44: Is there a reason for not just using `os.makedirs(save_path, exist_ok=True)`? That's a bit shorter and more readable.

micro_sam/util.py
```diff
@@ -258,6 +285,7 @@ def export_custom_sam_model(
     checkpoint_path: Union[str, os.PathLike],
     model_type: str,
     save_path: Union[str, os.PathLike],
+    device: str = None,
```
constantinpape, line 288: We don't want to have the device as an argument here: this function serialize a SAM checkpoint from our training to the normal weight format, so that it can be used by the napari tool (or by any other software that uses the SAM library).
This export should always go to the CPU, to ensure the model can be easily loaded.
So please remove the device argument here.

micro_sam/util.py
```diff
@@ -269,7 +297,7 @@ def export_custom_sam_model(
         save_path: Where to save the exported model.
     """
     _, state = get_custom_sam_model(
-        checkpoint_path, model_type=model_type, return_state=True, device=torch.device("cpu"),
+        checkpoint_path, model_type=model_type, return_state=True, device=device,
```
constantinpape, line 300: You can just hard-code `device="cpu"` here.

test/test_training.py
```diff
@@ -110,13 +110,14 @@ def _train_model(self, model_type, device):
         )
         trainer.fit(epochs=1)
 
-    def _export_model(self, checkpoint_path, export_path, model_type):
+    def _export_model(self, checkpoint_path, export_path, model_type, device):
```
constantinpape, line 113: No need to set the device here, see above.

constantinpape (approved):
Thanks @GenevieveBuckley! I ran this again and it works nicely. I also tested with custom weights and it seems to work. To fully check all of this we will need to integrate with the other widgets, but we can go ahead with this now.

micro_sam/sam_annotator/_widgets.py
```diff
@@ -0,0 +1,76 @@
...
+        if not save_path.is_dir():
+            raise NotADirectoryError(
+                f"The user selected 'save_path' is not a direcotry: {save_path}"
+            )
+        if len(os.listdir(save_path)) > 0:
```
constantinpape, line 49: I think this defeats the purpose of caching the affinities: the idea is that a user can close the annotation tool and open it again later while pointing to the same `save_path` so that the embeddings don't have to be recomputed.
This check prevents this and makes caching the embeddings not useful.

constantinpape (changes requested):
While going through the code again I noticed an issue with the core logic. I think this is easy to fix by removing the check but won't do it myself since I may overlook some intention behind adding this check.

micro_sam/util.py
constantinpape: I have fixed this already.

micro_sam/sam_annotator/_widgets.py
constantinpape: This also relates to your comment from above:

> I've put in a bit of code to make sure the save_path directory exists and is empty. It prints an error message to the terminal if the directory is not empty.

Printing the error here is not a good idea. Maybe you intended this because of the problem with the hash comparison, but I don't want to print an error for the correct behavior of using already cached affinities.
We could use a warning instead for now and remove this once we figure out the issue with the hash.

micro_sam/sam_annotator/_widgets.py
GenevieveBuckley: Done!

micro_sam/util.py
GenevieveBuckley: Ah, ok - thank you!

micro_sam/sam_annotator/_state.py
```diff
@@ -67,3 +68,14 @@ def initialized_for_tracking(self):
                 f"Invalid AnnotatorState: {init_sum} / 2 parts of the state "
                 "needed for tracking are initialized."
             )
+
+    def reset_state(self):
```
GenevieveBuckley, line 72: It seems like it will be easy to accidentally end up with a mixed state situation, where some of the attributes belong to one image, but others are from a different computation.
Eg: We have previously computed some state, including amg and/or tracks state information. Then, we want to compute the 2d image embeddings for a different input image. This will overwrite the `.image_embeddings` and `.predictor` attributes of the state, but because there is no amg or tracking stuff happening now those attributes are left unaltered.
The simplest solution to this problem seems to me that we should totally reset the state every time before we create or load new image embeddings. (Alternatively, we could try to add lines setting each of the unaltered attributes back to None... but that seems easier to make a mistake and accidentally forget some)

micro_sam/sam_annotator/_widgets.py
```diff
@@ -0,0 +1,82 @@
...
+        state.predictor  = get_sam_model(device=device, model_type=model.name,
+                                         checkpoint_path=optional_custom_weights)
+        # Compute the image embeddings
+        state.image_embeddings = precompute_image_embeddings(
+            predictor = state.predictor,
+            input_ = image_data,
+            save_path = str(save_path),
+            ndim=ndim,
+        )
+        state.image_shape = image_data.shape
```
GenevieveBuckley, line 79: Is the image_shape attribute supposed to include the RGB dimension in the case of a colour image?
Or does this need to be:
```python
if image.rgb:
    state.image_shape = image_data.shape[:-1]
else:
    state.image_shape = image_data.shape
```

micro_sam/util.py
```diff
@@ -552,7 +579,7 @@ def precompute_image_embeddings(
                     continue
                 # check whether the key signature does not match or is not in the file
                 if key not in f.attrs or f.attrs[key] != val:
-                    warnings.warn(
+                    raise RuntimeError(
```
GenevieveBuckley, line 582: Upgrading this from a user warning to an actual error allows users to see a little pop up happen in the bottom of the napari viewer if an error happens. I think this is sufficient for what we need.

micro_sam/sam_annotator/_widgets.py
GenevieveBuckley: That's reasonable - I've changed the code to fix this problem.
Now it works in all situations: (1) creating new image embeddings, (2) returning/loading existing image embeddings if the data_signature of the save directory matches our input image, and (3) making an error popup in the napari viewer if something goes wrong (data signature does not match, save directory is not a zarr array or empty folder, anything else unexpected that raises an actual error).

micro_sam/sam_annotator/_state.py
constantinpape: I think that's a very good idea.

micro_sam/sam_annotator/_widgets.py
constantinpape: The shape is supposed to only include the spatial dimensions. So we indeed need to add the condition for rgb images. Thanks for catching this!

micro_sam/util.py
constantinpape: That's good! I think it's good to update this to a real error in any case, since having different embeddings will lead to non-sensical results. I think we had a warning there in the beginning because we were still testing this functionality out, and then never got around to updating it to an error.

constantinpape (commented):
Thanks for the changes @GenevieveBuckley! This looks good now. The only thing left to do is updating how the image_shape is set.

constantinpape (approved):

GenevieveBuckley:
> Thanks for the changes @GenevieveBuckley! This looks good now. The only thing left to do is updating how the image_shape is set.

Updated!

constantinpape (approved):

PR #240, opened 2023-10-23
Implement singleton state and use it in the 2d annotator

### LINKED ISSUE #201: Refactor global state

constantinpape:
Fixes #201

### DIFF
--- a/micro_sam/sam_annotator/_state.py
+++ b/micro_sam/sam_annotator/_state.py
@@ -1,5 +1,5 @@
 from dataclasses import dataclass
-from typing import Optional
+from typing import Dict, Optional, Tuple
 
 from micro_sam.instance_segmentation import AMGBase
 from micro_sam.util import ImageEmbeddings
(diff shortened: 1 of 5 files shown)


constantinpape:
Hi @GenevieveBuckley ,
the state refactoring is good to go now. If you want you can quickly review if this also makes sense for you. If yes, feel free to merge this.
I think with these changes you can then go ahead with #235.

GenevieveBuckley (commented):
I keep coming back to this because I'm not sure I can make a meaningful review. Sorry!

This implementation looks like the metaclass singleton ([as described here](https://itnext.io/deciding-the-best-singleton-approach-in-python-65c61e90cdc4)), so it seems like a reasonable approach. But I don't have enough expertise to really know if this is definitely the best implementation approach, or assess any of the discussion around singletons in python.

I'd like to say, let's just try it and see what breaks. The test coverage isn't great, so I'm not hugely confident that everything will go smoothly. But on the other hand, we need to move forward. So I think we should go ahead and accept that maybe some unexpected problems might pop up. 

(The alternative is for me to pull this branch into my working branch, and check if singleton + embedding widget work well together. The disadvantage there is that multiple small PRs are usually quicker to get merged than one or two giant PRs).

constantinpape:
Thanks for the review @GenevieveBuckley!

> This implementation looks like the metaclass singleton ([as described here](https://itnext.io/deciding-the-best-singleton-approach-in-python-65c61e90cdc4)), so it seems like a reasonable approach. But I don't have enough expertise to really know if this is definitely the best implementation approach, or assess any of the discussion around singletons in python.

Yes, it follows the metaclass pattern. I don't use inheritance to have a separate implementation class that is not yet a singleton as in the example. I don't think it's needed for us, but it could be added later if required.

> I'd like to say, let's just try it and see what breaks. The test coverage isn't great, so I'm not hugely confident that everything will go smoothly. But on the other hand, we need to move forward. So I think we should go ahead and accept that maybe some unexpected problems might pop up.

Ok, let's go ahead with it! Re test coverage: I have added two unit-tests for the state now. And for high-level / integration tests we will need to wait till the plugin structure takes more shape. I don't think it makes too much sense to add many tests for the current tool design, as this will change a bit to accommodate the plugins.

> (The alternative is for me to pull this branch into my working branch, and check if singleton + embedding widget work well together. The disadvantage there is that multiple small PRs are usually quicker to get merged than one or two giant PRs).

Yes, I also prepare having smaller PRs.

PR #241, opened 2023-10-24
Add EM Generalist Finetuning

anwai98:
- The training scripts for lm generalist datasets (still need to test it and add necessary docstrings)
- TODO: need to make sure the data exists before checking RoIs (for getting data splits), else the script will crash


anwai98:
This is good to go. I tested the training for a few epochs, looks fine. Thanks!

finetuning/generalists/training/electron_microscopy/train_em_generalist.py
```diff
@@ -0,0 +1,97 @@
...
+import micro_sam.training as sam_training
+from micro_sam.util import export_custom_sam_model
+
+from obtain_em_datasets import get_generalist_em_loaders
+
+
+def finetune_em_generalist(args):
+    """Example code for finetuning SAM on multiple electron microscopy datasets"""
+    # override this (below) if you have some more complex set-up and need to specify the exact gpu
+    device = "cuda" if torch.cuda.is_available() else "gpu"
```
constantinpape, line 16: This should be `cpu` instead of `gpu`!

finetuning/generalists/training/light_microscopy/obtain_lm_datasets.py
```diff
@@ -0,0 +1,116 @@
+import os
```
constantinpape, line 1: I am confused why these files are part of this PR, they should have been merged via #228 already!

finetuning/generalists/training/light_microscopy/train_lm_generalist.py
```diff
@@ -0,0 +1,97 @@
+import os
```
constantinpape, line 1: I am confused why these files are part of this PR, they should have been merged via #228 already!

constantinpape (commented):
Found one typo and am a bit confused why the LM files are still in this PR.

finetuning/generalists/training/light_microscopy/obtain_lm_datasets.py
anwai98: Ah I might know the reason. Probably because I created a branch from `lm-generalist` branch, hence it also took those new files into consideration (which at that moment weren't a part of master)

finetuning/generalists/training/light_microscopy/train_lm_generalist.py
anwai98: Same as above

PR #250, opened 2023-10-26
Documenation build script - add check for missing docs files

GenevieveBuckley:
I've noticed that it's very easy to add a new documentation page and then forget to include it in the `micro_sam/__init__.py` module docstring. This means `pdoc` can't find it, and it will never appear on the website.

This PR:
1. Adds a check to the build_docs.py script, that looks for any missing markdown or RST files that exist in the "doc" folder but are not included in the `micro_sam/__init__.py` docstring. It raises a RuntimeError if any missing files are found, with instructions on how to fix it.
2. Edits the `.github/workflows/build_docs.yaml` script so that it uses the same `build_docs.py` script to generate the documentation. Advantages: (i) we can use the same check for missing files automatically, and (ii) it's simpler to use the same process everywhere
3. Adds the currently missing doc files `band.md` and `development.md` to the docstring in `micro_sam/__init__.py`. Now, `pdoc` will be able to find this content and include it on the website
4. Changed the heading levels the markdown file `band.md`. By making the subsections have level 2 markdown headings, instead of just bold text, the different sections will appear in the left hand sidebar menu of the website, and people can click on them to go directly to that section.

### DIFF
--- a/.github/workflows/build_docs.yaml
+++ b/.github/workflows/build_docs.yaml
@@ -43,11 +43,11 @@ jobs:
       # We use a custom build script for pdoc itself, ideally you just run `pdoc -o docs/ ...` here.
       - name: Run pdoc
         shell: bash -l {0}
-        run: pdoc --docformat google --logo "https://raw.githubusercontent.com/computational-cell-analytics/micro-sam/master/doc/images/micro-sam-logo.png" -o docs/ micro_sam
+        run: python build_doc.py --out
 
       - uses: actions/upload-pages-artifact@v1
         with:
-          path: docs/
+          path: tmp/
 
   # Deploy the artifact to GitHub pages.
   # This is a separate job so that only actions/deploy-pages has the necessary permissions.
--- a/build_doc.py
+++ b/build_doc.py
@@ -1,17 +1,44 @@
 import argparse
+import glob
+import os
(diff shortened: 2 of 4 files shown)


GenevieveBuckley:
Also - I've opened this PR against the master branch, because I think we need it everywhere. Is that the right way to go? 

I notice that the `band.md` documentation file does not exist on the `dev` branch, only in `master`, and that made me a little confused about where my pull request needed to go. It's possible `dev` and `master` have drifted a little bit out of sync, might need to check that.

constantinpape:
> Also - I've opened this PR against the master branch, because I think we need it everywhere. Is that the right way to go?

Yes, that's ok.

> I notice that the `band.md` documentation file does not exist on the `dev` branch, only in `master`, and that made me a little confused about where my pull request needed to go. It's possible `dev` and `master` have drifted a little bit out of sync, might need to check that.

Yes, they got a little out of sync. I pushed BAND to the master branch to make it easier to access by a collaborator who is using it. (But it should be included in the documentation, since it is a generally useful feature).

In general: we should merge changes into `dev` if they change existing functionality, but it's ok to merge on `master` if this is not the case (as here).

constantinpape (approved):

PR #254, opened 2023-10-30
Refactor Iterative Prompting Evaluation

anwai98:
WIP - adapting iterative prompting evaluation according to the updated iterative prompt generator.

### DIFF
--- a/micro_sam/evaluation/inference.py
+++ b/micro_sam/evaluation/inference.py
@@ -308,7 +308,7 @@ def run_inference_with_prompts(
     prompt_save_dir: Optional[Union[str, os.PathLike]] = None,
     batch_size: int = 512,
 ) -> None:
-    """Run segment anything inference for multiple images using prompts derived form groundtruth.
+    """Run segment anything inference for multiple images using prompts derived from groundtruth.
 
     Args:
         predictor: The SegmentAnything predictor.
@@ -395,22 +395,14 @@ def _save_segmentation(masks, prediction_path):
     imageio.imwrite(prediction_path, segmentation)
 
 
+@torch.no_grad()
 def _run_inference_with_iterative_prompting_for_image(
-    model,
-    image,
-    gt,
-    n_iterations,
-    device,
-    use_boxes,
-    prediction_paths,
-    batch_size,
+    model, image, gt, n_iterations, device, use_boxes, prediction_paths, batch_size
 ):
     assert len(prediction_paths) == n_iterations, f"{len(prediction_paths)}, {n_iterations}"
     to_sam_inputs = ConvertToSamInputs()
 
-    image = torch.from_numpy(
-        image[None, None] if image.ndim == 2 else image[None]
-    )
+    image = torch.from_numpy(image[None, None] if image.ndim == 2 else image[None])
     gt = torch.from_numpy(gt[None].astype("int32"))
 
     n_pos = 0 if use_boxes else 1
@@ -419,7 +411,7 @@ def _run_inference_with_iterative_prompting_for_image(
     input_images = torch.stack([model.preprocess(x=x["image"].to(device)) for x in batched_inputs], dim=0)
     image_embeddings = model.image_embeddings_oft(input_images)
 
-    multimasking = n_pos == 1
+    multimasking = (n_pos == 1)
     prompt_generator = IterativePromptGenerator()
 
     n_samples = len(sampled_ids[0])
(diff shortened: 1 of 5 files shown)


micro_sam/evaluation/inference.py
```diff
@@ -395,44 +394,47 @@ def _save_segmentation(masks, prediction_path):
...
+    multimasking = (n_pos == 1)
     prompt_generator = IterativePromptGenerator()
 
     n_samples = len(sampled_ids[0])
     n_batches = int(np.ceil(float(n_samples) / batch_size))
 
     for iteration in range(n_iterations):
-        final_masks = []
+        final_masks, all_sampled_binary_y = [], []
         for batch_idx in range(n_batches):
```
constantinpape, line 432: You should use https://github.com/computational-cell-analytics/micro-sam/blob/dev/micro_sam/inference.py#L17 here.

anwai98:
This is working as expected (tested it for a few samples per setting). Let me know how it looks. Thanks!

micro_sam/evaluation/inference.py
```diff
@@ -392,118 +389,127 @@ def _save_segmentation(masks, prediction_path):
...
     masks = [{"segmentation": mask, "area": mask.sum()} for mask in masks]
     segmentation = mask_data_to_segmentation(masks, shape, with_background=True)
-    imageio.imwrite(prediction_path, segmentation)
+    imageio.imwrite(prediction_path, segmentation, compression=5)
+
+
+def extract_instances_from_batched_outputs(batched_outputs):
+    masks = [m["segmentation"][None] for m in batched_outputs]
+    masks = torch.stack(masks)
+    masks = (masks > 0.5).to(torch.float32)
```
constantinpape, line 398: The masks are already binary, so you don't need this. Overall, this is simple enough to write it as a one-liner and then you don't need to refactor this into a function.

constantinpape: See comment below.

micro_sam/evaluation/inference.py
```diff
@@ -392,118 +389,127 @@ def _save_segmentation(masks, prediction_path):
...
-                mask, l_mask = torch.stack(mask), torch.stack(l_mask)
-                masks.append(mask)
-                logits_masks.append(l_mask)
+        batched_outputs = batched_inference(
+            predictor, image, batch_size,
+            boxes=boxes, points=points, point_labels=point_labels,
+            multimasking=multimasking, embedding_path=embedding_path,
+            return_instance_segmentation=False
+        )
+        masks = extract_instances_from_batched_outputs(batched_outputs)
```
constantinpape, line 450: You can do this with a one-liner here:
```python
masks = torch.stack([m["segmentation"][None] for m in batched_outputs]).to(torch.float32)
```

micro_sam/evaluation/inference.py
```diff
@@ -392,118 +389,127 @@ def _save_segmentation(masks, prediction_path):
...
+
+
+def extract_instances_from_batched_outputs(batched_outputs):
+    masks = [m["segmentation"][None] for m in batched_outputs]
+    masks = torch.stack(masks)
+    masks = (masks > 0.5).to(torch.float32)
+    return masks
+
+
+def get_channelwise_gt_objects(gt, gt_ids, device=None):
```
constantinpape, line 402: We have a function for this already and don't need this.

constantinpape: See comment below.

micro_sam/evaluation/inference.py
```diff
@@ -392,118 +389,127 @@ def _save_segmentation(masks, prediction_path):
...
+        gt, gt_ids, use_points, use_boxes,
+        n_positives=n_positives,
+        n_negatives=n_negatives,
+        dilation=dilation
     )
-    gt = torch.from_numpy(gt[None].astype("int32"))
 
-    n_pos = 0 if use_boxes else 1
-    batched_inputs, sampled_ids = to_sam_inputs(image, gt, n_pos=n_pos, n_neg=0, get_boxes=use_boxes)
+    sampled_binary_gt = get_channelwise_gt_objects(gt, gt_ids)
```
constantinpape, line 436: You can use https://github.com/computational-cell-analytics/micro-sam/blob/master/micro_sam/util.py#L696 instead.

micro_sam/evaluation/inference.py
```diff
@@ -392,118 +389,127 @@ def _save_segmentation(masks, prediction_path):
...
 def run_inference_with_iterative_prompting(
-    checkpoint_path: Union[str, os.PathLike],
-    model_type: str,
+    predictor: SamPredictor,
     image_paths: List[Union[str, os.PathLike]],
     gt_paths: List[Union[str, os.PathLike]],
-    prediction_root: Union[str, os.PathLike],
+    embedding_dir: Union[str, os.PathLike],
+    prediction_dir: Union[str, os.PathLike],
+    use_points: bool,
```
constantinpape, line 474: Having both `use_points` and `use_boxes` here are too many options. We should only have two choices:
- We start with a single positive point.
- We start with a bounding box.

So it's enough to have a single boolean for this, e.g. `start_with_box_prompt`.

micro_sam/evaluation/inference.py
```diff
@@ -392,118 +389,127 @@ def _save_segmentation(masks, prediction_path):
...
+        embedding_dir: The directory where the image embeddings will be saved or are already saved.
+        prediction_dir: The directory where the predictions from SegmentAnything will be saved per iteration.
+        use_points: Whether to use the first prompt as 1 positive point
+        use_boxes: Whether to use the first prompt as bounding box
+        dilation: The dilation factor for the radius around the ground-truth obkect
+            around which points will not be sampled.
+        batch_size: The batch size used for batched predictions.
+        n_iterations: The number of iterations for iterative prompting.
+    """
+    if not (use_points or use_boxes):
```
constantinpape, line 496: We don't need this check if we simplify the choices here, see comment above.

micro_sam/evaluation/inference.py
```diff
@@ -514,7 +520,10 @@ def run_inference_with_iterative_prompting(
...
         gt = relabel_sequential(gt)[0]
 
-        with torch.no_grad():
-            _run_inference_with_iterative_prompting_for_image(
-                model, image, gt, n_iterations, device, use_boxes, prediction_paths, batch_size,
-            )
+        embedding_path = os.path.join(embedding_dir, f"{os.path.splitext(image_name)[0]}.zarr")
+
+        _run_inference_with_iterative_prompting_for_image(
+            predictor, image, gt, use_points=use_points, use_boxes=use_boxes,
```
constantinpape, line 526: Change this to the simpler choice with just one boolean, see above.

micro_sam/evaluation/inference.py
constantinpape: Note: this requires the gt to be of type `"int64"`, you can just pass it like this: `gt.astype("int64")` to make it work.

micro_sam/evaluation/inference.py
```diff
@@ -392,118 +389,127 @@ def _save_segmentation(masks, prediction_path):
...
+    embedding_path,
+    n_iterations,
+    prediction_paths
+) -> None:
+    prompt_generator = IterativePromptGenerator()
 
-    image = torch.from_numpy(
-        image[None, None] if image.ndim == 2 else image[None]
+    gt_ids = np.unique(gt)[1:]
+    n_positives = 1 if not use_boxes else 0
```
constantinpape, line 426: You can simplify this a bit once we have the single boolean `start_with_box_prompt`:
```python
n_negatives = 0
if start_with_box_prompt:
  use_boxes, use_points = True, False
  n_positives = 0
  multimasking = False
else:
  use_boxes, use_points = False, True
  n_positives = 0
  multimasking = True
```

constantinpape: See also below for comment w.r.t. multi-masking.

micro_sam/evaluation/inference.py
```diff
@@ -392,118 +389,127 @@ def _save_segmentation(masks, prediction_path):
...
-
-    multimasking = n_pos == 1
-    prompt_generator = IterativePromptGenerator()
-
-    n_samples = len(sampled_ids[0])
-    n_batches = int(np.ceil(float(n_samples) / batch_size))
+    # Use multi-masking only if we have a single positive point without box
+    multimasking = False
+    if not use_boxes and (n_positives == 1 and n_negatives == 0):
+        multimasking = True
```
constantinpape, line 441: You can make the mulitmasking decision above already, it's easier to read this if it's in a single if-else. And you need to always set it to False after the first loop, because we will then have more than one prompt and should not use multimasking anymore.

micro_sam/evaluation/inference.py
```diff
@@ -392,118 +389,127 @@ def _save_segmentation(masks, prediction_path):
...
-                _inp["point_labels"] = updated_point_labels
-                _inp["mask_inputs"] = logits
+        if point_labels is not None:
+            point_labels = np.concatenate([point_labels, next_labels], axis=1)
+        else:
+            point_labels = next_labels
 
-        final_masks = torch.cat(final_masks, dim=1)
-        _save_segmentation(final_masks, prediction_paths[iteration])
+        _save_segmentation(masks, prediction_paths[iteration])
```
constantinpape, line 465: Set `multimasking = False` at the end of the loop. Otherwise it will stay true if it was set to true in the beginning. But that is wrong because we have more than one prompt after the first iteration.

constantinpape (commented):
This looks good overall now! I left a few detailed comments, but these are mostly minor changes.

micro_sam/evaluation/inference.py
anwai98: Ahha yes, that's correct. Fixed it, thanks!

micro_sam/evaluation/inference.py
anwai98: Ahha good to know. I'll use this from here-on. Thanks!

micro_sam/evaluation/inference.py
anwai98: Okay this is crucial. Thanks for pointing it out.

constantinpape (approved):

PR #257, opened 2023-11-01
Fix Bounding Box Sorting

anwai98:
Observed that when we expect bounding boxes for all objects in an image, there was a minor fix expected (see [here](https://github.com/computational-cell-analytics/micro-sam/pull/256)). Will test with it if the finetuning still works (in theory it should but just in case)


constantinpape:
> Will test with it if the finetuning still works (in theory it should but just in case)

You don't need to test. This fix is correct. The previous code would not have worked, we just did not run into it yet.

anwai98:
Okay, thanks Constantin!

PR #258, opened 2023-11-02
Add MitoEM Specialist Finetuning

anwai98:
This PR adds MitoEM specialist finetuning on 50% of the dataset (WIP: still to be trained and evaluated)


anwai98:
This PR is now taken care of in a different generalist setup. I will go ahead and close this. Thanks!

PR #261, opened 2023-11-02
Fix Predictor Initialization

anwai98:
This (probably) fixes the iterative prompting. It still needs to be tested, I will get back on this soon.


anwai98:
Okay, the results for the first iteration look consistent (with the weights present in our project directory at scratch). This is good to go.

constantinpape (approved):

PR #265, opened 2023-11-06
Experiments with Mask Inputs Probability

anwai98:
(WIP) Experiments without the input mask prompts while iterative prompting in the training scheme

### DIFF
--- a/micro_sam/training/sam_trainer.py
+++ b/micro_sam/training/sam_trainer.py
@@ -42,7 +42,7 @@ class SamTrainer(torch_em.trainer.DefaultTrainer):
         mse_loss: torch.nn.Module = torch.nn.MSELoss(),
         _sigmoid: torch.nn.Module = torch.nn.Sigmoid(),
         prompt_generator: PromptGeneratorBase = IterativePromptGenerator(),
-        mask_prob: float = 0.0,
+        mask_prob: float = 0.5,
         **kwargs
     ):
         super().__init__(**kwargs)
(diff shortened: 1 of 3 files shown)


anwai98:
The complete set of results from mask input probabilities are [here](https://docs.google.com/spreadsheets/d/10GgjA9CDe0y4Ww2hfuma2mBV6L3KmTQLqCKPrTSe9D0/edit?usp=sharing): we see that the batched inference results are a little better when the probability of using mask inputs is less than 1. Let's stick to 0.5 as the default value as we discussed.

anwai98:
I just updated the default value (to `0.5` in the trainer). This now looks good to go. Thanks!

micro_sam/training/sam_trainer.py
```diff
@@ -250,7 +255,14 @@ def _get_updated_points_per_mask_per_subiter(self, masks, sampled_binary_y, batc
 
             _inp["point_coords"] = updated_point_coords
             _inp["point_labels"] = updated_point_labels
-            _inp["mask_inputs"] = logits
+
+            if self.mask_prob > 0:
+                # using mask inputs for iterative prompting while training, with a probability
+                use_mask_inputs = True if random.random() < self.mask_prob else False
```
constantinpape, line 261: I am not 100% sure about the order here.

anwai98: This is now fixed with the latest commit, and is good to go. Thanks!

constantinpape (approved):

PR #267, opened 2023-11-08
Run github actiosns CI tests on Windows, Mac and Linux

GenevieveBuckley:
I think it's sensible to run the github actions CI tests on Windows, Mac and Linux, instead of just Linux. 

That way we can catch any silly errors like accidentally hard coding forward slashes in cache directory paths, etc. that might work on Linux but fail on Windows (one example of a mistake I'm hoping not to make, but it would be nice to have the CI run a backup check for any silly problems like this)

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -19,7 +19,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest]
+        os: [ubuntu-latest, windows-latest, macos-latest]
         python-version: ["3.10"]
 
     steps:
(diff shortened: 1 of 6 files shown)


constantinpape:
I agree that this is a good idea!
I had a look at the current test failures and they seem to be due to:
- Windows: the GUI test doesn't work. I suggest we disable it for windows.
- Mac: the CI uses MPS but the memory is so small that some tests run out of memory. I suggest we set the device to CPU for those tests. (Maybe we want to introduce some check that we are in the CI for this.)

constantinpape:
I went ahead and finished this up @GenevieveBuckley

Issue #264, opened 2023-11-04
Use a consistent cache directory


constantinpape:
We currently use different cache directories for saving the models vs. saving the sample image data.
- The models use `$HOME/.sam_models`.
- The sample data uses `$HOME/.cache/micro-sam`

It would be good to be consistent here and I think `$HOME/.cache/micro_sam` is a good choice. For the models we could use  `$HOME/.cache/micro_sam/models`.

What do you think @GenevieveBuckley ?

GenevieveBuckley:
I say it's a good idea and I'm happy to do that.

constantinpape:
Ok! The path for the downloads is currently set here: https://github.com/computational-cell-analytics/micro-sam/blob/master/micro_sam/util.py#L50

The intention behind this code was to enable changing the cache directory via the `SAM_MODELS` environment variable. But this is not documented anywhere so we can change it.

My suggestion would be that we introduce a new environment variable `MICRO_SAM_CACHE` instead that defaults to `$HOME/.cache/micro_sam`. We should keep the option to set a different cache path for systems where the storage in `$HOME` is limited (e.g. computer clusters).

Issue #249, opened 2023-10-26
Housekeeping: doc update, release, codecov, templates


constantinpape:
- [ ] Update the [developer documentation](https://github.com/computational-cell-analytics/micro-sam/blob/master/doc/development.md) to reflect new plugin design
- [ ] Add [BAND](https://github.com/computational-cell-analytics/micro-sam/blob/master/doc/band.md) to the doc. (Formulating this more general for future services like KissKI or BioEngine)
- [ ] Figure out what's going on with codecov, see #234.
- [ ] Create an FAQ for the Documentation, to address common installation and performance issues.
- [ ] Create issue (referencing FAQ) and PR templates (referencing the new Contributing section).

GenevieveBuckley:
> Add [BAND](https://github.com/computational-cell-analytics/micro-sam/blob/master/doc/band.md) to the doc. (Formulating this more general for future services like KissKI or BioEngine)

PR https://github.com/computational-cell-analytics/micro-sam/pull/250 might help with the first part of this (but not KissKI or BioEngine)

claydugo:
This is pointing the readme to `docs/` instead of the actual directory `doc/` FYI

constantinpape:
Hi @claydugo ,
I am not quite sure what you mean by your commen. Where do we point to `docs` instead of `doc`?
Thanks!

claydugo:
In the README for the [contributing and developer documentation links](https://github.com/computational-cell-analytics/micro-sam/pull/270).

anwai98:
Adding PyTorch / CUDA Toolkit minimum version requirements (or tested ones) somewhere in the documentation (for finetuning and features requiring PyTorch)?

anwai98:
Generating and uploading our `micro-sam` package to PyPI (heard some interests on this)

PR #268, opened 2023-11-08
Use a consistent cache directory

GenevieveBuckley:
Closes https://github.com/computational-cell-analytics/micro-sam/issues/264

This PR should make the cache directory handling consistent between the model weights and sample data. 
By default, we use:
* default cache dir + 'micro_sam/models'
* default cache dir + 'micro_sam/sample_data`

[From the pooch.os_cache() docs](https://www.fatiando.org/pooch/latest/api/generated/pooch.os_cache.html):
> Default cache location based on the operating system.
>
> The folder locations are defined by the platformdirs package using the user_cache_dir function. Usually, the locations will be following (see the [platformdirs documentation](https://platformdirs.readthedocs.io/)):
>
> * Mac: ~/Library/Caches/<AppName>
> * Unix: ~/.cache/<AppName> or the value of the XDG_CACHE_HOME environment variable, if defined.
> * Windows: C:\Users\<user>\AppData\Local\<AppAuthor>\<AppName>\Cache

Additionally, users can set an environment variable named `MICROSAM_CACHEDIR` to choose a different cache directory somewhere else (not the default cache dir + 'micro_sam').

The documentation for the environment variable cache dir control is a bit hidden, currently it is only mentioned:
* in the module docstring for `micro_sam/sample_data.py`, and 
* in the docstring for `get_sam_model` function in `micro_sam/util.py`.

### DIFF
--- a/micro_sam/sample_data.py
+++ b/micro_sam/sample_data.py
@@ -1,5 +1,15 @@
 """
 Sample microscopy data.
+
+You can change the download location for sample data and model weights
+by setting the environment variable: MICROSAM_CACHEDIR
+
+By default sample data is downloaded to a folder named 'micro_sam/sample_data'
+inside your default cache directory, eg:
+    * Mac: ~/Library/Caches/<AppName>
+    * Unix: ~/.cache/<AppName> or the value of the XDG_CACHE_HOME environment variable, if defined.
+    * Windows: C:\Users\<user>\AppData\Local\<AppAuthor>\<AppName>\Cache
+
 """
 
 import os
@@ -56,7 +66,8 @@ def sample_data_image_series():
     # Check the documentation for more information about the
     # add_image_kwargs
     # https://napari.org/stable/api/napari.Viewer.html#napari.Viewer.add_image
-    default_base_data_dir = pooch.os_cache('micro-sam')
+    _CACHE_DIR = os.environ.get('MICROSAM_CACHEDIR') or pooch.os_cache('micro_sam')
+    default_base_data_dir = os.path.join(_CACHE_DIR, 'sample_data')
     data_directory = fetch_image_series_example_data(default_base_data_dir)
     fnames = os.listdir(data_directory)
     full_filenames = [os.path.join(data_directory, f) for f in fnames]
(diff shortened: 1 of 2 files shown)


GenevieveBuckley:
Questions:
1. Do I need to run `os.path.expanduser` on the `pooch.os_cache` return value / environment variable set by the user? Currently I am not doing this, not sure if there's likely to be a situation where that might cause a problem or not.
2. Is there anywhere else in the documentation where we need to write about the option to set the os environment variable `MICROSAM_CACHEDIR`?

Interesting thought: maybe we could even make a napari widget for setting the cache directory path to something other than the default value. Then we could rely less on people reading every single page of the API documentation, and have better discoverability.

constantinpape (approved):

constantinpape:
> 1. Do I need to run `os.path.expanduser` on the `pooch.os_cache` return value / environment variable set by the user? Currently I am not doing this, not sure if there's likely to be a situation where that might cause a problem or not.

No I don't think so. For me `os_cache` returns a user expanded path already:
```
In [4]: import pooch
In [5]: pooch.os_cache("micro_sam")
Out[5]: PosixPath('/home/pape/.cache/micro_sam')
```

> 2\. Is there anywhere else in the documentation where we need to write about the option to set the os environment variable `MICROSAM_CACHEDIR`?

There isn't really a place to put it yet. I have made a note in #249 on this.

> Interesting thought: maybe we could even make a napari widget for setting the cache directory path to something other than the default value. Then we could rely less on people reading every single page of the API documentation, and have better discoverability.

I like that idea in principle, but we probably need to think a bit about the implementation and try how well it works in practice. I will make an issue out of this so that we can keep track of it. But I think we should only revisit this once we have made some more advances towards the plugin.

GenevieveBuckley:
> > 1. Do I need to run `os.path.expanduser` on the `pooch.os_cache` return value / environment variable set by the user? Currently I am not doing this, not sure if there's likely to be a situation where that might cause a problem or not.
> 
> No I don't think so. For me `os_cache` returns a user expanded path already:
> 
> ```
> In [4]: import pooch
> In [5]: pooch.os_cache("micro_sam")
> Out[5]: PosixPath('/home/pape/.cache/micro_sam')
> ```

That's the behaviour I see too, but the [pooch os_cache documentation](https://www.fatiando.org/pooch/latest/api/generated/pooch.os_cache.html) does say "User directories ('~') are not expanded" so perhaps this may not *always* happen (or might not be guaranteed in the future). 

Anyway, I don't think this is causing problems, but I could do it if we felt we needed to be super safe.

constantinpape:
> Anyway, I don't think this is causing problems, but I could do it if we felt we needed to be super safe.

I see. Then maybe it's a good idea to expand the path since it should not hurt.

PR #270, opened 2023-11-08
Fixup README 404s

claydugo:

PR #272, opened 2023-11-08
Fix readme links

GenevieveBuckley:
Pair to https://github.com/computational-cell-analytics/micro-sam/pull/270

That PR fixed the master branch, this PR fixes the dev branch so it doesn't overwrite master later (maybe not a realistic worry, git probably handles it, but just in case). Alternatively, we could merge the changes from the master branch into dev.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ Please check out [the documentation](https://computational-cell-analytics.github
 
 We welcome new contributions!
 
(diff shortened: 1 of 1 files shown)


constantinpape:
> Alternatively, we could merge the changes from the master branch into dev.

I would rather merge master into dev. We have recently found an important fix in training that we want to push to master directly and then merge back onto dev. I will follow up on this.

Issue #36, opened 2023-06-03
Use `pooch` for downloading the model:


GenevieveBuckley:
> We could also think about using `pooch` for downloading the model: https://github.com/computational-cell-analytics/micro-sam/blob/master/micro_sam/util.py#L39-L63

_Originally posted by @constantinpape in https://github.com/computational-cell-analytics/micro-sam/issues/35#issuecomment-1573316601_

hmaarrfk:
FYI, i've found that SHA256 is quite slow at computing hashes for large files. (i.e. 2GB models...)

you might want to play around with maybe cryptographically non-secure hashes. I've used XXH128 with great success.

constantinpape:
Hi @hmaarrfk,

I think that would make sense. I don't think we require the hashing to be secure since the goal here is not to avoid malicious attacks, but just to make sure that files do not change.

In case you want to look into this: contributions are very welcome. (If not: no worries, but this would be on my list of lower priorities, so I can't really promise when we'd have time to look into this).
For details on how to contribute see https://github.com/computational-cell-analytics/micro-sam/blob/master/doc/contributing.md.

GenevieveBuckley:
That should be straightforward to do, pooch supports the [python-xxhash](https://github.com/ifduyue/python-xxhash) bindings. 

The docs even have this nice example about how to use it, you just append `xxh128:` to the start of the hash.
https://www.fatiando.org/pooch/latest/hashes.html#other-supported-hashes

```python
import datetime
import pooch

# Get the current data to store the files in separate folders
CURRENT_DATE = datetime.datetime.now().date()

GOODBOY = pooch.create(
    [...],
    registry={
        "store.zip": "xxh128:6a71973c93eac6c8839ce751ce10ae48",
    },
)
```

constantinpape:
Implemented in #276

PR #276, opened 2023-11-14
Download models with pooch

GenevieveBuckley:
I've made a start on https://github.com/computational-cell-analytics/micro-sam/issues/36

* This PR downloads the models using pooch
* We are still using the SHA256 hashes for validation, although at some future point we could use a (non-cryptographic) XXH128 hash [as discussed here](https://github.com/computational-cell-analytics/micro-sam/issues/36#issuecomment-1809426270)
* I've removed the 'checkpoint' kwarg from `get_sam_model`. I think it makes more sense to have custom weights of any sort go through the `get_custom_sam_model` function (I haven't touched this function yet), and also if we download weights from different checkpoints then the hash validation is going to break.
* Problem: I can't test a lot of this, because...
    * we have low test coverage, and
    * the example scripts often use hard coded paths to locations on Constantin's computer, so I can't run them (eg: `examples/annotator_with_custom_model.py`)

### DIFF
--- a/micro_sam/util.py
+++ b/micro_sam/util.py
@@ -35,42 +35,7 @@ try:
 except ImportError:
     from tqdm import tqdm
 
-_MODEL_URLS = {
-    # the default segment anything models
-    "vit_h": "https://dl.fbaipublicfiles.com/segment_anything/sam_vit_h_4b8939.pth",
-    "vit_l": "https://dl.fbaipublicfiles.com/segment_anything/sam_vit_l_0b3195.pth",
-    "vit_b": "https://dl.fbaipublicfiles.com/segment_anything/sam_vit_b_01ec64.pth",
-    # the model with vit tiny backend fom https://github.com/ChaoningZhang/MobileSAM
-    "vit_t": "https://owncloud.gwdg.de/index.php/s/TuDzuwVDHd1ZDnQ/download",
-    # first version of finetuned models on zenodo
-    "vit_h_lm": "https://zenodo.org/record/8250299/files/vit_h_lm.pth?download=1",
-    "vit_b_lm": "https://zenodo.org/record/8250281/files/vit_b_lm.pth?download=1",
-    "vit_h_em": "https://zenodo.org/record/8250291/files/vit_h_em.pth?download=1",
-    "vit_b_em": "https://zenodo.org/record/8250260/files/vit_b_em.pth?download=1",
-}
-_CACHE_DIR = os.environ.get('MICROSAM_CACHEDIR') or pooch.os_cache('micro_sam')
-_CHECKPOINT_FOLDER = os.path.join(_CACHE_DIR, 'models')
-_CHECKSUMS = {
-    # the default segment anything models
-    "vit_h": "a7bf3b02f3ebf1267aba913ff637d9a2d5c33d3173bb679e46d9f338c26f262e",
-    "vit_l": "3adcc4315b642a4d2101128f611684e8734c41232a17c648ed1693702a49a622",
-    "vit_b": "ec2df62732614e57411cdcf32a23ffdf28910380d03139ee0f4fcbe91eb8c912",
-    # the model with vit tiny backend fom https://github.com/ChaoningZhang/MobileSAM
-    "vit_t": "6dbb90523a35330fedd7f1d3dfc66f995213d81b29a5ca8108dbcdd4e37d6c2f",
-    # first version of finetuned models on zenodo
-    "vit_h_lm": "9a65ee0cddc05a98d60469a12a058859c89dc3ea3ba39fed9b90d786253fbf26",
-    "vit_b_lm": "5a59cc4064092d54cd4d92cd967e39168f3760905431e868e474d60fe5464ecd",
-    "vit_h_em": "ae3798a0646c8df1d4db147998a2d37e402ff57d3aa4e571792fbb911d8a979c",
-    "vit_b_em": "c04a714a4e14a110f0eec055a65f7409d54e6bf733164d2933a0ce556f7d6f81",
-}
-# this is required so that the downloaded file is not called 'download'
-_DOWNLOAD_NAMES = {
-    "vit_t": "vit_t_mobile_sam.pth",
-    "vit_h_lm": "vit_h_lm.pth",
-    "vit_b_lm": "vit_b_lm.pth",
-    "vit_h_em": "vit_h_em.pth",
-    "vit_b_em": "vit_b_em.pth",
-}
+
 # this is the default model used in micro_sam
 # currently set to the default vit_h
 _DEFAULT_MODEL = "vit_h"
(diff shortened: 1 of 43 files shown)


GenevieveBuckley:
If/when we go ahead with https://github.com/computational-cell-analytics/micro-sam/pull/276, we'll need to make sure `pooch.os_cache` is recomputed in the pooch MODELS registry every time, instead of just once when the module is imported.

micro_sam/util.py
```diff
@@ -84,49 +49,38 @@
...
-
-        # download the checkpoint if necessary
-        if not os.path.exists(checkpoint_path):
-            os.makedirs(_CHECKPOINT_FOLDER, exist_ok=True)
-            _download(checkpoint_url, checkpoint_path, model_type)
-    elif not os.path.exists(checkpoint_path):
-        raise ValueError(f"The checkpoint path {checkpoint_path} that was passed does not exist.")
-
-    return checkpoint_path
+_CACHE_DIR = os.environ.get('MICROSAM_CACHEDIR') or pooch.os_cache('micro_sam')
```
constantinpape, line 52: Should we maybe wrap all of this into function(s)? 
This seems to be a good idea in general and is the easiest solution I can think of for updating the cache path re #277.

And we should probably have two functions: one that returns the cache path (and that can also be used by other functionality that uses it) and one that returns the pooch model registry.

constantinpape:
Overall this is great!

> * We are still using the SHA256 hashes for validation, although at some future point we could use a (non-cryptographic) XXH128 hash [as discussed here](https://github.com/computational-cell-analytics/micro-sam/issues/36#issuecomment-1809426270)

That's good to know, and we can address this in a follow up.

> * I've removed the 'checkpoint' kwarg from `get_sam_model`. I think it makes more sense to have custom weights of any sort go through the `get_custom_sam_model` function (I haven't touched this function yet), and also if we download weights from different checkpoints then the hash validation is going to break.

I think that makes sense. I need to think about this a bit more because `get_custom_sam_model` currently also supports other kinds of weights. I will read this more carefully later and follow up.

> Problem: I can't test a lot of this, because...
> 
>     * we have low test coverage, and
> 
>     * the example scripts often use hard coded paths to locations on Constantin's computer, so I can't run them (eg: `examples/annotator_with_custom_model.py`)

- Regarding test coverage: we should probably add some more unit-test for the weights, I will think about it.
- Regarding example scripts: good point. I will update those later and remove as much hard-coded paths as possible.

micro_sam/util.py
```diff
@@ -177,7 +131,6 @@ def _available_devices():
 def get_sam_model(
     model_type: str = _DEFAULT_MODEL,
     device: Optional[str] = None,
-    checkpoint_path: Optional[Union[str, os.PathLike]] = None,
```
constantinpape, line 180: I think we should keep this. Merging this into `get_custom_sam_model` would be confusing and I don't see a big issue keeping it here. See my main comment for more details.

micro_sam/util.py
```diff
@@ -203,7 +156,7 @@ def get_sam_model(
     Returns:
         The segment anything predictor.
     """
-    checkpoint = _get_checkpoint(model_type, checkpoint_path)
+    checkpoint = MODELS.fetch(model_type)
```
constantinpape, line 159: We can just add something like this to support `checkpoint_path`:
```python
if checkpoint_path is None:
  checkpoint = MODELS.fetch(model_type)
else:
  checkpoint = checkpoint_path
```

constantinpape: So I don't see any problem with mixing up hashes.

constantinpape:
Hi @GenevieveBuckley ,
I had some time to look into this now:

## get_sam_model

I think we should keep the (optional) `checkpoint_path` argument there. Mixing this up with `get_custom_sam_model` would be confusing. To clarify this:
- `get_sam_model` loads a model from weights stored in the default format. These can come from two sources: either from weights that are available online (either the original SAM weights or our fine-tuned models) and that we have registered. Or they can be loaded from a **local** filepath that is specified via `checkpoint_path`. 
- `get_custom_sam_model` loads a model from a training checkpoint from our finetuning code. This function is not very relevant for most users, but is important for us to easily evaluate the finetuned models without having to convert them to the default format (which is done by `export_custom_sam_model`)

So these two functions do something conceptually different, and it would also complicate the implementation if `get_custom_sam_model` would accept both types of weight files, because we would then need to find out which of the two formats it is before loading the model.

> * and also if we download weights from different checkpoints then the hash validation is going to break.

We don't want to support downloading weights from arbitrary locations, but only for known models (either original SAM or our finetuned ones). So I don't really see any problems with hash validation here. 
(If a user wants to use some other model available online they can download it themselves and use the `checkpoint_path` argument).

I hope that makes sense to you @GenevieveBuckley.
All of this should probably be documented better, I can follow up with a PR to improve the docstrings after we finish this one.

## test coverage and example scripts

- I have updated the example scripts so that we use `get_cache_directory` in there, see 1c79323
- I have added more tests to extend the coverage for loading models in #280. I will wait for your review before merging since I have written the test for `get_sam_model` according to the current logic.

constantinpape:
Hi @GenevieveBuckley,
I have merged #280 now to increase the test coverage. Most of the changes here should be covered by tests now, so I think you can continue here.
Let me know if something is still unclear with respect to `get_sam_model` vs. `get_custom_sam_model` or if there is some problem in resolving the merge conflicts.

micro_sam/util.py
GenevieveBuckley: Done. 
I'm not entirely sure whether the dock widgets get created once when the plugin is imported into napari, or whether a new instance is created every time we open the dock widget from the plugin menu. So I'm not 100% sure whether changes made by the user while napari is open will get picked up by this, but we've done the best we can now.

micro_sam/util.py
GenevieveBuckley: Ok. I have restored the `checkpoint_path` keyword argument. 

I will need you to test this (because all the example scripts involving checkpoint_path rely on data you personally have). The `_get_checkpoint` function is not covered by any of the existing tests, so it's safe to say we're not testing anything involving `checkpoint_path` right now. 
Could you please run:
* `examples/annotator_with_custom_model.py`
* `examples/finetuning/use_finetuned_model.py`
* `examples/use_as_library/instance_segmentation.py`

It would be great if we could upload the data and custom weights for `examples/annotator_with_custom_model.py` to zenodo or similar - or even make a similar example but with smaller 2d data and model weights. Ideally then anybody could run this example as a test.

micro_sam/util.py
GenevieveBuckley: I'm confused - if the weights are different, the hash value for the object should be different too, right? So how can anybody possibly do some extra fine tuning on a particular model, save the new weights, and then use checkpoint_path and give it the hash value stored in the pooch registry?

I might misunderstand what is actually happening with `checkpoint_path`.

micro_sam/util.py
```diff
@@ -84,48 +49,82 @@
...
+        checkpoint_url = models().urls[model_name]
+        checkpoint_name = models().download_names.get(model_name, checkpoint_url.split("/")[-1])
+        checkpoint_path = os.path.join(microsam_cachedir()/"models", checkpoint_name)
 
         # download the checkpoint if necessary
         if not os.path.exists(checkpoint_path):
-            os.makedirs(_CHECKPOINT_FOLDER, exist_ok=True)
-            _download(checkpoint_url, checkpoint_path, model_type)
+            os.makedirs(microsam_cachedir()/"models", exist_ok=True)
+            pooch.retrieve(url=checkpoint_url, known_hash=models().registry.get(model_name))
```
GenevieveBuckley, line 125: Does `known_hash` need to be None here (i.e. unknown)? I don't quite understand how someone can do fine tuning of model weights, but have the hash stay unchanged (which is I think what would have to happen for what's written here to make any sense).

micro_sam/util.py
```diff
@@ -203,14 +202,18 @@ def get_sam_model(
...
+        checkpoint = models().fetch(model_type)
+    else:
+        checkpoint = _get_checkpoint(model_type, checkpoint_path)
+
     # Our custom model types have a suffix "_...". This suffix needs to be stripped
     # before calling sam_model_registry.
     model_type_ = model_type[:5]
     assert model_type_ in ("vit_h", "vit_b", "vit_l", "vit_t")
-    if model_type == "vit_t" and not VIT_T_SUPPORT:
+    if model_type_ == "vit_t" and not VIT_T_SUPPORT:
```
GenevieveBuckley, line 216: It seems possible that in the future people might fine tune models based off `vit_t`, so we'd have `model_type="vit_t_em"`, etc. 
I find the `model_type`/`model_type_` distinction here confusing, I'd kinda like to rename `model_type` -> `model_name` instead.

micro_sam/util.py
```diff
@@ -219,7 +222,7 @@ def get_sam_model(
     sam = sam_model_registry[model_type_](checkpoint=checkpoint)
     sam.to(device=device)
     predictor = SamPredictor(sam)
-    predictor.model_type = model_type
+    predictor.model_type = model_type_
```
GenevieveBuckley, line 225: Is this correct?

micro_sam/util.py
constantinpape: Hi @GenevieveBuckley,
the models are all available via zenodo already. See https://github.com/computational-cell-analytics/micro-sam/blob/master/micro_sam/util.py#L46-L49. (I agree though that we should list this in the doc, I have added that to #249)
 and I changed the examples in #280 so that all the examples use sample data.

> The `_get_checkpoint` function is not covered by any of the existing tests, so it's safe to say we're not testing anything involving `checkpoint_path` right now.

I don't think that's true either since merging #280. [This](https://github.com/computational-cell-analytics/micro-sam/blob/dev/test/test_util.py#L36-L40) now uses a custom checkpoint. But you would need to merge the current `dev` branch in here first to get those tests.

I will still go ahead and test the examples with this branch.

micro_sam/util.py
constantinpape: Sorry I should have written a better description here. I will try to explain what happens if a `checkpoint_path` is passed or not:
- if `checkpoint_path` is given: this is a local path to sam weigths. E.g. if a user has fine-tuned on their own data. In this case we don't download any weights but initialize the model from these local weights. We do not check the weight file against any hashes.
- if `checkpoint_path` is not given: we download the SAM weights corresponding to `model_type`. (I agree that this name is confusing, see comment below). We can check their hash.

> So how can anybody possibly do some extra fine tuning on a particular model, save the new weights, and then use checkpoint_path and give it the hash value stored in the pooch registry?

When finetuning we don't check the weights against the hash values in the pooch registry.

micro_sam/util.py
constantinpape: Ok, I think I don't fully understand what `pooch` is doing. I will check this out and follow up.

micro_sam/util.py
constantinpape: I fully agree, we should have a better distinction here and the current distinction is confusing. Let's do what you suggest and rename `model_type -> model_name` and `model_type_ -> model_type`

micro_sam/util.py
constantinpape: Yes, that's correct.

constantinpape (approved):

constantinpape:
Thanks @GenevieveBuckley! This cleans up things significantly and I hope now the usage of the `get_sam_model` functions is much clearer.

Issue #269, opened 2023-11-08
Enable setting cachedir from napari


constantinpape:
Interesting thought: maybe we could even make a napari widget for setting the cache directory path to something other than the default value. Then we could rely less on people reading every single page of the API documentation, and have better discoverability.

_Originally posted by @GenevieveBuckley in https://github.com/computational-cell-analytics/micro-sam/issues/268#issuecomment-1801104008_

constantinpape:
Done.

PR #277, opened 2023-11-14
napari widget to set cache directory (for model weights and sample data downloads)

GenevieveBuckley:
Closes https://github.com/computational-cell-analytics/micro-sam/issues/269


constantinpape:
I would like to generalize this a bit already: let's call it `setting_widget`, so that it can be extended later if we want to expose more global settings to users.

Apart from that this looks good, though I think we should first work on #276 to implement the functions that actually update the cachepath.

constantinpape (approved):

constantinpape:
Thanks!

PR #280, opened 2023-11-18
Increase test coverage

constantinpape:
by adding tests for
- `get_sam_model` (with and without local weight paths)
- `get_custom_sam_model`
- `run_inference_with_iterative_prompting`

Test for the latter two are integrated with the training integration test.

### DIFF
--- a/micro_sam/util.py
+++ b/micro_sam/util.py
@@ -49,8 +49,7 @@ _MODEL_URLS = {
     "vit_h_em": "https://zenodo.org/record/8250291/files/vit_h_em.pth?download=1",
     "vit_b_em": "https://zenodo.org/record/8250260/files/vit_b_em.pth?download=1",
 }
-_CACHE_DIR = os.environ.get('MICROSAM_CACHEDIR') or pooch.os_cache('micro_sam')
-_CHECKPOINT_FOLDER = os.path.join(_CACHE_DIR, 'models')
+
 _CHECKSUMS = {
     # the default segment anything models
     "vit_h": "a7bf3b02f3ebf1267aba913ff637d9a2d5c33d3173bb679e46d9f338c26f262e",
@@ -87,7 +86,7 @@ def get_cache_directory() -> None:
 
     Users can set the MICROSAM_CACHEDIR environment variable for a custom cache directory.
     """
-    default_cache_directory = os.path.expanduser(pooch.os_cache('micro-sam'))
+    default_cache_directory = os.path.expanduser(pooch.os_cache('micro_sam'))
     cache_directory = Path(os.environ.get('MICROSAM_CACHEDIR', default_cache_directory))
     return cache_directory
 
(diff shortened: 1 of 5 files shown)


micro_sam/util.py
```diff
@@ -87,7 +86,7 @@ def get_cache_directory() -> None:
 
     Users can set the MICROSAM_CACHEDIR environment variable for a custom cache directory.
     """
-    default_cache_directory = os.path.expanduser(pooch.os_cache('micro-sam'))
+    default_cache_directory = os.path.expanduser(pooch.os_cache('micro_sam'))
```
constantinpape, line 89: The name of the cachedir was different from the previous one in `_CHECKPOINT_FOLDER` (see above).
I think `micro_sam` is better than `micro-sam` (because it's the name of the actual python package) but if there is a reason to switch to `micro-sam` instead I would also be fine with it.

GenevieveBuckley: I don't mind either, so that sounds fine. I don't have any reason for it. I think it's also a good idea to double check there's no remaining `micro-sam` left anywhere else in the code.

GenevieveBuckley (approved):
This looks good to me. Up to 41% test coverage is a big increase, which is great 🎉 

1. I don't know why codecov.io is saying "Token found by environment variables with length 50 did not pass validation" and not actually uploading the results. Given that, it might be a good idea to append `--cov-report term-missing` in addition to `--cov-report xml:coverage.xml` in the pytest options section of `pyproject.toml`
```toml
[tool.pytest.ini_options]
addopts = "-v --durations=10 --cov=micro_sam --cov-report xml:coverage.xml --cov-report term-missing"
```

2. Should the tests in `test/test_instance_segmentation.py` have the "skip" marker removed, or should they stay skipped? No pressure either way.
Turns out that un-skipping the test does not actually increase the overall test coverage that much (from 41% to 44%). They're also slow-ish tests, although not prohibitively so (they do turn up in our slowest 10 test duration list though).
```
==================================== slowest 10 durations =====================================
63.16s call     test/test_training.py::TestTraining::test_training
16.18s call     test/test_instance_segmentation.py::TestInstanceSegmentation::test_tiled_automatic_mask_generator
5.32s call     test/test_gui.py::test_annotator_2d
4.56s call     test/test_instance_segmentation.py::TestInstanceSegmentation::test_automatic_mask_generator
2.42s call     test/test_instance_segmentation.py::TestInstanceSegmentation::test_tiled_embedding_mask_generator
2.39s setup    test/test_instance_segmentation.py::TestInstanceSegmentation::test_automatic_mask_generator
2.23s call     test/test_util.py::TestUtil::test_tiled_prediction
1.54s call     test/test_sam_annotator/test_widgets.py::test_embedding_widget
1.39s call     test/test_instance_segmentation.py::TestInstanceSegmentation::test_embedding_mask_generator
0.85s call     test/test_prompt_based_segmentation.py::TestPromptBasedSegmentation::test_segment_from_mask_non_square
```

micro_sam/util.py
constantinpape: I quickly checked and this was the only location in the code.
(It's still a bit inconsistent in overall naming, but we can address that later on.)

constantinpape:
> I don't know why codecov.io is saying "Token found by environment variables with length 50 did not pass validation" and not actually uploading the results. Given that, it might be a good idea to append --cov-report term-missing in addition to --cov-report xml:coverage.xml in the pytest options section of pyproject.toml

I also don't know why this error is happening, but seems to be due to the unreliability of codecov. I have added the `term-missing` line.

> 2\. Should the tests in `test/test_instance_segmentation.py` have the "skip" marker removed, or should they stay skipped? No pressure either way.
>         Turns out that un-skipping the test does not actually increase the overall test coverage that much (from 41% to 44%). They're also slow-ish tests, although not prohibitively so (they do turn up in our slowest 10 test duration list though).

I want to remove this functionality soonish, so will leave it on skip since these tests do increase the test runtime a bit.
(The functionality implements an alternative approach for automatic instance segmentation that turned out to not work as well. We are now working on a much improved version included in model fine-tuning and once we integrate it in micro_sam I will remove this old functionality.)

PR #286, opened 2023-12-05
Update Rescaling Inputs

anwai98:
WIP - @constantinpape I went ahead and added the changes to the LiveCELL dataloaders (as I had the setup already hehe) (you could take over for the other dataloaders, as discussed).
Thanks!


constantinpape:
@anwai98 I went ahead and implemented an identity transform. For most `torch_em.get_default_dataloader` this has to be used, otherwise it will fall back to standardize.
(This is not the case for the livecell loader as you correctly figured out, but I use the identity there as well for consistency).

In addition I added a check in the training that throws a warning if the data appears to be normalized.

I have not updated all the data loaders for generalist training yet, we can do this once we update those experiments.

PR #297, opened 2023-12-31
New instance segmentation

constantinpape:
- Implement new instance segmentation functionality based on instance segmentation with a specialized decoder.
- Update instance segmentation grid search functionality to support multiple segmentation functions and arbitrary grid search parameters.

This is WIP and pending https://github.com/constantinpape/torch-em/pull/192 and https://github.com/computational-cell-analytics/micro-sam/pull/296.


micro_sam/training/trainable_sam.py
```diff
@@ -52,23 +60,19 @@ def image_embeddings_oft(self, input_images):
...
-            multimask_output: Whether to predict mutiple or just a single mask.
             image_embeddings: The precompute image embeddings. If not passed then they will be computed.
+            multimask_output: Whether to predict mutiple or just a single mask.
 
         Returns:
             The predicted segmentation masks and iou values.
         """
-        input_images = torch.stack([self.preprocess(x=x["image"].to(self.device)) for x in batched_inputs], dim=0)
-        if image_embeddings is None:
-            image_embeddings = self.sam.image_encoder(input_images)
```
anwai98, line 70: Q. For this part, does the updated forward pass of `TrainableSam` now always expect to pass the precomputed image embeddings? (it was a nice optional feature for validation while training, as we don't do iterative prompting there - only batched prompting, hence the need to calculate the image embeddings before isn't very critical)

constantinpape: I think it's better to be consistent here and always compute the image embeddings beforehand. That way we don''t need to duplicate the preprocessing functionality.

anwai98: Yes, makes sense (updated my scripts based on changes made by you). Thanks Constantin!

PR #299, opened 2024-01-02
Fix Resizing Input Prompts

anwai98:


micro_sam/training/sam_trainer.py
```diff
@@ -242,11 +244,17 @@ def _update_masks(self, batched_inputs, y, sampled_binary_y, sampled_ids, num_su
         return loss, mask_loss, iou_regression_loss, mean_model_iou
 
     def _get_updated_points_per_mask_per_subiter(self, masks, sampled_binary_y, batched_inputs, logits_masks):
+        # function to adjust the samples based on the longest side resizing
+        transform = ResizeLongestSide(self.model.sam.image_encoder.img_size)
```
constantinpape, line 248: You don't need this. You can just use `self.model.transform`. That guarantees that we use the same transform everywhere.

micro_sam/training/sam_trainer.py
```diff
@@ -242,11 +244,17 @@ def _update_masks(self, batched_inputs, y, sampled_binary_y, sampled_ids, num_su
...
+
         # here, we get the pair-per-batch of predicted and true elements (and also the "batched_inputs")
         for x1, x2, _inp, logits in zip(masks, sampled_binary_y, batched_inputs, logits_masks):
             # here, we get each object in the pairs and do the point choices per-object
             net_coords, net_labels, _, _ = self.prompt_generator(x2, x1)
 
+            # convert the point coordinates to the expected resolution for iterative prompting
+            net_coords = transform.apply_coords_torch(net_coords, sampled_binary_y.shape[-2:])
+
```
constantinpape, line 257: You probably don't need to update the masks, since they are always passed as lowres (256, 256) (I forgot about this earlier). But I think it's a good idea to leave a comment on this.

micro_sam/training/util.py
```diff
@@ -111,6 +113,8 @@ def _get_prompt_lists(self, gt, n_samples, prompt_generator):
     def __call__(self, x, y, n_pos, n_neg, get_boxes=False, n_samples=None):
         """Convert the outputs of dataloader and prompt settings to the batch format expected by SAM.
         """
+        # function to apply resizing for the input prompts
+        transform = ResizeLongestSide(1024)
```
constantinpape, line 117: Let's not hard-code this. I would suggest passing the transform as an argument to `__init__`. (And then you can use the `model.transform` when instantiating the class for training)

constantinpape (commented):
The logic looks correct. Only a few minor comments on the implementation.

micro_sam/training/sam_trainer.py
anwai98: Makes sense. Updated it in the latest commit. Thanks!

micro_sam/training/util.py
anwai98: Okie. Fixed it as well. Thanks!

constantinpape (approved):

PR #301, opened 2024-01-02
Simplify training implementations

constantinpape:
WIP


constantinpape:
Pending release of https://github.com/constantinpape/torch-em/pull/194

constantinpape:
@anwai98  I have mostly finished this now. The main changes:
- Removing a lot of nested loops. This makes the code more readable and significantly faster (for vit_b with joint trainer: ca 2s/it -> ca. 1.25s/it).
   - Note that this relies on a change in the Dice loss; it's available when updating torch_em. I have also removed the loss and metric as parameters and set them correctly in the init so that this is less error prone. 
- Renamed a couple of functions and variables to make things clearer, removed unnecessary things like `self._sigmoid` (which is always a sigmoid, so we can hard-code it.)
- Implement the box distortion in convert inputs (I think this is correct, but I haven't fully checked it).

I am running a training with this now, if the results look good then I think it should be ok.

Issue #206, opened 2023-09-22
Best design for plugins?


constantinpape:
Currently, the annotation tools are all implemented using separate functions that duplicate quite some functionality.
It would be better to find a design that enables sharing more of the common functionality between them.
There are two different approaches:
- Using a class and inheritance design.
  - It's a bit unclear to us how exactly this would work with magicgui, but maybe we could use https://github.com/hanjinliu/magic-class 
- Use auxiliary functions that refactor the common functionality. This is the current approach, but many things are not refactored yet. The util functions are in https://github.com/computational-cell-analytics/micro-sam/blob/master/micro_sam/sam_annotator/util.py

tlambert03:
> It's a bit unclear to us how exactly this would work with magicgui

also consider dataclasses: https://pyapp-kit.github.io/magicgui/dataclasses/

constantinpape:
We have decided now to go with a "normal" class design and use inheritance, see #304. I decided against dataclasses because the logic for initializing the plugins here is a bit more complex and requires a full fledged init.

Issue #200, opened 2023-09-22
Explore widget containers


constantinpape:
We could use widget containers to group all relevant functionality for a plugin in one container.
See https://pyapp-kit.github.io/magicgui/api/widgets/Container/ for widget containers.

cc @DragaDoncila (@GenevieveBuckley told us that you were also looking into a demo PR for this).

@paulhfu : probably wait to look into this till Draga answers ;).

DragaDoncila:
@constantinpape @paulhfu  👋 thanks for the ping! I have just [opened the PR ](https://github.com/napari/cookiecutter-napari-plugin/pull/168)to the napari plugin cookiecutter adding a `Container` example. Let me know if it seems like it would aid what you're trying to do here, if there's any questions I can answer, or if there's missing features you'd like us to explore!

constantinpape:
We have adopted widget containers in #304. Thanks for the example you have added in the cookiecutter repo @DragaDoncila - this was very helpful for the implementation.

PR #304, opened 2024-01-04
Prototype for annotator plugins

constantinpape:
## Minimal prototype for annotator plugin.

This is an attempt to refactor the current implementations of annotation tools into proper napari plugins.
It contains a minimal working example for the 2d annotation tool (not yet all features added).

## Design

- The base class `sam_annotator._annotator._AnnotatorBase` initializes all shared functionality for the annotation plugins.
- The annotation tools inherit from this class and add their specific functionality, e.g. `sam_annotator.annotator.Annotator2d`.
    - (There we could also implement a function as entry point from python / the CLI so that the tool can be started without going through the napari plugin interface).
- In addition, all widgets will be refactored to `_widgets.py`, and some more clean up. 

This would address #200 and #206.


constantinpape:
Need to update `test/test_gui.py` to match these changes!

constantinpape:
I have now ported all annotation tools to the new plugin design and everything is working!
The overall design approach is summarized here: https://github.com/computational-cell-analytics/micro-sam/blob/b440dcb493502721ff99ba0f4be1077905d0e892/doc/development.md

Next steps:
- Before merging this PR we will merge the current `dev` into `master` and make a `0.4.0` release (to publish the recent improvements to the training methodology).
- The core plugin functionality in this PR is working, but there are a lot of smaller follow-ups to beautify the plugin and extend functionality. We will work on this for version `0.5.0`.
  - This will also include support for the new decoder based instance segmentation in napari, see #305.

micro_sam/sam_annotator/_annotator.py
```diff
@@ -0,0 +1,165 @@
...
+        self._commit_widget = commit_widget()
+        self._clear_widget = clear_widget()
+        widget_list.extend([self._commit_widget, self._clear_widget])
+
+        # Add the widgets to the container.
+        self.extend(widget_list)
+
+    def _create_keybindings(self):
+        @self._viewer.bind_key("s")
+        def _segmet(viewer):
```
GenevieveBuckley, line 83:
```suggestion
        def _segment(viewer):
```

Small typo here - will also need a quick find/replace on any other locations in the code that use this function

micro_sam/sam_annotator/_annotator.py
```diff
@@ -0,0 +1,165 @@
...
+        @self._viewer.bind_key("Shift-C")
+        def _clear_annotations(viewer):
+            self._clear_widget(viewer)
+
+        if hasattr(self, "_segment_nd_widget"):
+            @self._viewer.bind_key("Shift-S")
+            def _seg_nd(viewer):
+                self._segment_nd_widget(viewer)
+
+    # TODO more clever way to integrate segmentation result so
```
GenevieveBuckley, line 103: Suggest creating a dedicated issue for this suggestion, and removing the code comment here

micro_sam/sam_annotator/_annotator.py
```diff
@@ -0,0 +1,165 @@
...
+if TYPE_CHECKING:
+    import napari
+
+
+class _AnnotatorBase(Container):
+    """Base class for micro_sam annotation plugins.
+
+    Implements the logic for the 2d, 3d and tracking annotator.
+    The annotators differ in their data dimensionality and the widgets.
+    """
```
GenevieveBuckley, line 19: I noticed you using `@private` in some docstrings elsewhere to stop them appearing in the API docs, should that happen here too? I have no big opinion either way.

micro_sam/sam_annotator/util.py
```diff
@@ -461,3 +406,157 @@ def _initialize_parser(description, with_segmentation_result=True, with_show_emb
     )
 
     return parser
+
+
+#
+# Helper functions for tracking annotator
```
GenevieveBuckley, line 412: Would it make more sense to put all of the tracking annotator content either in `annotator_tracking.py`, or if it has to be a different file how about a `util_tracking.py` file? This seems like an odd place to put it.

examples/annotator_2d.py
```diff
@@ -25,7 +25,7 @@ def livecell_annotator(use_finetuned_model):
         embedding_path = os.path.join(EMBEDDING_CACHE, "embeddings-livecell.zarr")
         model_type = "vit_h"
 
-    annotator_2d(image, embedding_path, show_embeddings=False, model_type=model_type)
```
GenevieveBuckley, line 28: 1. If the `show_embedding` kwarg is removed, does that mean there is documentation that needs to be changed? I don't remember if there are docs pages specifically about displaying a visualization of the embedding.

2. Does this mean there is now a separate function for displaying embeddings, or is it removed altogether? I don't have an opinion here (it's not a feature I've ever really used), just checking what the status is

GenevieveBuckley (approved):
This class based creation of dock widgets is a bit more heavy duty that the approach I tried (involving a magicgui container stuffed with the existing widgets). I hope/think that will make it more robust.

I'm broadly supportive of this design. Note that I've mostly looked at the code, and haven't spent time playing with it in the GUI, eg: to check things like whether the embeddings don't get recomputed unnecessarily, or that computing embeddings, then changing the selected input image produces something sensible, etc.

We've discussed this previously, and I still agree that refactoring to reduce duplication (eg: combining the 2D and 3D annotation widgets, etc.) is a good long term goal. Not in this PR, but something later. I think that structure you've made with this design will probably help with that long term goal. It seems a little easier to understand how the pieces fit together here.

test/test_sam_annotator/test_annotator_2d.py
```diff
@@ -0,0 +1,54 @@
+import platform
```
GenevieveBuckley, line 1: Should there be a `test_image_series_annotator.py` file to accompany the others in this folder?

test/test_sam_annotator/test_annotator_2d.py
```diff
@@ -0,0 +1,54 @@
...
+
+import numpy as np
+import pytest
+from skimage.data import binary_blobs
+
+import micro_sam.util as util
+from micro_sam.sam_annotator import annotator_2d
+
+
+def _check_layer_initialization(viewer, expected_shape):
```
GenevieveBuckley, line 11: Could this function be refactored? It appears in every `test_...` file included with this PR. If they are all almost identical, we could make a util file for the tests. 

If the expected layer names change slightly (does that happen for the tracking annotator? idk, maybe) then that could be passed in as a kwarg to the function.

micro_sam/sam_annotator/_annotator.py
constantinpape: Nice catch, thanks! And this is just a local function for the key bindings so we don't need to fix it anywhere else. (I also did a search for `segmet` and this is the only place where the typo occurs.)

micro_sam/sam_annotator/_annotator.py
constantinpape: Yes absolutely! I will remove the TODO's and create dedicated issues before merging this PR.

micro_sam/sam_annotator/_annotator.py
constantinpape: I think this is not needed here because this is in a file with prefix `_` so it will be ignored by pydoc already.

micro_sam/sam_annotator/util.py
constantinpape: Yes, I agree. I also thought about this when refactoring the code. I will see what makes most sense and move this code.

examples/annotator_2d.py
constantinpape: In general: I decided to remove this functionality to reduce the complexity because I also don't think it's particularly useful on the user end.

Re 1.: I don't think there was any documentation on this that would need to be updated.

Re 2.: There is no convenience function to display the emebddings any more. There is still the implementation for it in https://github.com/computational-cell-analytics/micro-sam/blob/master/micro_sam/visualization.py. For now I would leave it at that. I think displaying the embeddings is unnecessary for plugin users, so we don't need to expose this functionality to them. On the other hand it can be quite helpful for debugging and method development, so I don't want to fully remove the functionality.

test/test_sam_annotator/test_annotator_2d.py
constantinpape: Good point, I will see if I can add it.

examples/annotator_2d.py
constantinpape: Re updating the documentation in general: of course this will need quite some updates to the documentation, but we should work on that separately (there's some more changes necessary beforehand, I will create follow-up issues after merging this one.)

test/test_sam_annotator/test_annotator_2d.py
constantinpape: Yes, good point. I will refactor it. (And I decided to keep the same layernames for the tracking annotator; that requires less changes in the code and I also think it's easier from a user perspective.)

constantinpape:
Thanks for the review @GenevieveBuckley !

> I'm broadly supportive of this design. Note that I've mostly looked at the code, and haven't spent time playing with it in the GUI, eg: to check things like whether the embeddings don't get recomputed unnecessarily, or that computing embeddings, then changing the selected input image produces something sensible, etc.

I have tried at each tool a couple of times and everything seems to work. Of course there might still be some more subtle bugs or issue (e.g. that everything is set correctly when updating the image etc.).
I will do some more testing of this later, and we will also test it more in the lab before doing an actual release.

> We've discussed this previously, and I still agree that refactoring to reduce duplication (eg: combining the 2D and 3D annotation widgets, etc.) is a good long term goal. Not in this PR, but something later. I think that structure you've made with this design will probably help with that long term goal. It seems a little easier to understand how the pieces fit together here.

Yes, there's probably some more room to combine functionality here in follow up work. 

Regarding your more in-detail review and next steps:
- I have addressed the comments about the tracking util functions and the tests.
- I will make dedicated issues out of TODOs when finalizing this PR (and open a few more follow up issues).
- And I will do that as soon as we have made a new release with the updated training functionality and new fine-tuned models. (Or have merged the current `dev` into `master` as preparation for this, hopefully by the end of this week.)

examples/annotator_2d.py
GenevieveBuckley: > In general: I decided to remove this functionality to reduce the complexity because I also don't think it's particularly useful on the user end.

Agree, I support your decision here

GenevieveBuckley:
> I have tried each tool a couple of times and everything seems to work. Of course there might still be some more subtle bugs or issue (e.g. that everything is set correctly when updating the image etc.).

That's great! 😄 Yes, of course unexpected issues can always pop up later. It sounds like this is a good balance between doing enough real world testing to catch any obvious problems, but not too much that holds the project back.

micro_sam/sam_annotator/_state.py
```diff
@@ -57,14 +99,16 @@ def initialized_for_interactive_segmentation(self):
...
+        have_tracking_widget = self.tracking_widget is not None
+        init_sum = sum((have_current_track_id, have_lineage, have_committed_lineages, have_tracking_widget))
+        if init_sum == 4:
             return True
         elif init_sum == 0:
             return False
         else:
             raise RuntimeError(
-                f"Invalid AnnotatorState: {init_sum} / 2 parts of the state "
+                f"Invalid AnnotatorState: {init_sum} / 4 parts of the state "
```
GenevieveBuckley, line 111: If I saw this error message, I would still be a bit confused about what I needed to change to make things work again. Perhaps it could also tell users exactly which parts are not initialized correctly (out of `have_current_track_id, have_lineage`, `have_committed_lineages`, `have_tracking_widget`)

GenevieveBuckley:
I see this PR is now out of draft state. I don't have any substantive new comments to add since our last discussion.

constantinpape:
> I see this PR is now out of draft state. I don't have any substantive new comments to add since our last discussion.

Thanks @GenevieveBuckley. I moved this out of draft because I have merged `dev` into `master` now and made a new release and wanted to wait with merging this one until that.
I will clean up and merge this PR today.

micro_sam/sam_annotator/_annotator.py
constantinpape: I have added dedicated issues for all the TODOs now. (I decided to leave the actual TODOs in the code and link the issue for convenience.)

micro_sam/sam_annotator/_state.py
constantinpape: I have updated the error messages to print the variables that are actually not initialized.

PR #305, opened 2024-01-06
Implement new segmentation plugin

constantinpape:
Pending #304


micro_sam/sam_annotator/image_series_annotator.py
```diff
@@ -63,12 +63,21 @@ def image_series_annotator(
         return_viewer: Whether to return the napari viewer to further modify it before starting the tool.
         predictor: The Segment Anything model. Passing this enables using fully custom models.
             If you pass `predictor` then `model_type` will be ignored.
+<<<<<<< HEAD
```
GenevieveBuckley, line 66: Just FYI there's the remnants of a merge conflict that hasn't been totally resolved still in here 😄

constantinpape: Thanks for catching this @GenevieveBuckley !

constantinpape:
This now implements:
- Instance segmentation with decoder
- Automatic 3d instance segmentation (both with decoder and AMG)
- State precomputation for AMG and IS with decoder

There are still several improvements necessary, but I will merge this now and then work on the improvements as a follow-up.

PR #306, opened 2024-01-07
Update LM Generalist Dist. Trafo.

anwai98:


micro_sam/training/util.py
```diff
@@ -212,14 +212,15 @@ def __call__(self, raw):
 
 
 class ResizeLabelTrafo:
-    def __init__(self, desired_shape, padding="constant"):
+    def __init__(self, desired_shape, padding="constant", min_size=25):
```
constantinpape, line 215: Let's set the default to zero here too, so that we can avoid the bug with empty labels.

anwai98: Done.

constantinpape (approved):

PR #310, opened 2024-01-08
Add LM Specialist Evaluation

anwai98:


anwai98:
GTG from my side! (not sure why the tests are failing here though)

constantinpape:
> (not sure why the tests are failing here though)

Yes, something unrelated broke the tests, I will look into this in the evening.

micro_sam/evaluation/util.py
```diff
@@ -0,0 +1,43 @@
+import os
+from typing import Union, Optional, List
+
+from . import instance_segmentation, inference
+from ..instance_segmentation import AutomaticMaskGenerator
+
+
+def run_amg(
```
constantinpape, line 8: This function seems to be identical to https://github.com/computational-cell-analytics/micro-sam/blob/7d6010fb95533d5293769b32131a5be8d674e519/micro_sam/evaluation/inference.py#L528

Can we remove it (including this `util.py`) and use the `run_amg` from there instead?

anwai98: Yes, forgot to remove that. Fixed. Thanks for spotting!

PR #317, opened 2024-01-15
Add livecell specialist results

anwai98:


anwai98:
Moved the scripts to the `scripts` from here as well.

PR #319, opened 2024-01-15
Add livecell (lm) generalist results

anwai98:


anwai98:
I have moved the results to the destination as discussed (thought of a simple naming convention ~same as in experiment folders from inferences, will come up with something particular later)

PR #320, opened 2024-01-16
Add EM generalist evaluation

anwai98:


anwai98:
GTG from my side!

micro_sam/evaluation/instance_segmentation.py
anwai98, line 1: @constantinpape I found a minor issue in the instance segmentation. (in short, when there are no objects in the instance segmentation predicted, the regionprops returns no properties, which leads to some issues - to fix this, I check while we find no objects, we just save an image with all zeros)

anwai98:
@constantinpape This PR is good to go from my side. I would update some parts later (like, choice for making the split between val-test, path convention), for now this is ready with it's first initial evaluation setup. Let me know if you spot something.

micro_sam/evaluation/instance_segmentation.py
constantinpape: Good catch, just a few minor potential issues with the implementation

micro_sam/evaluation/instance_segmentation.py
```diff
@@ -275,7 +275,11 @@ def run_instance_segmentation_inference(
 
         segmenter.initialize(image, image_embeddings)
         masks = segmenter.generate(**generate_kwargs)
-        instances = mask_data_to_segmentation(masks, with_background=True, min_object_size=min_object_size)
+
+        if len(masks) == 0:  # the instance segmentation can have no masks, hence we just save empty labels
+            instances = np.zeros_like(image)
```
constantinpape, line 280: I think this may not do the right thing. The image may have multiple channels, but we always want this to be 2d.
You can get the correct image shape from the predictor. I think the following should work (but not tested, so you may need to change a few variable names):
```python
shape = tuple(segmenter._predictor.img_size)  # not sure if it's called img_size, maybe the name is different
image = np.zeros(shape, dtype="unit32")
```

constantinpape (commented):
Looks good overall. Just some minor changes in `evaluation/instance_segmentation.py` needed.

Also, I didn't fully review `preprocess_em_datasets.py`. We should visually check the exported data at some point to make sure this is really correct.

micro_sam/evaluation/instance_segmentation.py
anwai98: Makes sense. Fixed it! Thanks for spotting.

constantinpape (approved):

PR #323, opened 2024-01-17
Minor fix to experiment path convention

anwai98:


anwai98:
This addresses a minor fix. Gtg from my side (not sure why the mac build / test fails)

constantinpape (approved):

PR #340, opened 2024-01-18
Add Lucchi amg results

anwai98:


constantinpape:
Looks good, but let's make the PRs onto dev again.

PR #379, opened 2024-02-05
Update LM generalist training (with new decoder setup)

anwai98:


anwai98:
@constantinpape GTG from my side! (the defaults are now set to make use of conv transpose for upsampling in all relevant parts, as discussed)

constantinpape (approved):

PR #386, opened 2024-02-09
Minor update to finetuning scripts

anwai98:


anwai98:
GTG from my side!

finetuning/.gitignore
```diff
@@ -4,3 +4,4 @@ sam_embeddings/
 results/
 iterative_prompting_results/
 *.sh
+run_all_finetuning.py
```
constantinpape, line 7: Why do we add this to the gitignore? If it's a script shouldn't we just add it?

constantinpape (commented):
Looks good overall. Just a minor question why we exclude one of the scripts.

finetuning/.gitignore
anwai98: I wasn't sure if I should add it. It's a script to run all the finetunings at once. I added it just now!

PR #393, opened 2024-02-15
Add inference for cellpose 3.0

anwai98:


anwai98:
@constantinpape GTG from my side! (adding `cyto3` with all the other available models, in case we want to use it in future)

constantinpape (approved):

PR #404, opened 2024-02-29
Update JointSamTrainer's `save_checkpoint`

anwai98:
Pending https://github.com/constantinpape/torch-em/pull/223

### DIFF
--- a/micro_sam/training/joint_sam_trainer.py
+++ b/micro_sam/training/joint_sam_trainer.py
@@ -33,7 +33,9 @@ class JointSamTrainer(SamTrainer):
                 decoder_state.append((k, v))
         decoder_state = OrderedDict(decoder_state)
 
-        super().save_checkpoint(name, current_metric, best_metric, decoder_state=decoder_state, **extra_save_dict)
+        super().save_checkpoint(
+            name, current_metric=current_metric, best_metric=best_metric, decoder_state=decoder_state, **extra_save_dict
+        )
 
     def load_checkpoint(self, checkpoint="best"):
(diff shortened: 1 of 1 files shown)


micro_sam/training/joint_sam_trainer.py
```diff
@@ -25,15 +25,15 @@ def __init__(
...
+    def save_checkpoint(self, name, current_metric, best_metric, **extra_save_dict):
         current_unetr_state = self.unetr.state_dict()
         decoder_state = []
         for k, v in current_unetr_state.items():
             if not k.startswith("encoder"):
                 decoder_state.append((k, v))
         decoder_state = OrderedDict(decoder_state)
 
-        super().save_checkpoint(name, best_metric, decoder_state=decoder_state, **extra_save_dict)
+        super().save_checkpoint(name, current_metric, best_metric, decoder_state=decoder_state, **extra_save_dict)
```
constantinpape, line 36: Can you pass `current_metric` and `best_metric` by keyword here? I think the current code (passing them as positional arguments) will lead to a wrong result until we have the new version of `torch_em` here.

constantinpape: And passing the arguments by keyword will work with the new and old version.

constantinpape (changes requested):

micro_sam/training/joint_sam_trainer.py
anwai98: Fixed. Thanks for spotting!

constantinpape (approved):

PR #411, opened 2024-03-05
Minor fix to inference function for instance segmentation

anwai98:


anwai98:
@constantinpape GTG from my side (not sure why one of the macos runners fail)

constantinpape (approved):

PR #419, opened 2024-03-08
331 add options for tiling in the embdding widget

lufre1:
@constantinpape 
looks like there are no merge conflicts

### DIFF
--- a/micro_sam/sam_annotator/_widgets.py
+++ b/micro_sam/sam_annotator/_widgets.py
@@ -131,12 +131,51 @@ def create_prompt_menu(points_layer, labels, menu_name="prompt", label_name="lab
 
     return label_widget
 
+def _process_tiling_inputs(tile_shape_x, tile_shape_y, halo_x, halo_y):
+    tile_shape = (tile_shape_x, tile_shape_y)
+    halo = (halo_x, halo_y)
+    # check if tile_shape/halo are not set: (0, 0)
+    if isinstance(tile_shape, tuple):
+        if all(item == 0 for item in tile_shape):
+            tile_shape = None
+        # check if at least 1 param is given
+        elif tile_shape[0] != 0 or tile_shape[1] != 0:
+            max_val = max(tile_shape[0], tile_shape[1])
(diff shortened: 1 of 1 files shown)


micro_sam/sam_annotator/_widgets.py
```diff
@@ -131,12 +131,51 @@ def label_changed(new_label):
...
+        if all(item == 0 for item in tile_shape):
+            tile_shape = None
+        # check if at least 1 param is given
+        elif tile_shape[0] != 0 or tile_shape[1] != 0:
+            max_val = max(tile_shape[0], tile_shape[1])
+            if max_val < 256: # at least tile shape >256
+                max_val = 256
+            tile_shape = (max_val, max_val)
+        # if both inputs given, check if smaller than 256
+        elif tile_shape[0] != 0 and tile_shape[1] != 0:
```
lufre1, line 148: Sadly, I had to settle for this solution to set a minium value for tile_shape, because limiting the range to desired values broke the input mask resulting in unwanted values in tile_shape

constantinpape: That's ok. However, we will never run into this condition, because the `or` one above will also be triggered by it.

micro_sam/sam_annotator/_widgets.py
```diff
@@ -131,12 +131,51 @@ def label_changed(new_label):
...
+            if tile_shape[0] < 256:
+                tile_shape[0] = 256
+            elif tile_shape[1] < 256:
+                tile_shape[1] = 256 
+    if isinstance(halo, tuple):
```
constantinpape, line 153: I think we can also get rid of this if.

constantinpape (commented):
The overall logic looks good. But I think the logic in `_process_tiling_inputs` is not quite correct yet, and can overall be simplified.

micro_sam/sam_annotator/_widgets.py
lufre1: Implemented the changes according to suggestions on latest commit.

constantinpape (approved):
Thanks!

PR #420, opened 2024-03-10
Update supplementary experiments

anwai98:


anwai98:
@constantinpape This is GTG from my side!

constantinpape (approved):

PR #423, opened 2024-03-11
Add automatic and interactive segmentation for 3d datasets

anwai98:


anwai98:
@constantinpape This is GTG from my side! (sorry for the super long commit trace)
PS: The results are at `micro_sam/scripts/volumetric_evaluation/results.md`

PR #425, opened 2024-03-11
Update seg_with_decoder_2d.py - make inputs modular

anwai98:


anwai98:
I'll close this, as this is only for internal development purpose.

PR #428, opened 2024-03-12
Parallelize operations in commit

constantinpape:
to speed this up for large volumes.


constantinpape:
An alternative / additional improvement for 3d would be to keep track of the slices (during the interactive segmentation) and then use these to restrict all subsequent operations. That's conceptually not so difficult, but would need changes in a couple more spaces that don't make sense before merging things onto dev.

constantinpape:
Closed in favor of #436

PR #429, opened 2024-03-12
426 enable to hold tiled embeddings in memory

lufre1:
used [zarr.Group with in memory store](https://zarr.readthedocs.io/en/stable/api/storage.html#zarr.storage.MemoryStore) to utilize same behaviour like zarr.open() (function returns zarr.Group with directory store). 
Passed the zarr.Group(store=MemoryStore) to the _preprocessing functions that do the tiling etc.
Additionally, there was an unreachable code snippet: https://github.com/computational-cell-analytics/micro-sam/blob/426-enable-to-hold-tiled-embeddings-in-memory/micro_sam/util.py#L674
Made it reachable again


micro_sam/util.py
```diff
@@ -644,46 +646,47 @@ def precompute_image_embeddings(
...
-                        f"Embeddings file {save_path} is invalid due to unmatching {key}: "
-                        f"{f.attrs.get(key)} != {val}.Please recompute embeddings in a new file."
-                    )
-                    if wrong_file_callback is not None:
-                        save_path = wrong_file_callback(save_path)
-                        f = zarr.open(save_path, "a")
-                    break
-
+    else:
+        f = zarr.group()
```
constantinpape, line 656: I didn't know that it's so easy to get a in-memory zarr. That's neat!

micro_sam/util.py
```diff
@@ -644,46 +646,47 @@ def precompute_image_embeddings(
...
+    data_signature = _compute_data_signature(input_)
```
constantinpape, line 657: We should only do this if we actually store the embeddings to a zarr file. Otherwise we do some unnecessary computations.

micro_sam/util.py
```diff
@@ -644,46 +646,47 @@ def precompute_image_embeddings(
...
+            break
+
+    for key, val in key_vals:
+        if key not in f.attrs:
+            f.attrs[key] = val
 
     if ndim == 2:
-        image_embeddings = _compute_2d(input_, predictor) if save_path is None else\
-            _precompute_2d(input_, predictor, save_path, tile_shape, halo)
+        image_embeddings = _compute_2d(input_, predictor) if tile_shape is None else\
```
constantinpape, line 684: I think we are not caching the embeddings any more for the case `save_path is not None and tile_shape is None`!
This probably also explains the test error you see.

constantinpape (changes requested):
This goes in the right direction, but introduces some new problems.

It may be a good idea to change some of the logic, let's discuss together tomorrow.

constantinpape:
Once this is working we should address this FIXME: https://github.com/computational-cell-analytics/micro-sam/blob/dev/test/test_prompt_based_segmentation.py#L24

micro_sam/util.py
lufre1: the concerns should be addressed in the latest commit on.

constantinpape (commented):
This resolves the immediate issues.

constantinpape (approved):
I went ahead and finished this up, otherwise it would get too outdated. It should all work now.

PR #439, opened 2024-03-14
Update inference and evaluation top-level scripts (for LIVECell)

anwai98:


micro_sam/evaluation/evaluation.py
```diff
@@ -75,4 +74,55 @@ def run_evaluation(
     return results
 
 
+def run_evaluation_for_interactive_prompting(
```
constantinpape, line 77: Let's rename this to `run_evaluation_for_iterative_prompting`. We call this `iterative_prompting` in all other functions, so it would be confusing to change the naming convention only here.

constantinpape (commented):
Looks good! Just one minor comment.

micro_sam/evaluation/evaluation.py
constantinpape: I am not sure if you're using this anywhere else in this PR, if yes you of course also need to adapt the name there.

micro_sam/evaluation/livecell.py
```diff
@@ -238,137 +316,103 @@ def run_livecell_instance_segmentation_with_decoder(
...
+    # 2. interactive instance segmentation (interactive)
+    #     - iterative prompting
+    #         - starting with point
+    #         - starting with box
+    # 3. automatic segmentation (auto)
+    #     - automatic mask generation (amg)
+    #     - automatic instance segmentation (ais)
+
+    parser.add_argument("-p", "--precompute_embeddings", action="store_true")
+    parser.add_argument("-ip", "--interactive_segmentation", action="store_true")
```
anwai98, line 366: @constantinpape Just to be sure, I call the iterative prompt-based inference as interactive segmentation (to keep the name simple and to-the-point). Does this work, or should I update this to iterative prompting as well?

constantinpape (approved):

micro_sam/evaluation/livecell.py
anwai98: Fixed it. Thanks!

PR #449, opened 2024-03-15
Update resource efficient finetuning plots

anwai98:


anwai98:
@constantinpape GTG from my side!

anwai98:
Once the test are passed, I will go ahead and merge this. This script has some minor updates to the resource efficient finetuning plots.

PR #452, opened 2024-03-15
Minor fix to launching annotator using custom checkpoints

anwai98:
This led to errors (deserializing issues from default Segment Anything's way of [initializing](https://github.com/facebookresearch/segment-anything/blob/main/segment_anything/build_sam.py#L105) checkpoints) while using custom checkpoints out-of-the-box.

### DIFF
--- a/micro_sam/sam_annotator/_state.py
+++ b/micro_sam/sam_annotator/_state.py
@@ -75,7 +75,8 @@ class AnnotatorState(metaclass=Singleton):
         assert ndim in (2, 3)
         # Initialize the model if necessary.
         if predictor is None:
-            self.predictor = util.get_custom_sam_model(
+            get_model = util.get_sam_model if checkpoint_path is None else util.get_custom_sam_model
+            self.predictor = get_model(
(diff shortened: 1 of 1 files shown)


constantinpape:
Now the code won't work for SAM checkpoints anymore. So for now I don't want to support using the `torch_em` trainer checkpoints. Instead convert it to just the model weights via https://github.com/computational-cell-analytics/micro-sam/blob/master/micro_sam/util.py#L355.

micro_sam/sam_annotator/_state.py
```diff
@@ -75,7 +75,8 @@ def initialize_predictor(
         assert ndim in (2, 3)
         # Initialize the model if necessary.
         if predictor is None:
-            self.predictor = util.get_sam_model(
+            get_model = util.get_sam_model if checkpoint_path is None else util.get_custom_sam_model
```
anwai98, line 78: I was thinking this line would be the one taking care of initializing the models flexlibly, but I see your point. It makes to stay consistent.

PR #457, opened 2024-03-18
Add convenience function for training and add user study scripts

constantinpape:


constantinpape:
@anwai98 there are still some changes left here, but if you have time you could check if everything in the new high-level training function in https://github.com/computational-cell-analytics/micro-sam/blob/user-study/micro_sam/training/training.py looks allright to you.

anwai98:
Looks good to me. I have a few questions, will mention them below.

micro_sam/training/training.py
```diff
@@ -0,0 +1,117 @@
...
+        )
+    else:
+        trainer = trainers.SamTrainer(
+            name=name, train_loader=train_loader, val_loader=val_loader, model=model,
+            optimizer=optimizer, device=device, lr_scheduler=scheduler, logger=trainers.SamLogger,
+            log_image_interval=100, mixed_precision=True, convert_inputs=convert_inputs,
+            n_objects_per_batch=n_objects_per_batch, n_sub_iteration=n_sub_iteration, compile_model=False,
+            early_stopping=early_stopping, mask_prob=0.5, save_root=save_root,
+        )
+    trainer.fit(epochs=n_epochs)
```
anwai98, line 117: I think it would might make sense to expose the iterations parameter (as we finetune based on it and mention it in the paper the number of iterations we train for). The logic might get a bit complicated, but we can accept `**kwargs` in the `train_sam` function, and overwrite the use of epochs with iterations when passed by the user. What do you think?

micro_sam/training/training.py
```diff
@@ -0,0 +1,117 @@
...
+            instance_loss=instance_seg_loss, instance_metric=instance_seg_loss,
+            early_stopping=early_stopping, mask_prob=0.5, save_root=save_root,
+        )
+    else:
+        trainer = trainers.SamTrainer(
+            name=name, train_loader=train_loader, val_loader=val_loader, model=model,
+            optimizer=optimizer, device=device, lr_scheduler=scheduler, logger=trainers.SamLogger,
+            log_image_interval=100, mixed_precision=True, convert_inputs=convert_inputs,
+            n_objects_per_batch=n_objects_per_batch, n_sub_iteration=n_sub_iteration, compile_model=False,
+            early_stopping=early_stopping, mask_prob=0.5, save_root=save_root,
```
anwai98, line 115: Would it also make sense to expose the mask probability ratio? (we don't really need it as much anywhere, but just in case)

micro_sam/training/training.py
```diff
@@ -0,0 +1,117 @@
...
+                joint_model_params.append(params)
+
+        optimizer = torch.optim.Adam(joint_model_params, lr=lr)
+
+    else:
+        optimizer = torch.optim.Adam(model.parameters(), lr=lr)
+
+    if scheduler is None:
+        scheduler = torch.optim.lr_scheduler.ReduceLROnPlateau(
+            optimizer, mode="min", factor=0.9, patience=3, verbose=True
```
anwai98, line 95: I think these parameters can be exposed as an optional parameter, where the defaults could be the values as mentioned here, and can be updated using something like `scheduler_kwargs`.

constantinpape: I think this makes it too complex. The scheduler can already be passed, so I think it's better not to introduce any `scheduler_kwargs` here. If a different scheduling behavior is wanted the user can pass their own scheduler.

anwai98: Oh yes, you are right. Didn't spot it hehe. Then that should do the trick!

constantinpape:
Thanks for the comments @anwai98. I addressed most of it. I will go ahead and merge this once the tests pass.
(There are still some things to do, but we need the current changes to continue with the user study.)

PR #464, opened 2024-03-19
Fix issue with segmentation for non-square images and extend tests

constantinpape:


constantinpape:
Still need to add more tests

PR #467, opened 2024-03-20
Training UI

constantinpape:


constantinpape:
TODO: Enable selecting both files and folders (need to check how this works with magicgui).
See #473.

constantinpape:
@anwai98 : this implements the finetuning UI and is mostly done now. As discussed it would be good to test it a bit and think about how we name things in the UI.

@lufre1 : just fyi, this updates the `_add_path_param` a bit, but the changes are fully compatible with the previous design.

constantinpape:
I decided to go ahead and merge this so that it's easier to test and so that the fix for the file dialog is on dev

PR #470, opened 2024-03-20
Update livecell finetuning and evaluation scripts

anwai98:
This addresses the following:
- Fixes a bug in the `livecell_evaluation.py` top-level evaluation script.
- Refactors the `livecell_finetuning.py` top-level finetuning script, to make it modular
    - NOTE: This only unifies the floating snippets for training, and does not change any components.
- Refactors the `train_sam` function to expose some optional arguments.

### DIFF
--- a/finetuning/livecell_finetuning.py
+++ b/finetuning/livecell_finetuning.py
@@ -1,3 +1,4 @@
+import os
 import argparse
 
 import torch
@@ -6,6 +7,7 @@ from torch_em.data.datasets import get_livecell_loader
 from torch_em.transform.label import PerObjectDistanceTransform
 
 import micro_sam.training as sam_training
+from micro_sam.util import export_custom_sam_model
 
 
 def get_dataloaders(patch_shape, data_path, cell_type=None):
@@ -72,6 +74,14 @@ def finetune_livecell(args):
         export_path=args.export_path,
         save_every_kth_epoch=args.save_every_kth_epoch,
     )
+
+    if args.export_path is not None:
+        checkpoint_path = os.path.join(
+            "" if args.save_root is None else args.save_root, "checkpoints", checkpoint_name, "best.pt"
+        )
+        export_custom_sam_model(
+            checkpoint_path=checkpoint_path, model_type=model_type, save_path=args.export_path,
+        )
 
 
 def main():
--- a/finetuning/livecell_finetuning.py
+++ b/finetuning/livecell_finetuning.py
@@ -54,7 +54,7 @@ def finetune_livecell(args):
 
     # all the stuff we need for training
     train_loader, val_loader = get_dataloaders(patch_shape=patch_shape, data_path=args.input_path)
-    scheduler_params = {"patience": 10}
+    scheduler_kwargs = {"mode": "min", "factor": 0.9, "patience": 10, "verbose": True}
 
     # Run training.
     sam_training.train_sam(
@@ -70,8 +70,7 @@ def finetune_livecell(args):
         lr=1e-5,
         n_iterations=args.iterations,
         save_root=args.save_root,
-        scheduler_kwargs=scheduler_params,
-        export_path=args.export_path,
+        scheduler_kwargs=scheduler_kwargs,
         save_every_kth_epoch=args.save_every_kth_epoch,
     )
 
--- a/finetuning/livecell_finetuning.py
+++ b/finetuning/livecell_finetuning.py
@@ -1,15 +1,11 @@
-import os
 import argparse
 
 import torch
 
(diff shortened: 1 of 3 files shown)


micro_sam/training/training.py
```diff
@@ -93,11 +92,13 @@ def train_sam(
     freeze: Optional[List[str]] = None,
     device: Optional[Union[str, torch.device]] = None,
     lr: float = 1e-5,
-    scheduler: Optional[_LRScheduler] = None,
     n_sub_iteration: int = 8,
     save_root: Optional[Union[str, os.PathLike]] = None,
     mask_prob: float = 0.5,
     n_iterations: Optional[int] = None,
+    scheduler_kwargs: Optional[Dict[str, Any]] = None,
```
constantinpape, line 99: It's correct to remove the scheduler since it needs the optimizer, but it would be nice to enable another type of scheduler. We could introduce another argument `scheduler_class`, which defaults to `ScheduleLROnPlateau`, but makes the choice of scheduler fully flexible.

micro_sam/training/training.py
```diff
@@ -93,11 +92,13 @@ def train_sam(
...
+    export_path: Optional[Union[str, os.PathLike]] = None,
```
constantinpape, line 100: I don't want to integrate export here, because we will have more export options once the modelzoo export is integrated.

micro_sam/training/training.py
```diff
@@ -163,28 +168,73 @@ def train_sam(
...
+        trainer_fit_params = {"epochs": n_epochs}
+    else:
+        trainer_fit_params = {"iterations": n_iterations}
+
+    if save_every_kth_epoch is not None:
+        trainer_fit_params["save_every_kth_epoch"] = save_every_kth_epoch
+
+    trainer.fit(**trainer_fit_params)
+
+    if export_path is not None:
```
constantinpape, line 234: Let's keep the export separate, see above.

finetuning/livecell_finetuning.py
```diff
@@ -52,79 +48,30 @@ def finetune_livecell(args):
...
+        device=device,
+        lr=1e-5,
+        n_iterations=args.iterations,
+        save_root=args.save_root,
+        scheduler_kwargs=scheduler_params,
+        export_path=args.export_path,
+        save_every_kth_epoch=args.save_every_kth_epoch,
     )
-    trainer.fit(args.iterations, save_every_kth_epoch=args.save_every_kth_epoch)
-    if args.export_path is not None:
```
constantinpape, line 119: Let's keep the export here.

micro_sam/training/training.py
anwai98: I agree, makes sense to expose it as well. Updated it.

micro_sam/training/training.py
```diff
@@ -163,28 +170,65 @@ def train_sam(
     else:
         optimizer = torch.optim.Adam(model.parameters(), lr=lr)
 
-    if scheduler is None:
-        scheduler = torch.optim.lr_scheduler.ReduceLROnPlateau(
-            optimizer, mode="min", factor=0.9, patience=3, verbose=True
-        )
+    scheduler_params = {"mode": "min", "factor": 0.9, "patience": 3, "verbose": True}
```
constantinpape, line 173: This will fail if we have a different scheduler class, because it will also get these params.
Let's simplify this:
```
if scheduler_kwargs is None:
  schedulder_kwargs = {...}  # the current defaults
```
Hence, if the user passes the kwargs every option needs to be set.

anwai98: Fixed. Thanks for spotting!

micro_sam/training/training.py
```diff
@@ -163,28 +170,63 @@ def train_sam(
     else:
         optimizer = torch.optim.Adam(model.parameters(), lr=lr)
 
-    if scheduler is None:
-        scheduler = torch.optim.lr_scheduler.ReduceLROnPlateau(
-            optimizer, mode="min", factor=0.9, patience=3, verbose=True
-        )
+    if scheduler_kwargs is None:
+        scheduler_params = {"mode": "min", "factor": 0.9, "patience": 3, "verbose": True}
```
constantinpape, line 174: This needs to be `scheduler_kwargs` now.

micro_sam/training/training.py
```diff
@@ -163,28 +170,63 @@ def train_sam(
...
+
+    scheduler = scheduler_class(optimizer=optimizer, **scheduler_params)
```
constantinpape, line 176: kwargs, see prev. comment.

constantinpape (approved):

PR #482, opened 2024-03-27
Extend CellPose benchmarking option

anwai98:


anwai98:
The new benchmarking results are ready from here. I'll merge this later today.

PR #483, opened 2024-03-27
Add dict for saving widgets in state

constantinpape:
This adds a dict to the `AnnotatorState` to save widgets, in order to save their values in the program.
We will use this to sync the state of the embedding widget with the parameters passed to the function that start the annotator, and to set model specific default values in the widgets.

I have checked this and it works as expected. I will wait with further implementing the widget update on the current work to make fields in the widgets collabsible by @lufre1. (The changes here don't really interfere with this, but in order to update the values in a widget the layout has to be known and the work by Luca may change that.)


constantinpape:
Closed in favor of #486

PR #486, opened 2024-03-28
Update gui design WIP

constantinpape:


constantinpape:
This PR refactors some of the GUI code to:
- Use QT for the embedding, segment_nd and autosegment widget.
- This enables more customization of the UI, in particular collapsible widgets for advanced settings.
- All three widgets now also make use of a threadworker to wrap the potentially long running calculation, and the progressbar of this calculation is exposed to napari via custom QT signals.

There is some follow-up left for the embedding widget, but we can do this in a separate PR, cc @lufre1.

Issue #332, opened 2024-01-18
Handle non-initialized image embeddings better in the plugins


constantinpape:
Now that we have the embedding widget integrated in the napari plugin, it is possible to click a button that requires the model and image embeddings without having initialized them.  
This fails with a cryptic error message:
```
File ~/Work/my_projects/micro-sam/micro_sam/sam_annotator/_widgets.py:229, in segment_widget(viewer=Viewer(camera=Camera(center=(0.0, 255.5, 255.5),....<locals>._clear_annotations at 0x7f7471efc4a0>}), box_extension=0.1)
    228 image_embeddings = AnnotatorState().image_embeddings
--> 229 if image_embeddings["original_size"] is None:  # tiled prediction
        image_embeddings = None
    230     seg = vutil.prompt_segmentation(
    231         predictor, points, labels, boxes, masks, shape, image_embeddings=image_embeddings,
    232         multiple_box_prompts=True, box_extension=box_extension,
    233     )

TypeError: 'NoneType' object is not subscriptable
```
It would be good to throw a more meaningful error message instead.

We could also indicate visually in the embedding plugin if embeddings are initialized or not.
This could be done via a checkbox `is_initialized` or so  that is not writable from the plugin and just shows if the embeddings are computed or not. (Not sure how to do this exactly and if this is really a good idea.)

constantinpape:
In the long run it would be good to handle this using validation functions in magicgui: https://github.com/pyapp-kit/magicgui/issues/634.
This will probably take a while before being implemented, so we should find an intermediate solution.

constantinpape:
Partially related to this: storing the embedding widgets in the state to have access to it; update the values in the widget to match what is passed in an annotator function, see #483

constantinpape:
This should also encompass handling image embeddings that are wrongly initialized (e.g. for a different image). See also #266

constantinpape:
This is ready to be addressed now that #486 is merged. Relevant locations in the widget are
https://github.com/computational-cell-analytics/micro-sam/blob/dev/micro_sam/sam_annotator/_widgets.py#L617 and https://github.com/computational-cell-analytics/micro-sam/blob/dev/micro_sam/sam_annotator/_widgets.py#L643.

constantinpape:
Implemented in #498

PR #493, opened 2024-04-03
490 finish settings for embedding widget

lufre1:


micro_sam/sam_annotator/_widgets.py
```diff
@@ -123,6 +123,58 @@ def _add_shape_param(self, names, values, min_val, max_val, step=1):
...
+    #     directory_textbox = QtWidgets.QLineEdit()
+    #     directory_textbox.setText(value)
+    #     layout.addWidget(directory_textbox)
+
+    #     directory_button = QtWidgets.QPushButton("Browse")
+    #     directory_button.clicked.connect(lambda: self._get_directory(name, directory_textbox))
+    #     layout.addWidget(directory_button)
+
+    #     return layout
+    def _add_directory_param(self, name, value, title=None, select_file=False):
```
constantinpape, line 140: From what I understand this now works for directories and files (which is good!) I would then rename it to `_add_path_param`.

micro_sam/sam_annotator/_widgets.py
```diff
@@ -123,6 +123,58 @@ def _add_shape_param(self, names, values, min_val, max_val, step=1):
...
+        layout.addWidget(QtWidgets.QLabel(name if title is None else title))
+
+        directory_textbox = QtWidgets.QLineEdit()
+        directory_textbox.setText(value)
+        layout.addWidget(directory_textbox)
+
+        button_text = "Browse Directory" if not select_file else "Browse File"  # Adjust button text
+        directory_button = QtWidgets.QPushButton(button_text)
+        # Call appropriate function based on select_file
+        directory_button.clicked.connect(lambda: getattr(self, "_get_{}_path".format(
```
constantinpape, line 151: It looks like this will fail, because the function is called `_get_directory`. See comment below.

micro_sam/sam_annotator/_widgets.py
```diff
@@ -123,6 +123,58 @@ def _add_shape_param(self, names, values, min_val, max_val, step=1):
...
+            "directory" if not select_file else "file"))(name, directory_textbox))
+        layout.addWidget(directory_button)
+
+        return layout
+
+    def _get_directory(self, name, directory_textbox):
```
constantinpape, line 157: Rename to `_get_directory_path`?

micro_sam/sam_annotator/_widgets.py
```diff
@@ -123,6 +123,58 @@ def _add_shape_param(self, names, values, min_val, max_val, step=1):
...
+    #     return layout
+    def _add_directory_param(self, name, value, title=None, select_file=False):
+        layout = QtWidgets.QHBoxLayout()
+        layout.addWidget(QtWidgets.QLabel(name if title is None else title))
+
+        directory_textbox = QtWidgets.QLineEdit()
+        directory_textbox.setText(value)
+        layout.addWidget(directory_textbox)
+
+        button_text = "Browse Directory" if not select_file else "Browse File"  # Adjust button text
```
constantinpape, line 148: Just a minor comment, but it's generally more readable to write the `"A" if select_file else "B"` instead of `"B" if not select_file else "A"`

micro_sam/util.py
```diff
@@ -463,7 +463,7 @@ def _compute_tiled_features_3d(predictor, input_, tile_shape, halo, f, pbar_init
 
 def _compute_2d(input_, predictor, f, save_path, pbar_init, pbar_update):
     # Check if the embeddings are already cached.
-    if save_path is not None and "input_size" in f.attrs:
+    if save_path is not None and "features" in f.attrs:
```
constantinpape, line 466: `"features"` will never be in `f.attrs`, so this doesn't make sense.

micro_sam/sam_annotator/_widgets.py
lufre1: changed function name to `_get_directory_path` so it's similar to `_get_file_path`

micro_sam/util.py
lufre1: changed it back to "input_size" - there was a confusion on my side regarding the behavior of loading (tiled) embeddings: Loading tiled embeddings with no tiling information (given in the GUI) and vice versa leads to unclear messages. But this belongs to another issue: https://github.com/computational-cell-analytics/micro-sam/issues/332

constantinpape (approved):

PR #496, opened 2024-04-04
Update label transform class name for ASEM (ER) specialist

anwai98:


anwai98:
@constantinpape Changed the class name to avoid facing the pickling issue in future. I'll merge this once the evaluations are completed later today (and ofc after the tests pass)

PR #503, opened 2024-04-08
Minor update to object import in training scheme

anwai98:
@constantinpape Minor fix to an issue I encountered, should be GTG!

### DIFF
--- a/micro_sam/training/training.py
+++ b/micro_sam/training/training.py
@@ -7,7 +7,12 @@ import torch
 import torch_em
 
 from elf.io import open_file
-from qtpy.QtCore import QObject
+
+try:
+    from qtpy.QtCore import QObject
+except ModuleNotFoundError:
+    QObject = Any
+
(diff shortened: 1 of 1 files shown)


constantinpape:
Just fyi, I would rather fix this with updating the dependencies. Will leave it open for now as a reminder.

constantinpape:
The PR with the dependency update for this is merged (both here and on conda-forge), so I am closing this.

anwai98:
I'll go ahead and close this then. Thanks!

anwai98:
@constantinpape For running scripts using python modules, I think we don't need the `qtpy` import mandatorily on top, in cases not desired. I am guessing this should wrap around a try-catch to avoid encountering issues? (I mention this because I faced this recently while loading the model checkpoints)

constantinpape:
@anwai98 yeah, this makes sense. I think we should merge this to keep the training independent of qt. I will merge it once the tests pass.

PR #506, opened 2024-04-10
Update limited livecell data plots for vit tiny

anwai98:


anwai98:
Minor changes, GTG from my side. Will merge this once the test pass. Thanks!

PR #509, opened 2024-04-12
New model updates

constantinpape:
This PR will add the new models, and the (bioimage.io based) code to create the model releases.
It will not fully integrate bioimage.io, since this is still pending the new core updates.


constantinpape:
I will go ahead and merge this, the base functionality for new models is now implemented and works; we just need a few more things for convenient model uploads.

PR #510, opened 2024-04-12
505 update installer scripts

lufre1:
in the future it is possible to add an input masks to the workflow-trigger (e.g. to add version numbers, automatic uploads of new builds) see: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch


constantinpape:
Thanks @lufre1. I cleaned this up a bit to remove the things we ended up not needing, and I redid changes to the env that would clash with other things. I haven't tested it, but think it should work.

I will go ahead and merge it, and we can test it on Monday.

PR #515, opened 2024-04-12
Add model settings for dynamic updates of widget parameters

### LINKED ISSUE #337: Set default values for segmentation widgets based on quantitative experiments
The segmentation widgets (especially for 3d segmentation, tracking and automated segmentation) have parameters for extending boxes, projection modes etc. The default values are currently based on intuition (or copying the defaults from SAM in case of the AMG). We should instead implement quantitative evaluations for the underlying segmentation and use these to set the defaults.

This would also allow us to evaluate how different models perform in these settings.

constantinpape:
Fixes #337 

@anwai98 the parameters from grid searches can be added in the `_model_settings.py`  file. They will then be set in the widgets when the respective model is chosen.


anwai98:
I documented the parameters for auto. segmentation (both AMG and AIS), for all the datasets from both domains here. We can discuss setting the defaults on Monday.

constantinpape:
> I documented the parameters for auto. segmentation (both AMG and AIS), for all the datasets from both domains here. We can discuss setting the defaults on Monday.

Sounds good, let's discuss the details on Monday.

anwai98:
@constantinpape Updated the default parameters for all finetuned models (both lm and em organelles) (btw, the thumbnails look really cool)

PR #518, opened 2024-04-16
Update user study plots

constantinpape:


anwai98:
The user study plots are all completed and good to go from my side. I'll make some minor touch-ups, and merge it later!

PR #521, opened 2024-04-17
Add documentation for LM and EM datasets

anwai98:


anwai98:
The basic addition is GTG for now. If needed, we can add details and hint to the dataloader links at `torch-em`. I'll go ahead and merge this for now.

PR #519, opened 2024-04-16
517 features in embedding widget not working properly

lufre1:
implemented some changes as we discussed for the 2d annotator
@constantinpape

PR #522, opened 2024-04-17
520 add info dialog to image series annotator

lufre1:
@constantinpape


constantinpape:
I will take a look later.

micro_sam/sam_annotator/image_series_annotator.py
```diff
@@ -180,17 +180,23 @@ def next_image(*args):
...
         _save_segmentation(images[next_image_id], next_image_id, segmentation)
 
         # Load the next image.
         next_image_id += 1
         if next_image_id == len(images):
-            print("You have annotated the last image.")
+            msg = "You have annotated the last image."
+            print(msg)
+            # inform the user via dialog
+            # abort = widgets._generate_message("error", msg)
```
constantinpape, line 199: I agree that we should open a dialog here. Let's discuss tomorrow what exactly we do in it.

constantinpape:
I checked out the changes here (and from #519) now, and it looks like everything is working!
I also updated the pbar handling, so that the changes here don't break other code.

We can add the dialog for the last image tomorrow and then this should be good to go.

constantinpape (approved):

PR #524, opened 2024-04-18
487 helper texts for menu options in annotator tool

lufre1:
I implemented tooltips for the qt widgets using the _tooltips.py
The tooltips for widgets created with magic_factory can only be added via docstrings in the functions directly. Sadly, with magic_factory design it is not possible to add tooltips to "call_buttons", but it is possible to add tooltips to every parameter in the function.
Note: magicgui widgets and qt widget are capable of adding tooltips to call_buttons
@constantinpape

### DIFF
--- a/micro_sam/sam_annotator/_tooltips.py
+++ b/micro_sam/sam_annotator/_tooltips.py
@@ -1,0 +1,59 @@
+"""Tooltips for widgets in the annotator.
+"""
+
+tooltips = {
+  "embedding": {
+    "custom_weights": "Select a path for custom weights.",
+    "device": "Select the computational device to use for processing.",
+    "embeddings_save_path": "Path to save the computed image embeddings.",
+    "halo": """Enter values to activate halo. Only x value for quadratic halo and both for custom.
+    Only active when tiling is active.""",
+    "image": "Select an image to segment objects.",
+    "model": "Select a model to use for segmentation tasks.",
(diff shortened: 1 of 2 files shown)


constantinpape:
Thanks @lufre1! 

> Sadly, with magic_factory design it is not possible to add tooltips to "call_buttons", but it is possible to add tooltips to every parameter in the function.

That's ok. I think the tool tips for the call buttons are not so important; the ones for the parameters are more important.

Does it now also work for settings within the collapsible?

lufre1:
yes, it now works within the collapsable sections, but i havent found a solution to add tooltips to the "settings" button which can un-/collapse the section. 
Could you go through the _tooltips.py and add more fitting tooltips for "segmentnd" and "autosegment"?

constantinpape:
> yes, it now works within the collapsable sections 

That's great!

> but i havent found a solution to add tooltips to the "settings" button which can un-/collapse the section.

That's ok, we don't need a tooltip for this.

> Could you go through the _tooltips.py and add more fitting tooltips for "segmentnd" and "autosegment"?

Yes I will do that later (after merging).

> Also, there is a possibility to add tooltips on a more fine-grained level:
> 
>     * adding different tooltips to label and widget
> 
>     * adding tooltips for each option in dropdown menus

Ok. We can discuss next week if we want to extend this.

constantinpape (approved):

PR #528, opened 2024-04-21
Update experiments for hyperparameter search

anwai98:


anwai98:
The experiments are done here, it's GTG from my side. I'll go ahead and merge this once the tests pass.

PR #532, opened 2024-04-23
Update the training UI

constantinpape:


constantinpape:
FYI @anwai98 : this finalizes the implementation of the training UI.

But we should still double check the names in the UI and then add proper tooltips.

PR #535, opened 2024-04-23
Implement preserve committed

### LINKED ISSUE #32: New object segmentations will overwrite older ones
I noticed that if a new object segmentation overlaps all or part of an older object, committing the new segmentation will overwrite the old one.

This is probably not a bug and is the desired behaviour. Making this more clearly understood in the user documentation is probably a useful idea.

Context: I shared some first impressions using micro-sam with Constantin. He asked me to write them down here, so we don't forget what was said.

constantinpape:
Fixes #32


constantinpape:
Something in the MAC CI seems broken. I hope this fixes itself and will merge despite this test failure.

PR #546, opened 2024-04-26
539 update doc

lufre1:
updated docs regarding embedding and some other things. also changed the settings-buttons names in the annotator tool.

@constantinpape


constantinpape (approved):
I will go ahead and merge this once tests pass, so that we have these changes for the release.

PR #547, opened 2024-04-26
Update precomputation of embeddings for 3d volume

anwai98:


anwai98:
This precomputes the image embeddings as expected, I think now it's missing the amg / ais state computation, which is the one you mentioned previously. I'll leave this open for now then!

constantinpape:
Thanks! I will check out the AMG / AIS precomputation later.

constantinpape:
I had another look and we need to slightly redesign this. I will close this and create another PR.

PR #567, opened 2024-04-30
Move suggestions and limitations to FAQ section

anwai98:
@constantinpape 

I will move this from the main documentation part to the FAQs. In addition, I decided on:
- Removing the note on the requirement of precomputing embeddings while using tiling (now we support doing it on-the-fly)
- Remove the known limitation of AMG (as we now support AIS for finetuned models, and we now support full volumetric segmentation)
- I wasn't sure whether to keep this in limitations, posting here in case we decide otherwise:
```
Prompt bounding boxes do not provide the full functionality for tracking yet (they cannot be used for divisions or for starting new tracks). See also [this github issue](https://github.com/computational-cell-analytics/micro-sam/issues/23).
```


constantinpape:
Sounds good! Feel free to merge once the tests pass.

> * I wasn't sure whether to keep this in limitations, posting here in case we decide otherwise:

We can remove it; it's a technical detail and not really a limitation.

PR #571, opened 2024-05-01
Adding doc for using samplers in finetuning

anwai98:


anwai98:
This is a minor addition which helps elaborate one of the FAQs under finetuning. I'll go ahead and merge this once the tests pass. THanks!

PR #573, opened 2024-05-02
changed old logo path to new path

lufre1:
@constantinpape 
Now, all the paths should be changed to the new logo


deployment/windows_menu.json
```diff
@@ -5,7 +5,7 @@
             {
                 "script": "${PREFIX}/Scripts/napari.exe",
 				"name": "micro_sam",
-				"icon": "${PREFIX}/Menu/micro-sam-logo.ico",
+				"icon": "${PREFIX}/Menu/logo.png",
```
constantinpape, line 8: I think you have to convert this to the .ico file format. Otherwise this will not be displayed correctly.
See https://en.wikipedia.org/wiki/ICO_(file_format) .

constantinpape (changes requested):

constantinpape:
And please change the PR to master (we'lll skip the dev for now)

deployment/windows_menu.json
lufre1: i used imagemagick cli tool to convert png to ico with: `convert logo.png -define icon:auto-resize=64,48,32,16 logo.ico`

PR #577, opened 2024-05-02
Minor touch-up to finetuning directory

anwai98:


anwai98:
This is GTG from my side, takes care of maintainence for keeping only the working scripts (and also the scripts supported by the latest `micro_sam` library at the top-level, marking the rest as not relevant (by moving to `outdated` directories).

PR #586, opened 2024-05-05
Add batched inference and evaluation notebook

anwai98:


anwai98:
This PR will add the batched inference and evaluation notebook for all images in a dataset. I'll go and merge it once the tests pass (so that we can have the link at master, I'll continue to work on this meanwhile). Thanks!

Issue #558, opened 2024-04-29
Fix tests on macos-14 (currently the macos-latest github runner)


constantinpape:
The MAC tests fail due to some problem with running the GUI tests. This happens since a couple of days ago, maybe due to some update in the MAC runner?

See the corresponding error in the CI below.
```
Run aganders3/headless-gui@v2
running on darwin
/bin/bash -l -c pytest
INTERNALERROR> Traceback (most recent call last):
INTERNALERROR>   File "/Users/runner/micromamba/envs/sam/lib/python3.11/site-packages/_pytest/main.py", line 281, in wrap_session
INTERNALERROR>     config._do_configure()
INTERNALERROR>   File "/Users/runner/micromamba/envs/sam/lib/python3.11/site-packages/_pytest/config/__init__.py", line 1130, in _do_configure
INTERNALERROR>     self.hook.pytest_configure.call_historic(kwargs=dict(config=self))
INTERNALERROR>   File "/Users/runner/micromamba/envs/sam/lib/python3.11/site-packages/pluggy/_hooks.py", line 535, in call_historic
INTERNALERROR>     res = self._hookexec(self.name, self._hookimpls.copy(), kwargs, False)
INTERNALERROR>           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
INTERNALERROR>   File "/Users/runner/micromamba/envs/sam/lib/python3.11/site-packages/pluggy/_manager.py", line 120, in _hookexec
INTERNALERROR>     return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
INTERNALERROR>            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
INTERNALERROR>   File "/Users/runner/micromamba/envs/sam/lib/python3.11/site-packages/pluggy/_callers.py", line 139, in _multicall
INTERNALERROR>     raise exception.with_traceback(exception.__traceback__)
INTERNALERROR>   File "/Users/runner/micromamba/envs/sam/lib/python3.11/site-packages/pluggy/_callers.py", line 103, in _multicall
INTERNALERROR>     res = hook_impl.function(*args)
INTERNALERROR>           ^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #595, opened 2024-05-08
Fix pyqt error on macos-14 (macos-latest)

GenevieveBuckley:
## Problem
Closes https://github.com/computational-cell-analytics/micro-sam/issues/558

Tests were failing on the macos-latest github runner (which currently runs the macos-14 M1 Mac image). The cause of this appears to be that pyqt was first being installed by conda/micromamba, then pip was uninstalling that and re-installing another pyqt. Something is wrong with the pyqt from pip, and it causes errors on M1 Mac machines.

## Fix
The simplest solution is to remove `napari[all]` from the `requirements-dev.txt` file, so that pip does not try to uninstall the existing version of pyqt. All the github actions tests now finish without errors.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -42,16 +42,6 @@ jobs:
         shell: bash -l {0}
         run: pip install --no-deps -e .
 
-      - name: which python, which pip, which conda
-        run: |
-          which python
-          which pip
-          pip list
-          which conda
-
-      - name: conda list
-        run: conda list
-
       - name: Run tests
         uses: aganders3/headless-gui@v2
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yaml
```diff
@@ -31,7 +31,7 @@ jobs:
         with:
           environment-file: environment_cpu.yaml
 
-      # Setup Qt libraries for GUI testing
+      # Setup Qt libraries for GUI testing on Linux
```
GenevieveBuckley, line 34: This line is unrelated to the specific issue being solved, but I've improved the comment anyway. Mostly because I forgot this, so next time it will be more obvious now that it is only relevant on Linux.

constantinpape (approved):
Thanks a lot @GenevieveBuckley !

Issue #530, opened 2024-04-22
Put on napari hub


constantinpape:
https://www.napari-hub.org/

GenevieveBuckley:
Discussed previously here: https://forum.image.sc/t/publish-napari-plugin-with-conda-dependencies/65653

Summarizing:
* We need to release micro-sam on PyPI. Then napari-hub will automatically scrape PyPI and will find micro-sam, because we already have the `Framework :: napari` classifier.
* Because micro-sam requires some dependencies only available on conda (not pypi), we will...
    * Not worry about the fact it will not install everything correctly if users try `pip install micro-sam` (this is not great, but there is no good alternative if we want micro-sam discoverable on the napari-hub)
    * Add a line to the documentation (maybe [around here](https://github.com/computational-cell-analytics/micro-sam/blob/6a2a3fe87751650208000a443860fcf262186eed/doc/installation.md?plain=1#L8)) to tell users please do not use pip to install micro-sam. Hopefully anyone with installation problems will see this page and then try again with conda or from source.

GenevieveBuckley:
FYI: it seems that the "Image Modality" labels can only be controlled by adding a `.napari-hub/config.yml` file. This is because this key is not (yet?) supported by the napari manifest `napari.yaml` file.

We could add that too, I guess, but it does feel messy having so many config files scattered around. 

I only found one plugin on the hub using the Image Modality labels (probably because it is not well-documented how to set them up). It was the [napari-biomag-annotator plugin, see the example `.napari-hub/config.yml` file here](https://github.com/biomag-lab/napari-biomag-annotator/blob/main/.napari/config.yml). 

Note: That `.napari-hub/config.yml` file does list a lot more labels than I see actually being used by [the napari-hub page](https://www.napari-hub.org/plugins/napari-biomag-annotator?imageModality=Confocal+microscopy), so I'm not sure if some of these are effectively not being used. It's possible only specific values are accepted, but I can't find good documentation about it, so this one example is the best we have to go on.

constantinpape:
Thanks for looking into this @GenevieveBuckley. Let's go ahead and put a "dummy" micro_sam package on pip then so that it can be installed via napari.

Would you have time to look into this? (There's no hurry!)

GenevieveBuckley:
> Thanks for looking into this @GenevieveBuckley. Let's go ahead and put a "dummy" micro_sam package on pip then so that it can be installed via napari.
> 
[...]

PR #599, opened 2024-05-09
Update napari.yaml, include napari-hub category labels

GenevieveBuckley:
Adds "Segmentation" and "Annotation" category labels for our future napari-hub plugin page.

See https://napari.org/stable/plugins/manifest.html for valid categories. 
Currently allowed categories are: Acquisition, Annotation, Dataset, Image Processing, IO, Measurement, Segmentation, Simulation, Themes, Transformations, Utilities, Visualization

Related to (but does not close issue) https://github.com/computational-cell-analytics/micro-sam/issues/530

### DIFF
--- a/micro_sam/napari.yaml
+++ b/micro_sam/napari.yaml
@@ -1,8 +1,9 @@
 name: micro-sam
 display_name: SegmentAnything for Microscopy
+# see https://napari.org/stable/plugins/manifest.html for valid categories
+categories: ["Segmentation", "Annotation"]
 contributions:
   commands:
-
     # Commands for sample data.
     # We don't provide the sample data for image series via napari for now,
     # because the image series annotator cannot be called for image layers in napari (yet),


GenevieveBuckley:
FYI: it seems that the "Image Modality" labels can only be controlled by adding a `.napari-hub/config.yml` file. This is because this key is not (yet?) supported by the napari manifest `napari.yaml` file.

We could add that too, I guess, but it does feel messy having so many config files scattered around. 

I only found one plugin on the hub using the Image Modality labels (probably because it is not well-documented how to set them up). It was the [napari-biomag-annotator plugin, see the example `.napari-hub/config.yml` file here](https://github.com/biomag-lab/napari-biomag-annotator/blob/main/.napari/config.yml). 

Note: That `.napari-hub/config.yml` file does list a lot more labels than I see actually being used by [the napari-hub page](https://www.napari-hub.org/plugins/napari-biomag-annotator?imageModality=Confocal+microscopy), so I'm not sure if some of these are effectively not being used. It's possible only specific values are accepted, but I can't find good documentation about it, so this one example is the best we have to go on.

constantinpape:
Thanks @GenevieveBuckley.

Regarding the Imaging Modalities: I would not add this for now; we can still add it later on.

PR #611, opened 2024-05-21
Add LoRA Implementation

anwai98:


micro_sam/training/util.py
```diff
@@ -42,6 +42,7 @@ def get_trainable_sam_model(
     checkpoint_path: Optional[Union[str, os.PathLike]] = None,
     freeze: Optional[List[str]] = None,
     return_state: bool = False,
+    get_lora: bool = False,
```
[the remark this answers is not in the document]
anwai98: Yes, it does make sense to expose this a bit. For simplicity, I have kept botht he function signatures `use_lora` and `rank` in the signatures. A modular choice would be to expose just the `rank` and decide on LoRA or full finetuning, but I would leave it to a simpler structure now for the experiments. Thanks for the mention.

micro_sam/training/peft_sam.py
```diff
@@ -0,0 +1,155 @@
...
+import torch
+import torch.nn as nn
+
+from segment_anything.modeling import Sam
+
+
+class PEFTBase(nn.Module):
+    """PEFTBase is an interface to implement specific PEFT-based methods.
+    """
+    def __call__(
```
constantinpape, line 13: It would make more sense to have an `__init__` here, that defines a common signature for all PEFT adapter child classes.

micro_sam/training/peft_sam.py
```diff
@@ -0,0 +1,155 @@
...
+        """
+        raise NotImplementedError(
+            "PEFTBase is just a class template. Use a child class that implements the specific PEFT method instead."
+        )
+
+
+class LoRASurgery(PEFTBase):
+    """Operates on the attention layers for performing low-rank adaptation.
+    """
+    def __call__(
```
constantinpape, line 36: Use `__init__` instead, set all the relevant weights as members (`w_a_linear` etc.).  If needed implement a `get_qkv` function that you can call in `PEFT_Sam` to do the block update.

micro_sam/training/peft_sam.py
```diff
@@ -0,0 +1,155 @@
...
+        w_b_linear_v = nn.Linear(rank, dim, bias=False)
+
+        block.attn.qkv = _LoRA_Sam(
+            w_qkv_linear, w_a_linear_q, w_b_linear_q, w_a_linear_v, w_b_linear_v,
+        )
+
+        return block, [w_a_linear_q, w_b_linear_q, w_a_linear_v, w_b_linear_v]
+
+
+class _LoRA_Sam(nn.Module):
```
constantinpape, line 56: Absorb this into the `LoRASurgery` class.

micro_sam/training/peft_sam.py
```diff
@@ -0,0 +1,155 @@
...
+        else:   # Applies PEFT to the image encoder by default
+            self.peft_layers = list(
+                range(len(model.image_encoder.blocks))
+            )
+
+        self.peft_module = peft_module
+
+        # create for storage, then we can init them or load weights
+        self.w_As = []  # These are linear layers
+        self.w_Bs = []
```
constantinpape, line 128: Create a list `self.peft_blocks = []` instead.

micro_sam/training/peft_sam.py
```diff
@@ -0,0 +1,155 @@
...
+        for t_layer_i, blk in enumerate(model.image_encoder.blocks):
+            # If we only want specific layers for PEFT instead of all
+            if t_layer_i not in self.peft_layers:
+                continue
+
+            blk, linear_layers = self.peft_module(rank=rank, block=blk)
+            w_a_linear_q, w_b_linear_q, w_a_linear_v, w_b_linear_v = linear_layers
+
+            self.w_As.extend([w_a_linear_q, w_a_linear_v])
+            self.w_Bs.extend([w_b_linear_q, w_b_linear_v])
```
constantinpape, line 143: Do this instead:
```python
peft_block_i = self._peft_module(...)
blk.attn.qkv  = peft_block.get_qkv()
self.peft_blocks.append(peft_block_i)
```

constantinpape (commented):
I think it would make more sense to design the PEFT adapters as proper pytorch modules, see the comments for details.

micro_sam/training/peft_sam.py
constantinpape: We don't need to expose the weights `w_a_linear` etc. at all then, they live in the `peft_block`.

constantinpape: On second thought: I think it's even better to do `blk.attn.qkv  = peft_block.get_qkv()` in the `__init__` of the `PEFTBlock`, by passing `blk` to it. This way the details of the adaptation are all abstracted away.

anwai98:
@constantinpape Thanks for all the suggestions. I thought about all your feedback ,and imcorporated them now in a much more flexible design. Let me know if I missed something or you spot some more room for improvement. Thanks!

micro_sam/training/peft_sam.py
```diff
@@ -0,0 +1,99 @@
...
+        assert rank > 0
+
+        if attention_layers_to_update:
+            self.peft_layers = attention_layers_to_update
+        else:   # Applies PEFT to the image encoder by default
+            self.peft_layers = list(
+                range(len(model.image_encoder.blocks))
+            )
+
+        self.peft_module = peft_module
```
constantinpape, line 83: I think we need a list here that keeps the instantiated peft modules:
```
self.peft_blocks = []
```

micro_sam/training/peft_sam.py
```diff
@@ -0,0 +1,99 @@
...
+        # let's freeze all the pretrained image encoder layers first
+        for param in model.image_encoder.parameters():
+            param.requires_grad = False
+
+        for t_layer_i, blk in enumerate(model.image_encoder.blocks):
+            # If we only want specific layers for PEFT instead of all
+            if t_layer_i not in self.peft_layers:
+                continue
+
+            blk.attn.qkv = self.peft_module(rank=rank, block=blk)
```
constantinpape, line 94: I would change this in two ways: Add the instantiated module to the list and move the qkv update into the peft module (see also the comment above)
```python
peft_block = self.peft_module(rank=rank, block=block)
self.peft_blocks.append(peft_block)
```

micro_sam/training/peft_sam.py
```diff
@@ -0,0 +1,99 @@
...
+        super().__init__()
+        self.qkv = block.attn.qkv
+        self.dim = self.qkv.in_features
+
+        self.w_a_linear_q = nn.Linear(self.dim, rank, bias=False)
+        self.w_b_linear_q = nn.Linear(rank, self.dim, bias=False)
+        self.w_a_linear_v = nn.Linear(self.dim, rank, bias=False)
+        self.w_b_linear_v = nn.Linear(rank, self.dim, bias=False)
+
+        self.reset_parameters()
```
constantinpape, line 36: Add the update to the block here. I.e.
```
blk.attn.qkv = self
```
This way the updates to the encoder are fully encapsulated here, and we have more flexibility and could e.g. later add a PEFT block that also updates the linear layers.

constantinpape (commented):
This is going the right direction! A few more changes to make it more modular.

In addition, it would also be good to have a test for this, that instantiates the PEFT SAM, applies a forward pass and checks the gradients afterwards.

micro_sam/training/peft_sam.py
anwai98: Oh yes, this would have it completely independent. Thanks for the suggestion!

anwai98:
@constantinpape Thanks for the feedback. I've updated the `PEFT_Sam` block as mentioned, so that it is now independent of any assumptions to update parts of the backbone, and instead the `self.peft_module` will be responsible of doing this from now on (i.e. the surgery class provided by the user to handle everything for the attention blocks).

I also added a simple test for the `PEFT_Sam` class.

Let me know how things look now. Thanks!

anwai98:
To add, looks like tests at `test_bioimageio/test_model_export.py` are failing for some reason.

micro_sam/training/peft_sam.py
```diff
@@ -0,0 +1,103 @@
...
+        for param in model.image_encoder.parameters():
+            param.requires_grad = False
+
+        for t_layer_i, blk in enumerate(model.image_encoder.blocks):
+            # If we only want specific layers with PEFT instead of all
+            if t_layer_i not in self.peft_layers:
+                continue
+
+            peft_block = self.peft_module(rank=rank, block=blk)
+            self.peft_blocks.append(peft_block)
```
constantinpape, line 98: Can you do a
```python
self.peft_blocks = nn.ModuleList(self.peft_blocks)
```
after the loop? To make sure the gradients are propagated correctly.

test/test_peft_training.py
```diff
@@ -0,0 +1,49 @@
...
+        # Wrap the Segment Anything model with PEFT methods.
+        peft_sam_model = PEFT_Sam(model=sam_model, rank=4)
+
+        # Get the model outputs
+        outputs = peft_sam_model(batched_input=batched_inputs, multimask_output=False)
+
+        # Check the expected shape of the outputs
+        mask_shapes = [output["masks"].shape[-2:] for output in outputs]
+        for shape in mask_shapes:
+            self.assertEqual(shape, input_shape[1:])
```
constantinpape, line 45: Ideally we would check here that gradients are propagated when through the model. But I can also do this in a follow up PR.

constantinpape (commented):
This looks good now, only one minor thing.

test/test_peft_training.py
anwai98: Okay, I've left this in favor of the follow-up PR.

anwai98:
Hi @constantinpape, this PR is GTG from my side now. Thanks!

constantinpape (approved):

PR #612, opened 2024-05-21
Add MedSAM Reimplementation

anwai98:


constantinpape (commented):
This is duplicating a lot of functionality; making use of the flexible design we have already implemented it should be possible to do this in < 20 lines.

micro_sam/training/medsam_trainer.py
```diff
@@ -0,0 +1,164 @@
...
+import time
+
+import torch
+from torchvision.utils import make_grid
+
+import torch_em
+from torch_em.trainer.logger_base import TorchEmLogger
+
+
+class MedSAMTrainer(torch_em.trainer.DefaultTrainer):
```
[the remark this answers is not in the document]
anwai98: I've updated the trainer to a flexible design. which now is more or less no code changes at all. Thanks for the feedback!

micro_sam/training/sam_trainer.py
```diff
@@ -216,12 +222,13 @@ def _compute_iterative_loss(self, batched_inputs, y_one_hot, num_subiter, multim
             iou_regression_loss += net_iou_regression_loss
             mean_model_iou += net_mean_model_iou
 
-            # Determine the next prompts based on current predictions.
-            with torch.no_grad():
-                # Get the mask and logit predictions corresponding to the predicted object
-                # (per actual object) with the best IOU.
-                masks, logits = self._get_best_masks(batched_outputs, batched_iou_predictions)
-                batched_inputs = self._update_prompts(batched_inputs, y_one_hot, masks, logits)
+            if num_subiter > 1:
```
constantinpape, line 225: We could be more efficient here, and check if we still have a next subiteration to go for which we need the masks and updated prompts. Then normal training would also profit from this.
So something like 
```python
if i < (num_subiter - 1):
...
```
(I think that's correct, but please double check that it's not off by one!)

micro_sam/training/medsam_trainer.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class MedSAMTrainer(SamTrainer):
+    """Trainer class for replicating the trainer of MedSAM (https://arxiv.org/abs/2304.12306)
+    """
+    def __init__(
+        self,
+        **kwargs
+    ):
+        super().__init__(**kwargs)
```
constantinpape, line 11: We should set `n_sub_iterations` to 1 here. It doesn't make sense to require it as a kwarg.

constantinpape (commented):
Looks good overall, just a few minor comments.

micro_sam/training/sam_trainer.py
anwai98: Ahha yes, makes sense. Thanks for the mention!

micro_sam/training/medsam_trainer.py
anwai98: Done. Fixed the `n_sub_iteration` and `mask_prob` values in the trainer. Thanks for the mention!

anwai98:
Thanks for the feedback! This is GTG from my side. Let me know if you spot something else.

micro_sam/training/sam_trainer.py
```diff
@@ -216,12 +222,14 @@ def _compute_iterative_loss(self, batched_inputs, y_one_hot, num_subiter, multim
             iou_regression_loss += net_iou_regression_loss
             mean_model_iou += net_mean_model_iou
 
-            # Determine the next prompts based on current predictions.
-            with torch.no_grad():
-                # Get the mask and logit predictions corresponding to the predicted object
-                # (per actual object) with the best IOU.
-                masks, logits = self._get_best_masks(batched_outputs, batched_iou_predictions)
-                batched_inputs = self._update_prompts(batched_inputs, y_one_hot, masks, logits)
+            if num_subiter > 1:  # Update prompts for more than one sub-iterations
```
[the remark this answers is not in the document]
anwai98: Oh yeah true, overlooked it. Sorry! Thanks for the mention!

constantinpape (approved):

PR #625, opened 2024-05-31
Scripts for finetuning for chromosome segmentation in EM

anwai98:
Hi @constantinpape,

This is a finetuning script for Gunar, for finetuning micro-sam (`vit_b_em_organelles`) on their own data. I made a random split out of one unified dataset to avoid making sure of explicit segmented volumes.

Looks good to me. Let me know if you spot something, I'll then forward them this PR to take a look at this finetuning script.

(PS. It's not tested, but in theory just changing the global variable to declare the path to the parent directory where images and labels are stored in a certain way (I'll communicate this to them) should work out-of-the-box)

Thanks!


constantinpape:
Looks good from my end.

anwai98:
I'll drop this PR as we have a nice workflow for Gunar working already. In case they want to scale up now, I'll ask them to switch to CLI for other tasks. I'll close this for now!

PR #631, opened 2024-06-05
Add SimpleSamTrainer

anwai98:
This PR refactors `MedSAMTrainer` to `SimpleSamTrainer`, and makes using other methods much more flexible!


micro_sam/training/simple_sam_trainer.py
```diff
@@ -0,0 +1,51 @@
...
+class SimpleSamTrainer(SamTrainer):
+    """Trainer class for creating a simple SAM trainer for limited prompt-based segmentation.
+    """
+    def __init__(
+        self,
+        use_points: bool = False,
+        use_box: bool = False,
+        **kwargs
+    ):
+        super().__init__(**kwargs)
```
constantinpape, line 15: Please recheck what we had in the MedSAM trainer, where the relevant super kwargs `n_sub_iter=1, mask_prob=0` where set already.

micro_sam/training/simple_sam_trainer.py
```diff
@@ -0,0 +1,51 @@
...
+from . import SamTrainer
+
+
+class SimpleSamTrainer(SamTrainer):
+    """Trainer class for creating a simple SAM trainer for limited prompt-based segmentation.
+    """
+    def __init__(
+        self,
+        use_points: bool = False,
+        use_box: bool = False,
```
constantinpape, line 12: I would rather set both to `True`.

micro_sam/training/simple_sam_trainer.py
```diff
@@ -0,0 +1,51 @@
...
+        super().__init__(**kwargs)
+        self.use_points = use_points
+        self.use_box = use_box
+
+        if not self.use_points and not self.use_box:  # if user doesn't specify, we randomly choose box / point prompts
+            self.random_prompt_choice = True
+        else:
+            self.random_prompt_choice = False
+
+        assert (self.use_points + self.use_box) < 2, "Please choose either of the prompt-based segmentation."
```
constantinpape, line 24: Having both is totally fine. In that case we should randomly either select a point or mask.

constantinpape: However, we should check for `self.use_points + self.use_box != 0`

micro_sam/training/simple_sam_trainer.py
```diff
@@ -0,0 +1,51 @@
+import random
+
+from . import SamTrainer
+
```
constantinpape, line 4: Let's also add
```python
MedSamTrainer(SimpleSamTrainer):
  def __init__(self, **kwargs):
    super(use_points=False, use_box=True, **kwargs)
```

micro_sam/training/simple_sam_trainer.py
anwai98: Ahha okay, so I'll just reverse the current logic. Your suggestions makes more sense actually. Thanks!

micro_sam/training/simple_sam_trainer.py
anwai98: Ahha yes, makes sense to set these defaults. I'll do this for the new `MedSAMTrainer` as suggested below, by inheriting `SimpleSamTrainer` and setting the defaults for MedSAM.

micro_sam/training/simple_sam_trainer.py
```diff
@@ -0,0 +1,66 @@
...
+        )
+        self.use_points = use_points
+        self.use_box = use_box
+
+        if self.use_points and self.use_box:
+            self.random_prompt_choice = True
+        else:
+            self.random_prompt_choice = False
+
+        assert (self.use_points + self.use_box) != 0, "Please choose atleast one of the prompt-based method."
```
constantinpape, line 28:
```suggestion
        assert (self.use_points + self.use_box) != 0, "Please choose at least one of the prompt-based method."
```

constantinpape (approved):

PR #637, opened 2024-06-17
Add SemanticSamTrainer

anwai98:
@constantinpape Here is the trainer for semantic segmentation using SAM. Let me know if this aligns with what we discussed.

### DIFF
--- a/micro_sam/training/semantic_sam_trainer.py
+++ b/micro_sam/training/semantic_sam_trainer.py
@@ -1,6 +1,7 @@
 import time
 
 import torch
+from torch.nn import CrossEntropyLoss
 
 from torch_em.trainer import DefaultTrainer
 
@@ -17,18 +18,25 @@ class SemanticSamTrainer(DefaultTrainer):
         super().__init__(**kwargs)
         self.convert_inputs = convert_inputs
         self.num_classes = num_classes
+        self.compute_ce_loss = CrossEntropyLoss()
         self._kwargs = kwargs
 
-    def _compute_loss(self, y, masks):
+    def _compute_loss(self, y, gt_logits, masks, mask_logits):
         # Compute dice loss for the predictions
         dice_loss = self.loss(masks, y.to(self.device, non_blocking=True))
-        return dice_loss
+
+        # Compute cross entropy loss for the logits
+        ce_loss = self.compute_ce_loss(mask_logits, gt_logits.to(self.device, non_blocking=True))
+
+        net_loss = dice_loss + ce_loss
+        return net_loss
 
     def _get_model_outputs(self, batched_inputs):
         image_embeddings, batched_inputs = self.model.image_embeddings_oft(batched_inputs)
         batched_outputs = self.model(batched_inputs, image_embeddings, multimask_output=(self.num_classes > 1))
         masks = torch.stack([output["masks"].squeeze(0) for output in batched_outputs])
-        return masks
+        mask_logits = torch.stack([output["low_res_masks"].squeeze(0) for output in batched_outputs])
+        return masks, mask_logits
 
     def _train_epoch_impl(self, progress, forward_context, backprop):
         self.model.train()
@@ -37,17 +45,17 @@ class SemanticSamTrainer(DefaultTrainer):
         for x, y in self.train_loader:
             self.optimizer.zero_grad()
 
-            batched_inputs = self.convert_inputs(x, y)
+            batched_inputs, gt_logits = self.convert_inputs(x, y)
 
             with forward_context():
-                masks = self._get_model_outputs(batched_inputs)
-                loss = self._compute_loss(y, masks)
+                masks, mask_logits = self._get_model_outputs(batched_inputs)
+                net_loss = self._compute_loss(y, gt_logits, masks, mask_logits)
 
-            backprop(loss)
+            backprop(net_loss)
 
             if self.logger is not None:
                 lr = [pm["lr"] for pm in self.optimizer.param_groups][0]
-                self.logger.log_train(self._iteration, loss, lr, x, y, masks, log_gradients=True)
+                self.logger.log_train(self._iteration, net_loss, lr, x, y, masks, log_gradients=True)
 
             self._iteration += 1
             if self._iteration >= self.max_iteration:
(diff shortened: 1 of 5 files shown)


constantinpape (commented):
Looks good to me, I couldn't spot any issues.

anwai98:
Thanks @constantinpape. I tested this in a 2d dataset, looks like it's doing the job as expected. This is GTG from my side now (only pending a few minor discussion in the evaluation PR)

micro_sam/training/semantic_sam_trainer.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def _compute_loss(self, y, masks):
+        # Compute dice loss for the predictions
+        dice_loss = self.loss(masks, y.to(self.device, non_blocking=True))
+        return dice_loss
+
+    def _get_model_outputs(self, batched_inputs):
+        image_embeddings, batched_inputs = self.model.image_embeddings_oft(batched_inputs)
+        batched_outputs = self.model(batched_inputs, image_embeddings)
+        masks = torch.stack([output["masks"].squeeze(0) for output in batched_outputs])
```
constantinpape, line 28: I am rethinking which masks to use here.

constantinpape (commented):
We should revisit this with multi-class segmentation in mind, and potentially prefer the low-res masks.

anwai98:
Hi @constantinpape,

Looks like the mutli-class semantic segmentation works as expected now (atleast from the first looks of the Tensorboard logs). I am not a big fan of the workarounds I had to apply to make this work, but maybe we find a better way to make things work in a much more modular setup. Let me know if you spot something. We can discuss together tomorrow the details anyways.

micro_sam/training/semantic_sam_trainer.py
```diff
@@ -0,0 +1,92 @@
...
+    def __init__(
+        self,
+        convert_inputs,
+        num_classes: int = 1,
+        **kwargs
+    ):
+        super().__init__(**kwargs)
+        self.convert_inputs = convert_inputs
+        self.num_classes = num_classes
+        self.compute_ce_loss = CrossEntropyLoss()
```
constantinpape, line 21: Shouldn't we use BCE if `num_classes = 1` ?

micro_sam/training/util.py
```diff
@@ -210,6 +224,29 @@ def __call__(self, x, y, n_pos, n_neg, get_boxes=False, n_samples=None):
...
+        """
+        batched_inputs, gt_logits = [], []
+        for image, gt in zip(x, y):
+            batched_input = {"image": image, "original_size": image.shape[1:]}
+            batched_inputs.append(batched_input)
+
+            # downsize the labels
+            gt_shape = (gt.shape[0], 256, 256)
+            gt_logits.append(
+                resize(image=gt, output_shape=gt_shape, preserve_range=True, order=0, anti_aliasing=False)
```
constantinpape, line 241: This will change the dtype. I am not sure if this is a problem though. It can be prevented by adding `.asyupe(gt.dtype)` at the end.

micro_sam/training/util.py
```diff
@@ -210,6 +224,29 @@ def __call__(self, x, y, n_pos, n_neg, get_boxes=False, n_samples=None):
         return batched_inputs, batched_sampled_cell_ids_list
 
 
+class ConvertToSemanticSamInputs:
+    """
+    """
+    def __call__(self, x, y):
+        """Convert the outputs of dataloader to the batched format of inputs expected by SAM.
+        """
+        batched_inputs, gt_logits = [], []
```
constantinpape, line 233: Calling it `gt_logits` is a bit confusing, since these are not logits,  but just downsampled.

micro_sam/util.py
```diff
@@ -350,15 +353,29 @@ def get_sam_model(
         )
 
     state, model_state = _load_checkpoint(checkpoint_path)
-    sam = sam_model_registry[abbreviated_model_type]()
 
+    # Whether to update parameters necessary to initialize the model
+    if bool(model_kwargs):  # Checks whether model_kwargs have been provided or not
```
constantinpape, line 358: You don't need to cast to `bool` here.

micro_sam/util.py
```diff
@@ -379,6 +396,38 @@ def get_sam_model(
...
+    for k, v in model_state.items():
+        if reference_state[k].size() == v.size():
+            new_state_dict[k] = v
+        else:
+            mismatched_layers.append(k)
+
+    reference_state.update(new_state_dict)
+
+    if len(mismatched_layers) > 0:
+        print(f"The layers with size mismatch: {mismatched_layers}")
```
constantinpape, line 418: I would raise a warning instead.

constantinpape:
Feel free to merge this!

micro_sam/training/util.py
```diff
@@ -210,6 +224,29 @@ def __call__(self, x, y, n_pos, n_neg, get_boxes=False, n_samples=None):
...
+    def __call__(self, x, y):
+        """Convert the outputs of dataloader to the batched format of inputs expected by SAM.
+        """
+        batched_inputs, downsized_gt = [], []
+        for image, gt in zip(x, y):
+            batched_input = {"image": image, "original_size": image.shape[1:]}
+            batched_inputs.append(batched_input)
+
+            # downsize the labels
+            gt_shape = (gt.shape[0], 256, 256)
```
constantinpape, line 239: Do we still need `downsized_gt`? I think we don't anymore given the updates to the semantic training.
Can we remove it?

constantinpape (commented):
Looks good, but let's remove the GT downsampling. (I think we don't need it anyore)

micro_sam/training/util.py
anwai98: Ah yes, I think this should now work without the need of this. Hence, we can have the combined objective of the CrossEntropy (or BCE) and Dice over the `batch_masks` vs the original ground truth shape. I'll take care of this right away. Thanks for spotting.

anwai98:
@constantinpape,

I've removed the downscaling of masks. Should be GTG now. Thanks!

constantinpape (approved):

PR #638, opened 2024-06-20
Lora implementation in finetuning and evaluation

caroteu:


finetuning/specialists/lora/train_covid_if.py
caroteu, line 1: @anwai98 This is the script I used for finetuning on covid_if data. I refer to this one in the data I sent as 'my impl'.

finetuning/specialists/lora/train_covid_if.py
```diff
@@ -0,0 +1,216 @@
...
+    # let's get the parameters for SAM and the decoder from UNETR
+    joint_model_params = model.parameters()
+
+    joint_model_params = [params for params in joint_model_params]  # sam parameters
+    for name, params in unetr.named_parameters():  # unetr's decoder parameters
+        if not name.startswith("encoder"):
+            joint_model_params.append(params)
+
+    optimizer = torch.optim.Adam(joint_model_params, lr=1e-5)
+    scheduler = torch.optim.lr_scheduler.ReduceLROnPlateau(optimizer, mode="min", factor=0.9, patience=10)
```
caroteu, line 134: @anwai98 The patience for the learning rate scheduler was set to 10 here but to 3 in the 'Resource Efficient Impl' - Could that be the reason for the performance difference ?

anwai98: There are a few more changes. We had a lower learning rate and a different optimizer as well (https://github.com/caroteu/micro-sam/blob/dcd6ecdc5ef600e07670db27ccfba54e81f156f7/finetuning/specialists/resource-efficient/covid_if_finetuning.py#L135-L136)

And probably 100 epochs in the other experiments might translate to a bit more than 10k iterations here.

anwai98: Potentially that could lead to a bit of a performance difference (unless it's a very severe drop in performance, that would be a different discussion)

caroteu: The optimizer is actually the same - I ran this here with Adam and 1e-5 too to make it consistent with the other workflow. The results I send you both have adam and lr=1e-5

anwai98:
Hi @caroteu,

I'll close this PR in favor of arranging the experiments in the discussed repo from now on. Feel free to reopen if you think there are some crucial changes to be added via this PR.

PR #643, opened 2024-06-27
Add SemanticSam3dLogger

anwai98:


micro_sam/sam_3d_wrapper.py
```diff
@@ -6,16 +6,32 @@
...
+
+
+def get_3d_sam_model(
+    device,
+    n_classes,
+    image_size,
+    model_type="vit_b",
+    checkpoint_path=None,
+):
+    _, sam = get_sam_model(
```
constantinpape, line 19: Can't we just pass the checkpoint here? Then we don't need to duplicate the code below.

micro_sam/sam_3d_wrapper.py
```diff
@@ -6,16 +6,32 @@
...
+        return_sam=True,
+        flexible_load_checkpoint=True,
+        num_multimask_outputs=n_classes,
+        image_size=image_size,
     )
+
     sam_3d = Sam3DWrapper(sam)
     sam_3d.to(device)
+
+    if checkpoint_path is not None:
```
constantinpape, line 31: Get rid of this and just load the checkpoint in the beginning?! See comment above.

micro_sam/training/semantic_sam_trainer.py
```diff
@@ -56,11 +85,13 @@ def _train_epoch_impl(self, progress, forward_context, backprop):
 
             backprop(net_loss)
 
+            self._iteration += 1
+
             if self.logger is not None:
                 lr = [pm["lr"] for pm in self.optimizer.param_groups][0]
-                self.logger.log_train(self._iteration, net_loss, lr, x, y, masks, log_gradients=True)
+                predictions = torch.softmax(masks, dim=1)
```
constantinpape, line 92: I don't think it's  necessary to do the softmax here; we would then do it twice (see comment below)

micro_sam/training/semantic_sam_trainer.py
```diff
@@ -107,3 +140,39 @@ def _get_model_outputs(self, batched_inputs):
...
+
+        selection_image = np.s_[0] if x.ndim == 4 else np.s_[0, x.shape[2] // 2, :]
+        selection = np.s_[0] if x.ndim == 4 else np.s_[0, :, x.shape[2] // 2]
+
+        image = normalize_im(x[selection_image].cpu())
+        self.tb.add_image(tag=f"{name}/input",
+                          img_tensor=image,
+                          global_step=step)
+
+        prediction = torch.softmax(prediction, dim=1)
```
constantinpape, line 156: We do the softmax here, so no need to do it in the main training body.

micro_sam/training/semantic_sam_trainer.py
```diff
@@ -107,3 +140,39 @@ def _get_model_outputs(self, batched_inputs):
         # masks = torch.stack([output["masks"].squeeze(0) for output in batched_outputs])
         masks = batched_outputs["masks"]
         return masks
+
+
+class SemanticSamLogger3D(TensorboardLogger):
```
constantinpape, line 145: This logger should now work for 2d and 3d. Should we rename to `SemanticSamLogger` and then use it for the `SemanticSamTrainer` by default?

constantinpape (commented):
I went through this again and found a few more small points. Once this is addressed we can merge this.

micro_sam/sam_3d_wrapper.py
anwai98: Ah, that won't work as we need to pass the weights to the wrapper model (to initialize the adapter blocks)

micro_sam/training/semantic_sam_trainer.py
anwai98: Yep, makes sense. Updated it in a latest commit.

micro_sam/sam_3d_wrapper.py
anwai98: As mentioned above, this won't work unless we pull in the sam3d wrapper inside `get_sam_model`.

micro_sam/training/semantic_sam_trainer.py
```diff
@@ -13,31 +39,34 @@ class SemanticSamTrainer(DefaultTrainer):
...
+        num_classes: int,
         **kwargs
     ):
-        loss = DiceLoss()
-        metric = DiceLoss()
+        assert num_classes > 1
+
+        loss = CustomDiceLoss(num_classes=num_classes)
+        metric = CustomDiceLoss(num_classes=num_classes)
         super().__init__(loss=loss, metric=metric, **kwargs)
```
constantinpape, line 49: Let's pass the `SemanticSamLogger` for the logger class here.

micro_sam/sam_3d_wrapper.py
constantinpape: Hmm, I don't fully understand. Let's discuss tomorrow :) .

micro_sam/training/semantic_sam_trainer.py
anwai98: Done. Thanks!

micro_sam/sam_3d_wrapper.py
anwai98: Okay now I see already where your suspicion comes from.

TLDR: I think we might have to merge the `get_3d_sam_model` into `get_sam_model` for the best possible design (which leads to flexibly loading SAM checkpoints for finetuning, and the 3d-SAM checkpoints for downstream semantic inference).

I have a plan on this. I'll take care of this first thing in the morning. Thanks for spotting.

constantinpape: I think the current design should work at least for intialization. We can revisit this later to discuss how we do this for actually loading the trained 3d models.

constantinpape (approved):

micro_sam/sam_3d_wrapper.py
anwai98: Okie, I'll leave it to here for now then. Thanks!

PR #646, opened 2024-06-28
Minor fix to trainable sam model functionality

anwai98:
@constantinpape,

I might have created this redundant change while trying to make the inference work. This is unnecessary, as we have LoRA updates happening below in the code, which takes care of the two stage idea while loading trainable sam. This is GTG from my side. Thanks!

### DIFF
--- a/micro_sam/training/util.py
+++ b/micro_sam/training/util.py
@@ -75,6 +75,8 @@ def get_trainable_sam_model(
         checkpoint_path=checkpoint_path,
         return_sam=True,
         return_state=True,
+        use_lora=use_lora,
+        rank=rank,
         flexible_load_checkpoint=flexible_load_checkpoint,
         **model_kwargs
     )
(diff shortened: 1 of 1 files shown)


micro_sam/training/util.py
```diff
@@ -75,8 +75,6 @@ def get_trainable_sam_model(
         checkpoint_path=checkpoint_path,
         return_sam=True,
         return_state=True,
-        use_lora=use_lora,
-        rank=rank,
```
constantinpape, line 79: I would rather keep these here and remove lines 98-102. Then we don't duplicate the functionality of creating the PEFT SAM wrapper here and in `get_sam_model`.

anwai98: Okay, I went an extended route to take care of a few minor, yet important things I wanted to be careful about. Now it should be GTG. Thanks for the suggestion!

constantinpape (approved):

PR #648, opened 2024-07-01
Api cleanup

constantinpape:


constantinpape:
Also, I created a new submodule `models` and moved everything related to custom models there, + created some unit tests for them.

micro_sam/models/sam_3d_wrapper.py
```diff
@@ -1,15 +1,15 @@
...
 
 from segment_anything.modeling.image_encoder import window_partition, window_unpartition
 from segment_anything.modeling import Sam
 
-from .util import get_sam_model
+from ..util import get_sam_model
 
 
-def get_3d_sam_model(
+def get_sam_3d_model(
```
constantinpape, line 12: I changed these names to match the order of `sam_3d` we use everywhere else, cc @lufre1

micro_sam/models/sam_3d_wrapper.py
```diff
@@ -63,25 +54,42 @@ def __init__(self, sam_model: Sam, freeze_encoder: bool):
             for param in self.sam_model.image_encoder.parameters():
                 param.requires_grad = False
 
-    # FIXME
-    # - handling of the image size here is wrong, this only works for square images
-    # - this does not take care of resizing
-    # unclear how batches are handled
-    def forward(self, batched_input, multimask_output, image_size) -> torch.Tensor:
-        return self._forward_train(batched_input, multimask_output, image_size)
+    def forward(
```
constantinpape, line 57: This now matches the input and output signature of the original SAM model.
cc @lufre1 @anwai98

micro_sam/training/util.py
```diff
@@ -43,8 +43,8 @@ def get_trainable_sam_model(
     checkpoint_path: Optional[Union[str, os.PathLike]] = None,
     freeze: Optional[List[str]] = None,
     return_state: bool = False,
-    use_lora: bool = False,
-    rank: Optional[int] = None,
+    lora_rank: Optional[int] = None,
+    lora_kwargs: Optional[Dict] = None,
```
constantinpape, line 47: I removed the `use_lora`, since it is redundant with `rank = None` and renamed `rank` -> `lora_rank` to have a more explicit name.
I added `lora_kwargs`, so that we can also customize the PEFT SAM layers from here.

anwai98: Looks good to me. Thanks!

anwai98 (approved):
Thanks @constantinpape for the refactoring. This is GTG from my side.

Let's merge this once you think this is ready.

(PS. I was thinking if `TrainableSAM` is adapted with `PEFT_Sam` now as the test case seems to me that it is build in two stages: `get_sam_model` + `PEFT_Sam`, where it could be done in one-stage with `lora_rank`, but that's really minor and I can take care of it later once we look at the LoRA experiments)

micro_sam/training/util.py
lufre1: For me as well :handshake:

constantinpape:
Thanks for checking everyone, I will go ahead and merge!

PR #650, opened 2024-07-04
642 train 3d model with lucchi data

lufre1:


development/predict_3d_model_with_lucchi.py
```diff
@@ -0,0 +1,191 @@
...
+):
+    """
+    """
+    for image_path in tqdm(image_paths, desc="Run inference for semantic segmentation with all images"):
+        image_name = os.path.basename(image_path)
+
+        assert os.path.exists(image_path), image_path
+
+        # Perform segmentation only on the semantic class
+        for i, (semantic_class_name, _) in enumerate(semantic_class_map.items()):
```
constantinpape, line 99: I would suggest to remove this part already, it doesn't make sense in the context here.

development/predict_3d_model_with_lucchi.py
```diff
@@ -0,0 +1,191 @@
...
+
+
+def transform_labels(y):
+    return (y > 0).astype("float32")
+
+
+def predict(args):
+
+    device = "cuda" if torch.cuda.is_available() else "cpu"
+    if args.checkpoint_path is not None:
```
constantinpape, line 133: I am not sure why you would ever run prediction without a checkpoint. I would not make this optional.

development/predict_3d_model_with_lucchi.py
```diff
@@ -0,0 +1,191 @@
...
+        if os.path.exists(args.checkpoint_path):
+            # model = load_model(checkpoint=args.checkpoint_path, device=device) # does not work
+            
+            cp_path = os.path.join(args.checkpoint_path, "", "best.pt")
+            print(cp_path)
+            model = get_sam_3d_model(device, n_classes=args.n_classes, image_size=args.patch_shape[1],
+                                     lora_rank=4,
+                                     model_type=args.model_type,
+                                     checkpoint_path=cp_path
+                                     )
```
constantinpape, line 143: This will not work to actually load the checkpoint. Please read the code I send you carefully and see how I use `torch_em.util.load_model` to get the checkpoint. Not that this quite complex way is now necessary because of the way how we mix the 3d adapter and LoRA. I will think about a better way at some point, but for now we need to use this work-around.

development/train_3d_model_with_lucchi.py
```diff
@@ -0,0 +1,186 @@
...
+            ),
+            mode=self.padding
+        )
+
+        raw = self._set_channels_for_inputs(raw)
+
+        return raw
+
+
+class LucchiSegmentationDataset(SegmentationDataset):
```
constantinpape, line 64: This can now be removed.

anwai98:
This is currently outdated, in terms of the 3d wrapper. There have been a lot of upstream changes to adapt this for other projects. I will go ahead and close this for now!

PR #654, opened 2024-07-09
Minor refactor to LoRA model initialization in `get_trainable_sam_model`

anwai98:


micro_sam/training/training.py
```diff
@@ -146,6 +147,8 @@ def train_sam(
     scheduler_kwargs: Optional[Dict[str, Any]] = None,
     save_every_kth_epoch: Optional[int] = None,
     pbar_signals: Optional[QObject] = None,
+    optimizer_class: Optional[Optimizer] = torch.optim.Adam,
```
constantinpape, line 150: I would switch to AdamW as the default here too.

anwai98: Okie, I will switch it right away.

Just to keep you posted, the current `micro-sam` models (including the generalists) have been trained using `Adam`.

constantinpape (approved):

PR #669, opened 2024-07-25
Update trainers to compute loss over low_res_mask

anwai98:


micro_sam/training/sam_trainer.py
```diff
@@ -117,6 +118,24 @@ def _compute_iou(self, pred, true, eps=1e-7):
...
+        # Convert the labels to "low_res_mask" shape
+        # First step is to use the logic from `ResizeLongestSide` to resize the longest side.
+        target_length = self.model.transform.target_length
+        target_shape = self.model.transform.get_preprocess_shape(y_one_hot.shape[2], y_one_hot.shape[3], target_length)
+        y_one_hot = F.interpolate(input=y_one_hot, size=target_shape)
+        # Next, we pad the remaining region to (1024, 1024)
+        h, w = y_one_hot.shape[-2:]
+        padh = self.model.sam.image_encoder.img_size - h
+        padw = self.model.sam.image_encoder.img_size - w
+        y_one_hot = F.pad(input=y_one_hot, pad=(0, padw, 0, padh))
```
constantinpape, line 133: Is this not just `ResizeLongsetSide` until here? Can't we just use it instead of duplicating the code?

micro_sam/training/trainable_sam.py
```diff
@@ -15,15 +15,18 @@ class TrainableSAM(nn.Module):
     Args:
         sam: The SegmentAnything Model.
         device: The device for training.
+        upsampled_masks: Whether to return the output masks in the original input shape.
     """
     def __init__(
         self,
         sam: Sam,
         device: Union[str, torch.device],
+        upsampled_masks: bool = True,
```
constantinpape, line 24: Let's remove all the changes here and just always return low res and upsampled masks.

micro_sam/training/sam_trainer.py
anwai98: It's because of one parameter that I have to remplement this: (they set antialias to True, which for labels here, we need to set to False.) https://github.com/facebookresearch/segment-anything/blob/main/segment_anything/utils/transforms.py#L63-L65

micro_sam/training/trainable_sam.py
anwai98: This is restored now!

micro_sam/training/sam_trainer.py
anwai98: Also to add: with `ResizeLongestSide`, it's still a two stage agenda: (taking LIVECell (images of shape `(520, 704)` as an example)

- `ResizeLongestSide` would only adjust the inputs to the longest side: (eg. `(520, 704)` -? `(720, 1024)`).
    - https://github.com/facebookresearch/segment-anything/blob/main/segment_anything/utils/transforms.py#L55-L65
- There's still the next stage where the padding takes place to complete the image shape to the desired input shape: (eg. `(720, 1024)` -> `(1024, 1024)`)
    - This happens here: https://github.com/facebookresearch/segment-anything/blob/main/segment_anything/modeling/sam.py#L164-L174

constantinpape: I see, makes sense!

anwai98:
I'll close this in favor of https://github.com/computational-cell-analytics/micro-sam/pull/990

PR #673, opened 2024-07-29
Fix issue in AIS precomputation and extend training functionality

constantinpape:


anwai98:
AIS works as expected for me as well now. Thanks!

PR #678, opened 2024-08-27
Start implementation for automated tracking

constantinpape:
pending https://github.com/funkelab/motile/issues/98 and proper GUI integration


constantinpape:
Tracking now uses trackastra and this works in principle, but there are still some minor things to take care of. Also, we should get trackastra on conda-forge to properly depend on it. I will follow up on this.

PR #680, opened 2024-09-01
FacT-Finetuning SAM

caroteu:
Implementation of FacT Finetuning method on SAM. @anwai98

### DIFF
--- a/finetuning/evaluation/evaluate_instance_segmentation.py
+++ b/finetuning/evaluation/evaluate_instance_segmentation.py
@@ -7,7 +7,7 @@ from util import get_paths  # comment this and create a custom function with the
 from util import get_pred_paths, get_default_arguments
 
 
-def run_instance_segmentation_with_decoder_inference(dataset_name, model_type, checkpoint, experiment_folder):
+def run_instance_segmentation_with_decoder_inference(dataset_name, model_type, checkpoint, experiment_folder, lora_rank):
     val_image_paths, val_gt_paths = get_paths(dataset_name, split="val")
     test_image_paths, _ = get_paths(dataset_name, split="test")
     prediction_folder = run_instance_segmentation_with_decoder(
@@ -16,7 +16,8 @@ def run_instance_segmentation_with_decoder_inference(dataset_name, model_type, c
         experiment_folder,
         val_image_paths,
         val_gt_paths,
-        test_image_paths
+        test_image_paths,
+        lora_rank=lora_rank,
     )
     return prediction_folder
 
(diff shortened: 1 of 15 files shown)


finetuning/evaluation/evaluate_instance_segmentation.py
```diff
@@ -35,9 +35,9 @@ def eval_instance_segmentation_with_decoder(dataset_name, prediction_folder, exp
 
 def main():
     args = get_default_arguments()
-
+    peft_kwargs = {"rank": args.peft_rank, "module": args.peft_module}
```
caroteu, line 38: @anwai98 I changed the way the way the peft arguments are passed in the evaluation because I thought this would make things more flexible for different methods. Not sure if this is the most elegant way to do it

finetuning/evaluation/util.py
```diff
@@ -80,12 +80,10 @@ def get_dataset_paths(dataset_name, split_choice):
     return raw_dir, labels_dir
 
 
-def get_model(model_type, ckpt, lora_rank):
```
caroteu, line 83: @anwai98 This function should be unnecessary and can be replaced by directly calling get_sam_model

micro_sam/models/peft_sam.py
```diff
@@ -56,6 +56,49 @@ def forward(self, x):
...
+    ):
+        super().__init__()
+        self.qkv_proj = block.attn.qkv
+        self.dim = self.qkv_proj.in_features 
+
+        self.q_FacTs = nn.Linear(rank, rank, bias=False)
+        self.v_FacTs = nn.Linear(rank, rank, bias=False)
+
+        # NOTE : Dropout is not included in the original implementation
+        self.dp_q = nn.Dropout(0.1)
```
caroteu, line 81: @anwai98 Let's discuss whether or not it makes sense to keep the dropout layers, they were included in the MA-SAM implementation but not mentioned in the FacT paper

micro_sam/models/peft_sam.py
```diff
@@ -79,6 +122,10 @@ def __init__(
 
         assert rank > 0
 
+        dim = model.image_encoder.blocks[0].attn.qkv.in_features
+        self.FacTu = nn.Linear(dim, rank, bias=False)
```
caroteu, line 126: @anwai98 Do we need to add a condition here, so that these parameters are only initialized when FacT is used?

anwai98:
I'll close this PR in favor of moving it to https://github.com/computational-cell-analytics/micro-sam/pull/682.

finetuning/evaluation/evaluate_instance_segmentation.py
anwai98: I agree. This would be the most elegant way to handle the current functionalities.

finetuning/evaluation/util.py
anwai98: Yep, makes sense to get rid of it. I took care of it. Will push the changes in the new PR soon.

micro_sam/models/peft_sam.py
anwai98: Hmm, we definitely cannot hard-code them in the current location, would create a bit of a mess. While reviewing, Ill try to see how we can make it flexible upon the peft module.

PR #682, opened 2024-09-05
Add FaCT Finetuning for SAM

anwai98:
Hi @caroteu,

I moved your fork's branch to `micro-sam` repo with all your changes in place.

I'll review the PR soon (also check out your discussions from the other PR) and ping you here!. Thanks!


micro_sam/models/peft_sam.py
```diff
@@ -56,6 +56,49 @@ def forward(self, x):
...
+        super().__init__()
+        self.qkv_proj = block.attn.qkv
+        self.dim = self.qkv_proj.in_features 
+
+        self.q_FacTs = nn.Linear(rank, rank, bias=False)
+        self.v_FacTs = nn.Linear(rank, rank, bias=False)
+
+        # NOTE : Dropout is not included in the original implementation
+        self.dp_q = nn.Dropout(0.1)
+        self.dp_v = nn.Dropout(0.1)
```
anwai98, line 82: > Let's discuss whether or not it makes sense to keep the dropout layers, they were included in the MA-SAM implementation but not mentioned in the FacT paper

Okay, I'll check out the details of it and get back on this (the easiest way would be to do experiments with both to understand this better. Let's wait on it till the first review round)

micro_sam/training/util.py
```diff
@@ -84,8 +83,9 @@ def get_trainable_sam_model(
     # NOTE: This is done exclusive to "get_sam_model" here to use PEFT's layer-specific initialization on top.
     # Whether to use Parameter Efficient Finetuning methods to wrap around Segment Anything.
     # Overwrites the SAM model by freezing the backbone and allow low rank adaption to attention layers.
-    if lora_rank is not None:
-        sam = custom_models.peft_sam.PEFT_Sam(sam, rank=lora_rank, **({} if lora_kwargs is None else lora_kwargs)).sam
+    if peft_kwargs is not None:
+        assert peft_kwargs['module'] in ['LoRASurgery', 'FacTSurgery'], "Invalid PEFT module."
+        sam = custom_models.peft_sam.PEFT_Sam(sam, rank=peft_kwargs['rank'], peft_module=peft_kwargs['module']).sam
```
caroteu, line 88: This part should probably be changed equivalent to the refactored version in get_sam_model @anwai98 
```suggestion
    if peft_kwargs and isinstance(peft_kwargs, dict):
        if model_type[:5] == "vit_t":
            raise ValueError("'micro-sam' does not support parameter efficient finetuning for 'mobile-sam'.")

        peft_module = peft_kwargs.get("peft_module")
        if peft_module is not None:
            from micro_sam.models.peft_sam import LoRASurgery, FacTSurgery
            assert peft_module in [LoRASurgery, FacTSurgery], "Invalid PEFT module."

        sam = custom_models.peft_sam.PEFT_Sam(sam, **peft_kwargs).sam
```

anwai98: Yep, this should change as well. Thanks for spotting. I'll commit this in.

micro_sam/models/peft_sam.py
anwai98: For now, we decided to make this optional and do some tests on it's behavior with and without. Will switch the defaults later as per the experimental outcome.

anwai98:
Hi @constantinpape,

The FacT PR is good to go from our side (thanks to @caroteu). We have added an additional test for FacT method. We are working on doing some regression tests to understand a few minor details, but overall the method is good for another pair of eyes to review. Let us know if you spot something.

PS. We made a slight design change in `get_sam_model` (and linked functions) in favor of PEFT's usability.

test/test_models/test_peft_sam.py
```diff
@@ -1,17 +1,32 @@
 import unittest
 
+from micro_sam.models.peft_sam import FacTSurgery, LoRASurgery
```
constantinpape, line 3: I would move these imports in the corresponding tests. The idea here is to move the imports of the functionality being tested into the tests, so that we get a specific failure in the test if the import fails.

constantinpape (commented):
Looks good overall. Just one minor comment about the tests.

test/test_models/test_peft_sam.py
anwai98: Okay, yes. That's important. Thanks for the hint!

PR #692, opened 2024-09-22
Extend iterative prompt generators to return prompts for 3d

anwai98:
I've made some super minor changes to the iterative prompt generator, to support for coordinate information in z-direction (in addition to the already provided coordinates of Y and X).

I'll discuss the need for this support tomorrow in detail. I'll leave this open for now!


micro_sam/prompt_generators.py
```diff
@@ -320,24 +332,33 @@ def __call__(
...
         device = prediction.device
+        assert segmentation.shape == prediction.shape, \
+            "The segmentation and prediction tensors should have the same shape."
+
+        if segmentation.ndim == 5:  # masks in 3d must be tensors of shape NUM_OBJECTS x 1 x Z x H x W
+            is_3d = True
+        elif segmentation.ndim == 4:  # masks in 2d must be tensors of shape NUM_OBJECTS x 1 x H x W
+            is_3d = False
+        else:
+            raise ValueError("The segmentation and prediction tensors should have either '4' or '5' dimensions.")
```
anwai98, line 352: @constantinpape

Okay, there was a slight mistake from my side while explaining this, but now things should be in place now. (I've left comments on the findings and the latest setup here)

anwai98: And the tests pass as well with all the checks in place.

micro_sam/prompt_generators.py
```diff
@@ -252,64 +252,76 @@ def __call__(
...
         ]
-        # we sample one location for each object in the batch
+        # we sample one positive location for each object in the batch
         sampled_indices = [np.random.choice(len(pos_loc[0])) for pos_loc in positive_locations]
-        # get the corresponding coordinates (Note that we flip the axis order here due to the expected order of SAM)
-        pos_coordinates = [
-            [pos_loc[-1][idx], pos_loc[-2][idx]] for pos_loc, idx in zip(positive_locations, sampled_indices)
-        ]
+        # get the corresponding coordinates (NOTE: we flip the axis order here due to the expected order of SAM)
+        pos_coordinates = []
```
constantinpape, line 266: This is a minor comment, but you should try to keep list comprehensions whenever possible since they are faster. 
That is quite easy here. you can do:
```python
if is_3d:
   pos_coordinates = [....]
else:
  pos_coordinates = [...]
```
instead of getting rid of the list comprehension and writing out the full loop.

micro_sam/prompt_generators.py
```diff
@@ -252,64 +252,76 @@ def __call__(
...
+        # hence we might not find points sometimes there as well. therefore, we sample points from true background.
+        negative_locations = [
+            torch.where(true_obj == 0) if len(neg_loc[0]) == 0 else neg_loc
+            for neg_loc, true_obj in zip(negative_locations, true_object)
+        ]
+        # we sample one negative location for each object in the batch
+        sampled_indices = [np.random.choice(len(neg_loc[0])) for neg_loc in negative_locations]
+        # get the corresponding coordinates (NOTE: we flip the axis order here due to the expected order of SAM)
+        neg_coordinates = []
+        for neg_loc, idx in zip(negative_locations, sampled_indices):
```
constantinpape, line 314: See comment about list comprehensions above.

constantinpape (commented):
Only two minor comments, the rest looks good!

micro_sam/prompt_generators.py
anwai98: Okay, I'll keep that into account. This should be taken care of now. Thanks for the mention!

constantinpape (approved):

constantinpape:
There is some problem with the CI, the errors are unrelated and I will merge.

PR #696, opened 2024-09-24
Fix clearing shape layer

### LINKED ISSUE #695: Compatibility issues with napari 0.5
Napari 0.5 has some compatibility issues we should fix. Here are the ones @anwai98 found so far:
- [x] Keybindings don't work anymore
- [x] Box prompts are not cleared after committing the segmentation
- [x] The input image seems to flicker (appear and disappear) after committing a few segmented objects
- [x] Commit doesn't clear annotations (I think this is related to the first point)
- [x] Current image is still show in Image series annotator

constantinpape:
Fixes #695

### DIFF
--- a/environment_cpu.yaml
+++ b/environment_cpu.yaml
@@ -7,7 +7,7 @@ dependencies:
     - nifty =1.2.1=*_4
     - imagecodecs
     - magicgui
-    - napari <0.5
+    - napari
     - pip
     - pooch
     - pyqt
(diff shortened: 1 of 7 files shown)


constantinpape:
This fixes one of the issues with 0.5 (shape layer not being correctly removed).
@anwai98 can you make a PR onto this one that adds back the changes with the border attributes? Thanks!

constantinpape:
I think we should rather merge this on dev, so that we can have the napari 0.5 support in the next release. I will change this later.

PR #699, opened 2024-09-24
Add automatic segmentation cli

anwai98:


micro_sam/instance_segmentation.py
```diff
@@ -1139,3 +1143,145 @@ def get_amg(
...
+
+
+def automatic_instance_segmentation(
+    input_path: Union[os.PathLike, str],
+    output_path: Union[os.PathLike, str],
+    embedding_path: Optional[Union[os.PathLike, str]] = None,
+    model_type: str = util._DEFAULT_MODEL,
+    checkpoint_path: Optional[Union[os.PathLike, str]] = None,
+    key: Optional[str] = None,
+    ndim: Optional[int] = None,
```
constantinpape, line 1155: The current code will only work for 2d images. So it doesn't really make sense to expose `ndim` here.
We have two options: don't expose `ndim` or extend the function that we also support 3d segmentation by using functionality from `multi_dimensional_segmentation`.  We have to move this to a different file if we want to go with the latter option to avoid circular imports.

micro_sam/instance_segmentation.py
```diff
@@ -1139,3 +1143,145 @@ def get_amg(
...
+        model_type: The SegmentAnything model to use. Will use the standard vit_l model by default.
+        checkpoint_path: Path to a checkpoint for a custom model.
+        key: The key to the input file. This is needed for container files (eg. hdf5 or zarr)
+            or to load several images as 3d volume. Provide a glob patterm, eg. "*.tif", for this case.
+        ndim: The dimensionality of the data.
+        tile_shape: Shape of the tiles for tiled prediction. By default prediction is run without tiling.
+        halo: Overlap of the tiles for tiled prediction.
+        use_amg: Whether to use Automatic Mask Generation (AMG) as the automatic segmentation method.
+    """
+    predictor, state = util.get_sam_model(model_type=model_type, checkpoint_path=checkpoint_path, return_state=True)
```
constantinpape, line 1177: We load the checkpoint twice if we do this here and in line 1178. It would be good to load only once to avoid overhead.

micro_sam/instance_segmentation.py
```diff
@@ -1139,3 +1143,145 @@ def get_amg(
...
+        "-n", "--ndim", type=int, default=None,
+        help="The number of spatial dimensions in the data. Please specify this if your data has a channel dimension."
+    )
+    parser.add_argument(
+        "--amg", action="store_true", help="Whether to use automatic mask generation with the model."
+    )
+
+    args, parameter_args = parser.parse_known_args()
+    # NOTE: the script below allows the possibility to catch additional parsed arguments which correspond to
+    # the automatic segmentation post-processing parameters (eg. 'center_distance_threshold' in AIS)
```
constantinpape, line 1267: Nice!

constantinpape (commented):
Very good start! We should see how we handle 3d segmentation, and I would improve efficiency in loading the mode, see comments for detail.

micro_sam/instance_segmentation.py
anwai98: Yes, true. This one should be updated now!

micro_sam/instance_segmentation.py
anwai98: Thanks for the feedback!

I went ahead and added support for 3d segmentation. However, I spotted the [TODO](https://github.com/computational-cell-analytics/micro-sam/blob/master/micro_sam/multi_dimensional_segmentation.py#L356) for tiling-based prediction, which is probably not fully implemented. I'll leave the comment open so that we discuss this tomorrow.

anwai98: Ah, it does work. We had to the changes I made already in the code, and that's pretty much it (tested it just now on `lucchi_test.h5` with tile 512 and halo 128 - looks as expected).

micro_sam/automatic_segmentation.py
```diff
@@ -0,0 +1,181 @@
...
+    else:  # AMG
+        segmenter = get_amg(predictor=predictor, is_tiled=tile_shape is not None)
+
+    # Load the input image file.
+    if isinstance(input_path, np.ndarray):
+        image_data = input_path
+    else:
+        image_data = util.load_image_data(input_path, key)
+
+    if ndim == 3 or image_data.ndim == 3:
```
constantinpape, line 58: We should check for invalid combinations here, e.g. if `ndim == 3` but `image_data.ndim == 2`.

micro_sam/automatic_segmentation.py
```diff
@@ -0,0 +1,181 @@
...
+            elif isinstance(segmenter, AMGBase):
+                this_shape = segmenter._original_size
+            else:
+                this_shape = image_data.shape[-2:]
+
+            instances = np.zeros(this_shape, dtype="uint32")
+        else:
+            instances = mask_data_to_segmentation(masks, with_background=True, min_object_size=0)
+
+    # Save the instance segmentation
```
constantinpape, line 94: I would make the output path optional, and return the segmentation. This makes the function more flexible and we can re-use it independent from the CLI.

constantinpape (commented):
Looks good now! Only two more minor things.

micro_sam/automatic_segmentation.py
anwai98: Yes, makes sense. I've refactored this in a way such that while using the CLI, `output_path` is a mandatory argument, however it is not the same for the `automatic_instance_segmentation`.

micro_sam/automatic_segmentation.py
anwai98: Okie. It's updated now!

constantinpape (approved):

Issue #700, opened 2024-09-25
Add tests for automatic instance segmentation function


constantinpape:
that was just added to `dev` in #699

PR #702, opened 2024-09-25
Add tests for automatic segmentation function

anwai98:
Takes care of https://github.com/computational-cell-analytics/micro-sam/issues/700.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -42,11 +42,15 @@ jobs:
         shell: bash -l {0}
         run: pip install --no-deps -e .
 
-      - name: Run tests
+      - name: Install pytest-xdist for parallel testing
+        shell: bash -l {0}
+        run: python -m pip install pytest-xdist
+
+      - name: Run tests in parallel
         uses: aganders3/headless-gui@v2
         with:
           shell: bash -l {0}
-          run: pytest
+          run: pytest -n auto
 
       - name: Upload coverage reports to Codecov
         uses: codecov/codecov-action@v3
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -42,15 +42,11 @@ jobs:
         shell: bash -l {0}
         run: pip install --no-deps -e .
 
-      - name: Install pytest-xdist for parallel testing
-        shell: bash -l {0}
-        run: python -m pip install pytest-xdist
-
-      - name: Run tests in parallel
+      - name: Run tests
         uses: aganders3/headless-gui@v2
         with:
           shell: bash -l {0}
-          run: pytest -n auto
+          run: pytest
 
       - name: Upload coverage reports to Codecov
         uses: codecov/codecov-action@v3
--- a/test/test_automatic_segmentation.py
+++ b/test/test_automatic_segmentation.py
@@ -9,7 +9,7 @@ import micro_sam.util as util
 
 class TestAutomaticSegmentation(unittest.TestCase):
     model_type = "vit_t" if util.VIT_T_SUPPORT else "vit_b"
-    model_type_ais = "vit_t_lm"
+    model_type_ais = "vit_t_lm" if util.VIT_T_SUPPORT else "vit_b_lm"
     tile_shape = (512, 512)
     halo = (96, 96)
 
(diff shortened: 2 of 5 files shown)


anwai98:
One concern I have here is the amount of time the tests are taking now.

I added the tests for automatic segmentation in 3d images, and that seems to take ~13 minutes on GPU. (maybe I can reduce / remove some of the tests there or perform it for a smaller volume)

test/test_automatic_segmentation.py
```diff
@@ -0,0 +1,153 @@
...
+    def test_automatic_mask_generator_3d(self):
+        from micro_sam.automatic_segmentation import automatic_instance_segmentation
+
+        labels, volume = self.labels,  self.volume
+        instances = automatic_instance_segmentation(
+            input_path=volume, model_type=self.model_type, ndim=3, use_amg=True
+        )
+        self.assertEqual(labels.shape, instances.shape)
+
+    def test_tiled_automatic_mask_generator_3d(self):
```
constantinpape, line 118: I assume this is the long running test, correct? Let's skip it. 
You can use `@unittest.skip` for this.

test/test_automatic_segmentation.py
```diff
@@ -0,0 +1,153 @@
...
+    def test_tiled_instance_segmentation_with_decoder_2d(self):
+        from micro_sam.automatic_segmentation import automatic_instance_segmentation
+
+        mask, image = self.large_mask,  self.large_image
+        instances = automatic_instance_segmentation(
+            input_path=image, model_type=self.model_type, ndim=2, tile_shape=self.tile_shape, halo=self.halo,
+        )
+        self.assertEqual(mask.shape, instances.shape)
+
+    def test_automatic_mask_generator_3d(self):
```
constantinpape, line 109: If this one takes too long we can also skip it.

test/test_automatic_segmentation.py
```diff
@@ -0,0 +1,153 @@
...
+
+    @classmethod
+    def setUpClass(cls):
+        # Input 2d data for normal and tiled segmentation.
+        cls.mask, cls.image = cls._get_2d_inputs()
+        cls.large_mask, cls.large_image = cls._get_2d_inputs(shape=(1024, 1024))
+
+        # Input 3d data for normal and tiled segmentation.
+        cls.labels, cls.volume = cls._get_3d_inputs()
+        cls.large_labels, cls.large_volume = cls._get_3d_inputs(shape=(8, 1024, 1024))
```
constantinpape, line 56: Let's use fewer slice to speed things up. I suggest using 4.

test/test_automatic_segmentation.py
```diff
@@ -0,0 +1,153 @@
...
+        return mask, image
+
+    # create an input 2d image with three objects and stack 8 of them together
+    @classmethod
+    def _get_3d_inputs(cls, shape=(8, 256, 256)):
+        mask, image = cls._get_2d_inputs(shape[-2:])
+
+        # Create volumes by stacking the input image and respective mask.
+        volume = np.stack([image] * 8)
+        labels = np.stack([mask] * 8)
```
constantinpape, line 45: You should use `shape[0]` rather than hard-coding this.

test/test_automatic_segmentation.py
anwai98: Okay, makes sense. I'll skip this one.

test/test_automatic_segmentation.py
anwai98: Yep, this does take a while as well. I'll skip this too.

test/test_automatic_segmentation.py
anwai98: Fixed. Thanks for spotting!

.github/workflows/test.yaml
```diff
@@ -42,11 +42,15 @@ jobs:
         shell: bash -l {0}
         run: pip install --no-deps -e .
 
-      - name: Run tests
+      - name: Install pytest-xdist for parallel testing
```
constantinpape, line 45: This is probably a bad idea, as oversupscription of CPUs and RAM will slow down tests.

anwai98: Ah I see. Okay, I'll revert the script to remove parallelization. Thanks for the heads-up.

anwai98:
Okay, the tests are definitely faster while avoiding parallelization, and faster with rectangular tiles as well. The tests pass for ubuntu and macos in ~30mins. The windows tests fails for some reason (is it a patch-related issue while trying to setup using micromamba?)

constantinpape:
FYI @anwai98 I reduced the test times significantly by enabling passing kwargs to the AMG function and reducing the points per side, as well as fixing two typos that caused using the wrong segmentation mode in the tests for instance segmentation with decoder.

Issue #703, opened 2024-09-26
Support dry run for dataloaders to check if valid instances available


anwai98:
I think it would make sense to perform a dry run on the dataloaders to check whether the instance labels are as expected (i.e. there has to be atleast one object instance to perform finetuning), else notify the user (even before allowing to finetune the model) about annotating some objects / adding sampler in the dataset, etc.

Ref: https://github.com/computational-cell-analytics/micro-sam/blob/master/micro_sam/training/training.py#L130

anwai98:
This is taken care of at https://github.com/computational-cell-analytics/micro-sam/pull/705.

PR #705, opened 2024-09-27
Add dry run for loaders to check for valid instances

anwai98:
Resolves https://github.com/computational-cell-analytics/micro-sam/issues/703


micro_sam/training/training.py
```diff
@@ -71,38 +73,44 @@ def check_instance_channel(instance_channel):
...
-    else:
-        if n_channels_y != 1:
-            raise ValueError(
-                "Invalid number of channels in the target data from the data loader. "
-                "Expect 1 channel for training without an instance segmentation decoder,"
-                f"but got {n_channels_y} channels."
-            )
-        check_instance_channel(y)
+    name = "" if name is None else f"'{name}'"
+    for x, y in tqdm(loader, desc=f"Verifying labels in {name} dataloader"):
```
constantinpape, line 77: I am a bit hesitant to do this by default. If we have a large data loader it may delay the start of training quite a bit. Let's discuss on Monday. I think the best solution would be to define a number of samples that will be checked to make this flexible.

constantinpape:
Fyi @anwai98 I changed the target branch to `dev`, since this contained all recent `dev` commits.

micro_sam/training/training.py
anwai98: I exposed an additional arguments which controls whether we check labels for the entire dataloader or for the first n samples. I have left the current defaults for it to `None`, that corresponds to checking the entire dataloader.

micro_sam/training/training.py
```diff
@@ -157,6 +174,7 @@ def train_sam(
     pbar_signals: Optional[QObject] = None,
     optimizer_class: Optional[Optimizer] = torch.optim.AdamW,
     peft_kwargs: Optional[Dict] = None,
+    verify_n_labels_in_loader: Optional[int] = None,
```
constantinpape, line 177: I would set the default to something rather small (maybe 50). Otherwise `None` causes quite long wait times for a big data loader.

anwai98: Yep, that sounds reasonable. I've updated the defaults now to `50`. Thanks!

anwai98:
Yayy, the windows CI works now! 🥳 

cc: @constantinpape

constantinpape (approved):

PR #710, opened 2024-09-29
Napari updates

constantinpape:


constantinpape:
See #696 for details (this is just fixing stuff so that the PR goes onto dev instead of master)

PR #711, opened 2024-09-30
Overwrite keybindings of prompt layers

### LINKED ISSUE #695: Compatibility issues with napari 0.5

constantinpape:
Fixes #695


psobolewskiPhD:
Glad you got this working, I've really got to check out this plugin!
🚀

PR #714, opened 2024-10-01
Update dropout defaults for FacT finetuning method

anwai98:
I updated the default value for dropout (from our initial experiments, looks like dropout 0.1 works better compared to no dropout or other dropout values).

cc: @caroteu - I also removed the comment for the scaling factor as we discussed.


anwai98:
The issue with CIs seems to persist. I'll go ahead and merge this as they are unrelated to the this PR.

PR #718, opened 2024-10-02
Expose kwargs to allow changing peft module arguments

anwai98:
Hi @caroteu,

Thanks for the PR at https://github.com/computational-cell-analytics/micro-sam/pull/717. I've taken care of a couple of merge conflicts and updated the design to support `dropout` as a parameter to `PEFT_Sam` and the possibility to pass other parameters we might need in future.


anwai98:
Okay, the issue with CIs seems to persist (the runners are still stuck). I'll go ahead and merge this to `dev`.

PS. I checked the peft-related tests locally and it works.

PR #722, opened 2024-10-02
Minor update to ignore warnings in train_sam functionality

anwai98:


constantinpape:
Hi @anwai98 ,
I updated this a little bit, so that the warnings only get suppressed for the training function, but not for everything else.

PR #724, opened 2024-10-04
Update docs - add hints on starting napari

anwai98:
I added some explanations and extending parts to hint on starting `napari` from the activated environment.

Ref: https://forum.image.sc/t/micro-sam-unclear-if-previous-installation-of-napari-is-required/103303


anwai98:
Just to keep you posted @constantinpape,

I spotted the reason for windows CIs breaking: https://github.com/mamba-org/setup-micromamba - seems like it has something to do with a recent build at micromamba. They are working on the issue to release micromamba 2.0.2, which should fix this issue.

PR #727, opened 2024-10-08
Add SSF method

anwai98:
Superseeds https://github.com/computational-cell-analytics/micro-sam/pull/726.

Hi @caroteu,
Thanks for adding the method.

I gave it a refactor so that we avoid adding components to `PEFT_Sam` and rely on surgery classes for implementing peft methods. And in general for PEFT methods in future, we should rely on performing all updates via surgery classes and keep the minimal design of the peft sam functionality.

There's also a major issue anyways (observed this in your PR, tried to take a look at it, couldn't manage to find out what's going on), but the SSF method does not work atm (i.e. this can be easily checked with the change in parameters when we freeze the image encoder vs SSF - the parameter count is the same + there are no parameters which expect gradient updates, which is not supposed to be the case as we update the scale and shift parameters).

I suggest you take over and see if you can figure out the issue.

(In case you figure out something, create a PR from the remote at https://github.com/computational-cell-analytics/micro-sam to the branch `add-ssf` - this would help us develop over PRs together)

Let me know if you have any questions.


anwai98:
This seems to work now. I think we can merge this as our experiments are in place for this method.

cc: @constantinpape

constantinpape (approved):

constantinpape:
I had to fix some merge conflicts, will merge once the tests pass.

Issue #719, opened 2024-10-02
Add out of domain test datasets for benchmarking finetuned models


anwai98:
The idea for this would be to have scripts unified in one place for evaluating automatic and interactive segmentation capabilities of `micro-sam` on a diverse set of open-source datasets (i.e. download them, preprocess them, run inference with new finetuned models and generate relevant metric scores) (ideally, we would like to support inference in both 2d and 3d - wherever relevant).

anwai98:
This is resolved by https://github.com/computational-cell-analytics/micro-sam/pull/728.

PR #728, opened 2024-10-09
Add CLI for benchmarking datasets on SAM models

anwai98:
The prototype CLI for benchmarking microscopy datasets on SAM models is ready for the first round of review (tested on LIVECell and DeepBacs)

I'll work on testing it on 3d datasets and extend interactive segmentation functionality for 3d (using SAM)

### DIFF
--- a/micro_sam/automatic_segmentation.py
+++ b/micro_sam/automatic_segmentation.py
@@ -1,11 +1,9 @@
 import os
 from pathlib import Path
-from typing import Dict, Optional, Union, Tuple
+from typing import Optional, Union, Tuple
 
 import numpy as np
 import imageio.v3 as imageio
-
-import torch
 
 from . import util
 from .instance_segmentation import (
@@ -15,57 +13,38 @@ from .multi_dimensional_segmentation import automatic_3d_segmentation
 
 
 def automatic_instance_segmentation(
+    predictor: util.SamPredictor,
+    segmenter: Union[AMGBase, InstanceSegmentationWithDecoder],
     input_path: Union[Union[os.PathLike, str], np.ndarray],
     output_path: Optional[Union[os.PathLike, str]] = None,
     embedding_path: Optional[Union[os.PathLike, str]] = None,
-    device: Optional[Union[torch.device, str]] = None,
-    model_type: str = util._DEFAULT_MODEL,
-    checkpoint_path: Optional[Union[os.PathLike, str]] = None,
     key: Optional[str] = None,
     ndim: Optional[int] = None,
     tile_shape: Optional[Tuple[int, int]] = None,
     halo: Optional[Tuple[int, int]] = None,
-    use_amg: bool = False,
-    amg_kwargs: Optional[Dict] = None,
     verbose: bool = True,
     **generate_kwargs
 ) -> np.ndarray:
     """Run automatic segmentation for the input image.
 
     Args:
+        predictor: The SAM model.
+        segmenter: The instance segmentation class.
         input_path: input_path: The input image file(s). Can either be a single image file (e.g. tif or png),
             or a container file (e.g. hdf5 or zarr).
         output_path: The output path where the instance segmentations will be saved.
         embedding_path: The path where the embeddings are cached already / will be saved.
-        model_type: The SegmentAnything model to use. Will use the standard vit_l model by default.
-        checkpoint_path: Path to a checkpoint for a custom model.
         key: The key to the input file. This is needed for container files (eg. hdf5 or zarr)
             or to load several images as 3d volume. Provide a glob patterm, eg. "*.tif", for this case.
         ndim: The dimensionality of the data.
         tile_shape: Shape of the tiles for tiled prediction. By default prediction is run without tiling.
         halo: Overlap of the tiles for tiled prediction.
-        use_amg: Whether to use Automatic Mask Generation (AMG) as the automatic segmentation method.
-        amg_kwargs: optional keyword arguments for creating the AMG or AIS class.
-        generate_kwargs: optional keyword arguments for the generate function onf the AMG or AIS class.
+        verbose: Verbosity flag.
+        generate_kwargs: optional keyword arguments for the generate function of the AMG or AIS class.
 
     Returns:
         The segmentation result.
     """
-    predictor, state = util.get_sam_model(
-        model_type=model_type, device=device, checkpoint_path=checkpoint_path, return_state=True,
-    )
-
-    if "decoder_state" in state and not use_amg:  # AIS
-        decoder = get_decoder(predictor.model.image_encoder, state["decoder_state"])
-        segmenter = get_amg(
-            predictor=predictor, decoder=decoder, is_tiled=tile_shape is not None,
-            **({} if amg_kwargs is None else amg_kwargs)
-        )
-    else:  # AMG
-        segmenter = get_amg(
-            predictor=predictor, is_tiled=tile_shape is not None, **({} if amg_kwargs is None else amg_kwargs)
-        )
-
     # Load the input image file.
     if isinstance(input_path, np.ndarray):
         image_data = input_path
@@ -187,17 +166,28 @@ def main():
         parameter_args[i].lstrip("--"): _convert_argval(parameter_args[i + 1]) for i in range(0, len(parameter_args), 2)
     }
 
+    device = util._get_default_device
+    predictor, state = util.get_sam_model(
+        model_type=args.model_type, device=device, checkpoint_path=args.checkpoint, return_state=True,
+    )
+
+    segmenter = get_amg(
+        predictor=predictor, is_tiled=args.tile_shape is not None,
+        decoder=get_decoder(
+            image_encoder=predictor.model.image_encoder, decoder_state=state["decoder_state"], device=device
+        ) if "decoder_state" in state and not args.amg else None
+    )
+
     automatic_instance_segmentation(
+        predictor=predictor,
+        segmenter=segmenter,
         input_path=args.input_path,
         output_path=args.output_path,
         embedding_path=args.embedding_path,
-        model_type=args.model_type,
-        checkpoint_path=args.checkpoint,
         key=args.key,
         ndim=args.ndim,
         tile_shape=args.tile_shape,
         halo=args.halo,
-        use_amg=args.amg,
         **generate_kwargs,
     )
 
--- a/micro_sam/automatic_segmentation.py
+++ b/micro_sam/automatic_segmentation.py
@@ -1,6 +1,6 @@
 import os
 from pathlib import Path
-from typing import Optional, Union, Tuple
+from typing import Optional, Union, Tuple, Dict
 
 import numpy as np
 import imageio.v3 as imageio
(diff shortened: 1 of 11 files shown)


micro_sam/evaluation/benchmark_datasets.py
```diff
@@ -0,0 +1,448 @@
...
+        zip(image_paths, gt_paths), total=len(image_paths),
+        desc=f"Extracting patches for {dataset_choice}"
+    ):
+        image = imageio.imread(image_path)
+        gt = imageio.imread(gt_path)
+
+        if len(np.unique(gt)) == 1:  # There could be labels which does not have any annotated foreground.
+            continue
+
+        tiling = blocking([0, 0], gt.shape, tile_shape)
```
constantinpape, line 209: This will not work for 3d data, we need to passs `[0, 0, 0]` for this. It's probably best to define a variable `origin` for this. See above.

micro_sam/evaluation/benchmark_datasets.py
```diff
@@ -0,0 +1,448 @@
...
+            available_datasets[choice]()
+        else:
+            raise ValueError(f"'{choice}' is not a supported choice of dataset.")
+
+    return dataset_choice
+
+
+def _extract_slices_from_dataset(path, dataset_choice):
+    if dataset_choice in LM_2D_DATASETS or EM_2D_DATASETS:
+        ndim, tile_shape = 2, (512, 512)
```
constantinpape, line 170: `origin = [0, 0]`

micro_sam/evaluation/benchmark_datasets.py
```diff
@@ -0,0 +1,448 @@
...
+    else:
+        ndim, tile_shape = 3, (32, 512, 512)
```
constantinpape, line 172: `origin = [0, 0, 0]`

micro_sam/evaluation/benchmark_datasets.py
```diff
@@ -0,0 +1,448 @@
...
+        image = imageio.imread(image_path)
+        gt = imageio.imread(gt_path)
+
+        if len(np.unique(gt)) == 1:  # There could be labels which does not have any annotated foreground.
+            continue
+
+        tiling = blocking([0, 0], gt.shape, tile_shape)
+        n_tiles = tiling.numberOfBlocks
+        tiles = [tiling.getBlock(tile_id) for tile_id in range(n_tiles)]
+        crop_boxes = [[tile.begin[1], tile.begin[0], tile.end[1], tile.end[0]] for tile in tiles]
```
constantinpape, line 212: Again, needs some update for 3d.

constantinpape: It would be easiest if you just create a list of `tuple[slice]` here directly and then use it wherever relevant below. This would enable 3d easily and also reduce confusion with flipping around the axis orders below.

micro_sam/evaluation/benchmark_datasets.py
```diff
@@ -0,0 +1,448 @@
...
+            device=device,
+            checkpoint_path=checkpoint_path,
+            ndim=ndim,
+            verbose=False,
+            **auto_seg_kwargs
+        )
+
+    prediction_paths = natsorted(glob(os.path.join(prediction_dir, "*")))
+    res = run_evaluation(gt_paths=gt_paths, prediction_paths=prediction_paths, save_path=result_path)
+    print(res)
```
constantinpape, line 300: We should create a dataframe and then save it as csv.

constantinpape (commented):
Looks good overall. Two minor points:
- We can make the data extraction work easily for 2d and 3d, see comments.
- We should define a csv output that everything is saved to for easy further analysis

micro_sam/evaluation/benchmark_datasets.py
anwai98: The results are stored at `result_path` filepath, passed to the `save_path` argument. I'll refactor the evaluation for clarity. Thanks for pointing out!

anwai98:
Okay, in principle the pipeline is ready now.

@constantinpape let me know how it looks now.

PS: There are a few major bulks I still need to take care of:
- Run both 2d and 3d segmentation for volumetric data.
- Allow user to flexibly choose between AMG / AIS.
- (Optional) Remove the cached original inputs.
- (Optional) Remove the cached preprocessed inputs.
- (Optional) Remove the stored predictions.

constantinpape:
> * Allow user to flexibly choose between AMG / AIS.

I think the best would be to enable running both AMG and AIS, and by default only run AIS.

micro_sam/evaluation/benchmark_datasets.py
```diff
@@ -0,0 +1,499 @@
...
+    for per_paths in tqdm(paths_set, total=len(paths_set), desc=f"Extracting patches for {dataset_choice}"):
+        if ndim == 2:
+            image_path, gt_path = per_paths
+            image, gt = util.load_image_data(image_path), util.load_image_data(gt_path)
+        else:
+            image_path = per_paths
+            image = util.load_image_data(image_path, DATASET_CONTAINER_KEYS[dataset_choice][0])
+            gt = util.load_image_data(image_path, DATASET_CONTAINER_KEYS[dataset_choice][1])
+
+        skip_smaller_shape = (image.shape > tile_shape)
```
constantinpape, line 230: Does this work? We need to do some elementwise comparison here, but `image.shape` and `tile_shape` are tuples. I think you need to case them to arrays and then use `.all`:
```python
skip_smaller_shape = (np.array(image.shape) >= np.array(tile_shape)).all()
```

micro_sam/evaluation/benchmark_datasets.py
constantinpape: Ok. I think we should also remove the prints of results here at the end. Otherwise the output gets confusing.

micro_sam/evaluation/benchmark_datasets.py
```diff
@@ -0,0 +1,499 @@
...
+        output_folder: Filepath to the folder where we store all the results.
+        ndim: The number of input dimensions.
+        checkpoint_path: The filepath for stored checkpoints.
+        auto_seg_kwargs: Additional arguments for automatic segmentation parameters.
+    """
+    result_path = os.path.join(output_folder, "results", "automatic_segmentation.csv")
+    prediction_dir = os.path.join(output_folder, "automatic_instance_segmentation", "inference")
+    os.makedirs(prediction_dir, exist_ok=True)
+
+    for image_path in tqdm(image_paths, desc="Run automatic segmentation"):
```
constantinpape, line 310: This is currently a bit ineffecient: we load the model again for each image.I think for running AIS that can be quite some overhead.

constantinpape: Not sure how exactly we want to address it. One option would be enabling passing multiple paths to `automatic_instance_segmentation`.

micro_sam/evaluation/benchmark_datasets.py
```diff
@@ -0,0 +1,499 @@
...
+                volume=imageio.imread(image_path),
+                ground_truth=imageio.imread(gt_path),
+                model_type=model_type,
+                checkpoint_path=checkpoint_path,
+                # save_path=os.path.join(prediction_dir, os.path.basename(image_path)),
+                device=device,
+                interactive_seg_mode=prompt_choice,
+                min_size=10,
+            )
+            print(msa)
```
constantinpape, line 396: Remove prints for finalization.

micro_sam/evaluation/benchmark_datasets.py
```diff
@@ -0,0 +1,499 @@
...
+
+        prediction_paths = natsorted(glob(os.path.join(prediction_dir, "*")))
+        res = run_evaluation(
+            gt_paths=gt_paths, prediction_paths=prediction_paths,
+            save_path=os.path.join(output_folder, "results", f"interactive_segmentation_3d_with_{prompt_choice}.csv")
+        )
+        print(res)
```
constantinpape, line 403: Remove prints for finalization.

micro_sam/evaluation/benchmark_datasets.py
anwai98: Oh yeah you are right. There's the case where the first dimension for image shape might be higher while the later matching, then this would fail. Your suggestion would work for all dimensions. Thanks for the mention!

micro_sam/evaluation/benchmark_datasets.py
anwai98: I think I'll update `automatic_instance_segmentation` so that it can consider multiple paths. I'll take care of it. Thanks for the suggestion!

anwai98: It was slightly bit more complicated to enable multiple paths in the existing function, so I refactored a few things to avoid doing redundant things (eg. calling the predictor and segmenter for every image). This part should be taken care of now.

anwai98:
This PR is now ready (tested this on DeepBacs and Lucchi).

I am now working on finalizing all open-source datasets in the preprocessing part, and with that should be finally good to go from my side.

Resolves https://github.com/computational-cell-analytics/micro-sam/issues/719

test/test_automatic_segmentation.py
```diff
@@ -65,46 +65,63 @@ def tearDown(self):
             import torch.mps
             torch.mps.empty_cache()
 
+    # get the predictor and segmenter.
+    def _get_predictor_and_segmenter(self, ais=False, is_tiled=False, amg_kwargs={}):
```
constantinpape, line 69: I think it would make sense to refactor this to `automatic_instance_segmentation.py` as a proper documented function to make it easy for users to load the predictor and segmenter.

micro_sam/automatic_segmentation.py
```diff
@@ -179,17 +166,28 @@ def _convert_argval(value):
...
+    predictor, state = util.get_sam_model(
+        model_type=args.model_type, device=device, checkpoint_path=args.checkpoint, return_state=True,
+    )
+
+    segmenter = get_amg(
+        predictor=predictor, is_tiled=args.tile_shape is not None,
+        decoder=get_decoder(
+            image_encoder=predictor.model.image_encoder, decoder_state=state["decoder_state"], device=device
+        ) if "decoder_state" in state and not args.amg else None
+    )
```
constantinpape, line 179: Use the refactored function here, see prev. comment

micro_sam/evaluation/benchmark_datasets.py
```diff
@@ -0,0 +1,727 @@
...
+        model_type=model_type, device=device, checkpoint_path=checkpoint_path, return_state=True,
+    )
+
+    segmenter = get_amg(
+        predictor=predictor,
+        is_tiled=False,
+        decoder=get_decoder(
+            predictor.model.image_encoder, state["decoder_state"], device
+        ) if "decoder_state" in state and not run_amg else None
+    )
```
constantinpape, line 426: Can also use the refactored function here.

constantinpape (commented):
I think we should refactor the loading functionality for automatic segmentation, rest looks good from my end now.

anwai98:
Hi @constantinpape,

This PR is GTG from my side now. (I took care of the minor TODOs which I procrastinated since last week). Thanks!

constantinpape (approved):

PR #737, opened 2024-10-15
Fix SSF implementation

caroteu:
@anwai98 I made some changes in the ssf method. The number of trainable parameters for the method is ~4.27M.

### DIFF
--- a/micro_sam/models/peft_sam.py
+++ b/micro_sam/models/peft_sam.py
@@ -112,10 +112,10 @@ class ScaleShiftLayer(nn.Module):
         self.layer = layer
         self.scale = nn.Parameter(torch.normal(mean=1.0, std=0.2, size=(dim,)))
         self.shift = nn.Parameter(torch.normal(mean=0.0, std=0.2, size=(dim,)))
+        layer = self
 
     def forward(self, x):
         x = self.layer(x)
-
         assert self.scale.shape == self.shift.shape
         if x.shape[-1] == self.scale.shape[0]:
             return x * self.scale + self.shift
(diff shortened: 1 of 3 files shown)


micro_sam/models/peft_sam.py
```diff
@@ -133,37 +133,25 @@ class SSFSurgery(nn.Module):
...
+            block.mlp.lin2 = ScaleShiftLayer(block.mlp.lin2, block.mlp.lin2.out_features)
+            block.norm1 = ScaleShiftLayer(block.norm1, block.norm1.normalized_shape[0])
+            block.norm2 = ScaleShiftLayer(block.norm2, block.norm2.normalized_shape[0])
 
-    def forward(self, x):
-        for layer in self.scale_shift_layers:
-            x = layer(x)
+        # If we get the embedding block, add one ScaleShiftLayer
+        elif hasattr(block, "patch_embed"):
+            block.proj = ScaleShiftLayer(block.proj, block.proj.out_channels)
```
caroteu, line 151: @anwai98 I currently add the scale shift layers manually depending on whether an attention block or embedding block is passed

micro_sam/models/peft_sam.py
```diff
@@ -133,37 +133,25 @@ class SSFSurgery(nn.Module):
...
 
-        return self.block(x)
+    def forward(self, x):
+        return x
```
caroteu, line 154: @anwai98 From my understanding the forward function never gets called, instead this class only modifies the layers and their forward functions

micro_sam/models/peft_sam.py
anwai98: Nice, looks good!

micro_sam/models/peft_sam.py
anwai98: You are right, makes sense actually.

anwai98:
Hi @caroteu,

Thanks for the PR. I checked it out, and I do see the additional parameters now. Good job!

I'll make a few checks and merge this to the original SSF PR.

anwai98:
Okay, seems like the tests are breaking because of an issue in the latest release at `timm`.

It's hopefully taken care of as a post-fix in source for now (https://github.com/huggingface/pytorch-image-models/commit/60f517c88327a64d44fb7259bb427c7658fdbb1b). Ross mentioned in an issue that he would make a release soon.

I'll merge this to the running SSF PR anyways and wait until the `timm` release is in!

cc: @constantinpape

PR #741, opened 2024-10-17
Add AdaptFormer

caroteu:
@anwai98 I added AdaptFormer to the PEFT methods. The number of trainable parameters for vit_b is 4.147552 (~4.15M).


micro_sam/models/peft_sam.py
```diff
@@ -141,6 +142,54 @@ def forward(self, x):
...
+
+        self.mlp_proj = block.mlp
+        self.n_embd = block.mlp.lin1.in_features
+
+        if alpha == 'learnable_scalar':
+            self.alpha = nn.Parameter(torch.ones(1))
+        else:
+            self.alpha = float(alpha)
+
+        self.rank = rank
```
caroteu, line 164: @anwai98 This is technically not a rank as in LoRA but it determines the size of the adapter layers. The authors of AdaptFormer refer to this as 'bottleneck middle dimension'. Since the impact of the parameter is similiar to a LoRA rank, I propose naming it rank to keep the number of different peft hyperparameter names as low as possible.

micro_sam/models/peft_sam.py
```diff
@@ -141,6 +142,54 @@ def forward(self, x):
...
+        nn.init.zeros_(self.down_proj.bias)
+        nn.init.zeros_(self.up_proj.bias)
+
+    def forward(self, x):
+        residual = x
+        mlp_output = self.mlp_proj(x)
+
+        down = self.down_proj(x)
+        down = self.non_linear_func(down)
+        # down = nn.functional.dropout(down, p=self.dropout, training=self.training)
```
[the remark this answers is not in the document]
anwai98: We can expose the dropout rate to make this flexible, but I agree that we stick to our discussed plans!

micro_sam/models/peft_sam.py
anwai98: I would keep the connotation of `rank` only limited to matrix decomposition and call this new argument as: `projection_size` or something like that (+ this is also why we have `module_kwargs` in `PEFT_Sam`, which capture all additional variables).

PS. This is however details which we can refactor later. It's important to validate these methods first.

caroteu: Ok I'll take care of it

anwai98:
Hi @constantinpape,

This method is GTG from our side as well. Let us know if you spot something.

constantinpape (approved):

PR #778, opened 2024-11-13
Set image dim=2 for the computation of image embeddings

caroteu:
@anwai98 The image dimension needs to be passed here, else datasets with (1,width,height) type images will cause an error.


anwai98:
Hi @caroteu,
We shouldn't fix the number of dimensions in the class method itself.
Also, if we have monochannel inputs, we should use the inputs as `(H x W)` (which is the expectation in `micro-sam`)

PR #782, opened 2024-11-18
Minor update to avoid caching embeddings in evaluation

anwai98:


anwai98:
I made a slight mistake previously which creates problems for precomputing embeddings for RGB images. This should be fixed now.

Issue #773, opened 2024-11-12
micro_sam.automatic_segmentation error


aloboa:
If I do:
`(sam) alobo2@alobo-ws:~$ micro_sam.automatic_segmentation -i "/media/alobo2/SP PHD U3/Islandia/Alteration/Cristina_nDann/DI17_2/DI17_2.tif" -o "/media/alobo2/SP PHD U3/Islandia/Alteration/microsam/DI17_2labelsout.tif" -e /home/alobo2/embeddings -m vit_b --tile_shape 1024 1024 --halo 256 256 -d 'cpu' --pred_iou_thresh 0.7 --stability_score_thresh 0.5 --box_nms_thresh 0.5 --min_mask_region_area 10`

I get:

```
progress: : 0it [00:00, ?it/s]
Traceback (most recent call last):
  File "/home/alobo2/miniforge3/envs/sam/bin/micro_sam.automatic_segmentation", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/alobo2/micro-sam/micro_sam/automatic_segmentation.py", line 238, in main
    automatic_instance_segmentation(
  File "/home/alobo2/micro-sam/micro_sam/automatic_segmentation.py", line 124, in automatic_instance_segmentation
    masks = segmenter.generate(**generate_kwargs)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/alobo2/miniforge3/envs/sam/lib/python3.11/site-packages/torch/utils/_contextlib.py", line 116, in decorate_context
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/home/alobo2/micro-sam/micro_sam/instance_segmentation.py", line 561, in generate
    masks = self._postprocess_masks(data, min_mask_region_area, box_nms_thresh, crop_nms_thresh, output_mode)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/alobo2/micro-sam/micro_sam/instance_segmentation.py", line 243, in _postprocess_masks
    mask_data = self._postprocess_small_regions(
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #788, opened 2024-11-19
Fix choice of automatic segmentation mode in CLI

anwai98:
This PR fixes the issue discussed at https://github.com/computational-cell-analytics/micro-sam/issues/773. I'll go ahead and merge this once the tests pass!

### DIFF
--- a/micro_sam/automatic_segmentation.py
+++ b/micro_sam/automatic_segmentation.py
@@ -122,10 +122,8 @@ def automatic_instance_segmentation(
 
         if len(masks) == 0:  # instance segmentation can have no masks, hence we just save empty labels
             if isinstance(segmenter, InstanceSegmentationWithDecoder):
-                print("AIS")
                 this_shape = segmenter._foreground.shape
             elif isinstance(segmenter, AMGBase):
-                print("AMG")
                 this_shape = segmenter._original_size
             else:
                 this_shape = image_data.shape[-2:]
--- a/micro_sam/automatic_segmentation.py
+++ b/micro_sam/automatic_segmentation.py
@@ -46,17 +46,16 @@ def get_predictor_and_segmenter(
 
     if amg is None:
         amg = "decoder_state" not in state
+
     if amg:
         decoder = None
     else:
         if "decoder_state" not in state:
-            raise RuntimeError("You have passed amg=False, but your model does not contain a segmentation decoder.")
+            raise RuntimeError("You have passed 'amg=False', but your model does not contain a segmentation decoder.")
         decoder_state = state["decoder_state"]
         decoder = get_decoder(image_encoder=predictor.model.image_encoder, decoder_state=decoder_state, device=device)
 
-    segmenter = get_amg(
-        predictor=predictor, is_tiled=is_tiled, decoder=decoder, **kwargs
-    )
+    segmenter = get_amg(predictor=predictor, is_tiled=is_tiled, decoder=decoder, **kwargs)
 
     return predictor, segmenter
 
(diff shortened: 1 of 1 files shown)


anwai98:
Okay, now we have tests for automatic segmentation CLI (for most model-level / automatic segmentation mode-level in 2d images). GTG from my side!

micro_sam/automatic_segmentation.py
```diff
@@ -230,7 +230,7 @@ def _convert_argval(value):
         model_type=args.model_type,
         checkpoint=args.checkpoint,
         device=args.device,
-        amg=args.amg,
+        amg=args.amg if args.amg else None,
```
constantinpape, line 233: This can't really be corrected. We now have the following cases:
```
args.amg = None -> we pass on None
args.amg = False -> we pass on None
args.amg = False -> we pass on None 
```
so we never pass on `False`

anwai98: The condition here is:
- if `args.amg` is False, then `amg` is set to `None`, which makes an automatic choice depending on the model's decoder state availability.
- if `args.amg` is True, them `amg` forces to ignore the decoder.

(let me know if i misunderstood something)

micro_sam/automatic_segmentation.py
```diff
@@ -226,11 +227,19 @@ def _convert_argval(value):
         parameter_args[i].lstrip("--"): _convert_argval(parameter_args[i + 1]) for i in range(0, len(parameter_args), 2)
     }
 
+    amg = None
+    if args.mode is None:
```
constantinpape, line 231: You don't need this `if` since you have `amg = None` already in the line before.
Just do `if args.mode is not None` and then what you currently have in Line 234 and following.
(Just a minor thing, but it's more readable.)

test/test_sam_annotator/test_cli.py
```diff
@@ -83,9 +85,47 @@ def test_precompute_embeddings(self):
...
+        image_data = binary_blobs(512).astype("uint8") * 255
+        imageio.imwrite(im_path, image_data)
+
+        out_path = "output.tif"
+
+        # Test AMG with default model in default mode.
+        run(
+            ["micro_sam.automatic_segmentation", "-i", im_path, "-o",
+             out_path, "-m", self.default_model_type]
+        )
```
constantinpape, line 103: For these tests: 
- We should check that the output path exists after each `run` and then remove it (so that the checks are reliable for the next function).
- Update the kwargs for all that use `amg` to speed up the tests (amg with default arguments are very slow), e,g, set `points_per_side=4`.

micro_sam/automatic_segmentation.py
anwai98: Ah yeah right. I missed that. I'll fix this. Thanks for spotting!

test/test_sam_annotator/test_cli.py
anwai98: > We should check that the output path exists after each run and then remove it (so that the checks are reliable for the next function).

Okay, yes. This one's easy. I'll take care of it.

> Update the kwargs for all that use amg to speed up the tests (amg with default arguments are very slow), e,g, set points_per_side=4.

Merging this in our current argument parsing setup via CLI is a bit tricky (because some arguments, eg. `points_per_side` are expected in `__init__` and others, eg. `stability_score_thresh`, are expected in the `generate` method, which in the automatic segmentation functionality are at different levels). Maybe let's look at it together to figure out an optimal solution.

anwai98: Now that I gave it a second look, it is not as complicated (the internal logic devised at the moment is quite detailed), but on the CLI level it isn't.

I'll add the changes now in the latest commit. (thanks for mentioning this)

constantinpape (approved):
I think this is now correct.

anwai98:
I think the CLIs should do the trick for checking for failing cases from now on! (thanks for the hints, Constantin)

constantinpape (approved):

Issue #786, opened 2024-11-18
Clear cache after interactive / automatic segmentation in annotators


anwai98:
I got some feedback that with continuous usage of `micro-sam` for a while with interactive segmentation on volumetric data, it seems like it's caching something and slows down the system (users mention that they eventually see the RAM filling up)

I would suggest we identify this junctions (the easiest would be - once segmentations are moved to "committed_objects" layer, we can do garbage collection to free up some memory).

anwai98:
This is taken care of. I'll close this!

PR #789, opened 2024-11-19
Add garbage collection after commit and clear annotations

anwai98:
This PR adds garbage collection after committing objects / clearing objects actions.

And resolves https://github.com/computational-cell-analytics/micro-sam/issues/786.

@constantinpape What do you think about this? I think this is potentially the simplest version of what we discussed. (I briefly tested 2d annotator and nothing seems to functionally break so far)


constantinpape:
I changed the target branch to `dev`. Looks good otherwise.

Issue #787, opened 2024-11-18
Add verification window for recomputing image embeddings


anwai98:
There are two cases here:
- In case the user has precomputed embeddings, and by mistake clicks it, we should prompt a window indicating that the embeddings have been computed and initialized.
- In case the user has precomputed embeddings, annotated stuff and choses another model to click on compute embeddings, we should prompt a window indicating that the embeddings will be computed and all the segmentation layers will be cleared.

Ideally, we should address both (in theory, point one is a subset of point two)

constantinpape:
This is done

PR #790, opened 2024-11-19
Add validation window for avoiding recomputing embeddings

anwai98:
This PR addresses to take care of misclicks over 'compute embeddings' button, by checking whether or not the embeddings have been precomputed and loaded as well.

I am missing something:
- When one chooses to cancel recomputing the image embeddings, all the objects in all other layers goes away anyways (maybe we are reinitializing the state after the button triggers the action and moves to `__call__`) (I'll have a closer look at this)

@constantinpape Does the implementation look as expected? (I tested it via 2d annotator and does the job as expected - besides the aforementioned point)

If this works, should resolve https://github.com/computational-cell-analytics/micro-sam/issues/787.


anwai98:
Hi @constantinpape,

This looks like it's working as expected (I tested it on 2d annotator). Let me know how it looks! (thanks for the suggestions)

anwai98:
Btw, the CIs for macos are failing for the embedding widget: `test/test_sam_annotator/test_widgets.py::test_embedding_widget` (I'll double check what's going on)

anwai98:
Okay, this PR still works with the latest changes from `dev` in.
I will see if the CIs work for Mac now or try to check them out. Besides that, this is GTG from my side!

anwai98:
With the latest fix, the CIs pass locally for me on my Mac. Let's see if the CIs are fixed here as well!

anwai98:
So the failing tests pass now for the annotators, which is nice. But it is just stuck at the embedding widget test (not sure why). And this test works locally on my Mac, so idk what's going on!

cc: @constantinpape

micro_sam/sam_annotator/_widgets.py
```diff
@@ -1214,8 +1224,13 @@ def __call__(self, skip_validate=False):
         image = self.image_selection.get_value()
 
         # Update the image embeddings:
-        # Reset the state.
         state = AnnotatorState()
+        if self._validate_existing_embeddings(state):
+            state._embeddings_are_same = True  # Whether embeddings already exist to control existing objects in layers.
```
constantinpape, line 1229: I think that this name is confusing. You are actually asking to skip the embedding computation, not checking if they are the same.

anwai98: I agree. As discussed, I changed the name to it's action now. Thanks!

anwai98:
Aaaaaaaaaaaand the tests pass now. Yayy! 🥳

constantinpape (approved):

Issue #792, opened 2024-11-21
Add option to choose AMG as automatic segmentation in annotator


anwai98:
Our `micro-sam` annotator prioritizes AIS as the preferred automatic segmentation mode for models finetuned with additional IS decoder. There could be use-cases, where one might prefer AMG over AIS for automatic segmentation. Can we explore the possibility of enabling users to choose their automatic segmentation mode of choice via the annotators?

PS. I am having a discussion on these lines on image.sc: https://forum.image.sc/t/applying-fine-tuned-model-with-cli/105317/2

cc: @constantinpape

anwai98:
I think this (https://github.com/computational-cell-analytics/micro-sam/pull/793) aligns with our CLI changes we have made already, right @constantinpape? Should we merge this already?

PR #793, opened 2024-11-21
Add menu option for automatic segmentation modes

anwai98:
Hi @constantinpape,

I have a proposal (we can choose to not go ahead with this as well). I think allowing the users to choose their automatic segmentation console of choice explicitly would be a nice addition (this could be done previously by switching on and off the `Prefer Segmentation Decoder` check menu, however I think it's simpler to allow it as a drop-down choice of modes.

The only thing I am worried about is confusing users for the case where `automatic_segmentation_mode: "ais"` for default models, which at the moment will still work (in the background for AMG).

Let me know what your thoughts are!

PS. Either ways, this would close https://github.com/computational-cell-analytics/micro-sam/issues/792.

### DIFF
--- a/micro_sam/sam_annotator/_widgets.py
+++ b/micro_sam/sam_annotator/_widgets.py
@@ -976,13 +976,13 @@ class EmbeddingWidget(_WidgetBase):
         )
         setting_values.layout().addLayout(layout)
 
-        # Create UI for prefering the decoder.
-        self.prefer_decoder = True
-        widget = self._add_boolean_param(
-            "prefer_decoder", self.prefer_decoder, title="Prefer Segmentation Decoder",
-            tooltip=get_tooltip("embedding", "prefer_decoder")
+        # Create UI for the choice of automatic segmentation mode.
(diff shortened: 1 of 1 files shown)


micro_sam/sam_annotator/_widgets.py
```diff
@@ -1106,10 +1107,16 @@ def pbar_init(total, description):
                 pbar_signals.pbar_total.emit(total)
                 pbar_signals.pbar_description.emit(description)
 
+            # Whether to prefer decoder.
+            # With 'amg', it is set to 'False', else it is 'True' for the default 'auto' and 'ais' mode.
+            prefer_decoder = True
+            if self.automatic_segmentation_mode in "amg":
```
constantinpape, line 1113: This is a bit unclear to me. Shouldn't it be `if self.automatic_segmentation_mode == "amg"`?

anwai98: This is fixed now. Thanks!

constantinpape (approved):

PR #814, opened 2024-12-20
Update conda environments and CI

constantinpape:


constantinpape:
Hi @hmaarrfk ,
it looks like the installation of pytorch from `conda-forge` works in principle, but the pins you specified in https://github.com/computational-cell-analytics/micro-sam/issues/435#issuecomment-2556897852 and https://github.com/computational-cell-analytics/micro-sam/issues/435#issuecomment-2556898373 don't seem to work in the recipe files.

For now, I used `pytorch-cpu` and `pytorch-gpu` , but you said that you wouldn't use those.
Could you please elaborate a bit on:
- What is the difference between the `pytorch`, `pytorch-gpu` and `pytorch-cpu`  packages (from conda-forge).
- If I just specify `pytorch`, how would conda know wether to install the CUDA or CPU build (unless this is specified by the pins).
     - The reason I ask: if this works, we wouldn't need to have separate environment files for gpu and cpu. But to my (very limited) understanding it isn't possible to check for conda/mamba to determine at runtime if CUDA is available or not.
- How would I write the pins in an environment file?

Thanks!

environment_cpu.yaml
```diff
@@ -1,22 +1,20 @@
...
     - pip
     - pooch
-    - protobuf <5
     - pyqt
     - python-xxhash
     - python-elf >=0.4.8
-    - pytorch >=2.4
+    # This does not work for me.
+    # - "pytorch=*=cpu"
+    - pytorch-cpu
```
hmaarrfk, line 17:
```suggestion
    - "pytorch=*=cpu*"
    # - pytorch-cpu
```
you need the * on the cpu to catch the rest of the build number
image

environment_gpu.yaml
```diff
@@ -1,23 +1,20 @@
...
     - pooch
-    - protobuf <5
     - pyqt
     - python-xxhash
     - python-elf >=0.4.8
-    - pytorch >=2.4
-    - pytorch-cuda>=11.7  # you may need to update the cuda version to match your system
+    # This does not work for me.
+    # - pytorch >=2.4 "libtorch=*=cuda*"
+    - pytorch-gpu
```
hmaarrfk, line 17:
```suggestion
    - pytorch >=2.4 
    - libtorch=*=cuda*
    # - pytorch-gpu
```

libtorch is a dependent package of pytorch. They are designed to match cpu/cpu cuda/cuda.

it is difficult (impossible maybe????) to write a constraint that says "greater than this version, with this build number string". 

So I use a trick, where one package specifies the build version, the other specifies the build string I want.

hmaarrfk (commented):
> What is the difference between the pytorch, pytorch-gpu and pytorch-cpu packages (from conda-forge).

See: https://github.com/conda-forge/pytorch-cpu-feedstock/issues/71#issuecomment-955608532

From "build 6" https://github.com/conda-forge/pytorch-cpu-feedstock/blob/b498f465ef7ecd80c53041f914cd82dca85a2c91/recipe/meta.yaml#L329

`pytorch-gpu` is just a shortcut for `pytorch=*=cuda*`


> If I just specify pytorch, how would conda know wether to install the CUDA or CPU build (unless this is specified by the pins).

Type
```
conda info
```

on your GPU machine. You will see something like
```
__cuda=12.4=0
```
So conda knows you have cuda ;).
I hope you don't have cuda 11.8..... god forbid 11.2. We kinda lost steam in building 11.8 packages and dropped it a few months back. But I think we have a few builds for cuda 11.8 for pytorch 2.5.1
image

However, I think that if you are running Cuda 11.8 maybe you should update to Ubuntu 24.04???? Especially for a GUI software...

In short, I work very hard such that on a "modern" system, say Ubuntu 22.04 or 24.04, the two supported LTS', one typing
```
mamba install pytorch
```
will get the "best performing pytorch" for their system.

See https://github.com/conda-forge/pytorch-cpu-feedstock/issues/102

> The reason I ask: if this works, we wouldn't need to have separate environment files for gpu and cpu. But to my (very limited) understanding it isn't possible for conda/mamba to determine at runtime if CUDA is available or not.

Correct, this is one of the big advantages of going all in on conda-forge. Just tell users to install pytorch. It will be fine ;). With the exception of all the caveats!!!

> How would I write those pins in an environment file? (If it is actually necessary)

Not necessary, but see review comments.

They could help for users that have CUDA 11.8.... because for that one, it may not be compatible with other packages that have higher version numbers.

So CPU + New Version of OTHER packages, will take precedence of pytorch with GPU.
And you might want to prefer GPU Cuda instead of "the latest Qt"

constantinpape:
Thank you for the explanation @hmaarrfk !

> In short, I work very hard such that on a "modern" system, say Ubuntu 22.04 or 24.04, the two supported LTS', one typing
> 
> ```
> mamba install pytorch
> ```
> 
> will get the "best performing pytorch" for their system.

That is amazing and will make installation a lot easier in the future. I also checked, and on our GPU server we get the correct output from `mamba info`: `__cuda=12.4=0`.

I will go ahead and simplify the environments here! For older versions where this does not work we will just add a note in the installation instructions.

constantinpape:
This seems to all work as expected. I will update the installation instructions here later and then merge it

constantinpape:
I have now also updated the installation instructions, to describe the new way to install it - using only `conda-forge` dependencies on Mac OS / Linux, and pinned dependencies on Windows. I have also updated the doc to using conda instead of mamba (since conda now uses libmamba by default we don't need to enforce using mamba any more, which was confusing for some users).

cc @anwai98

PR #818, opened 2024-12-21
Reactivate training test

### LINKED ISSUE #751: Don't skip training test
The training test is very slow since yesterday. I am skipping it for now, see:
https://github.com/computational-cell-analytics/micro-sam/blob/master/test/test_training.py#L13-L14

But we should figure out why it's slower and reactivate it so that we notice if some change affects training.

I am not sure what the reason is for this slowdown. Nothing has changed in the training code, it seems to happen irrespective of the pytorch version, so is also not just for pytorch 2.5, and everything still works normally for me locally.

constantinpape:
Fixes #751


constantinpape:
Looks like the issue still persists.

PR #821, opened 2024-12-29
Minor fix to release overview doc 

anwai98:
This PR takes care of a minor fix to release overview doc! I'll merge this once the tests pass!


anwai98:
The tests for mac are breaking (the runners are missing to find something it seems - could be random as well). I'll merge this anyways for now. We can keep an eye on the mac tests in future.

cc: @constantinpape

PR #822, opened 2024-12-30
Update LM generalist scripts

anwai98:
This PR updates scripts for LM generalists. It's still a WIP, I'll put it for review once I have some final checks done!

cc: @constantinpape


micro_sam/training/util.py
```diff
@@ -258,19 +261,20 @@ def __init__(self, desired_shape, do_rescaling=False, padding="constant"):
         self.do_rescaling = do_rescaling
 
     def __call__(self, raw):
+        raw = to_rgb(raw)  # Ensure all images are in 3-channels: triplicate one channel to three channels.
+
         if self.do_rescaling:
-            raw = normalize_percentile(raw, axis=(1, 2))
-            raw = np.mean(raw, axis=0)
+            # NOTE: Below is done for TissueNet: to work with the valid channels (i.e. the first and second channels).
```
constantinpape, line 267: This is a bit hacky. Can we properly expose this? E.g. add a parameter `valid_channels`?

micro_sam/util.py
```diff
@@ -990,7 +993,12 @@ def segmentation_to_one_hot(segmentation: np.ndarray, segmentation_ids: Optional
         n_ids = int(segmentation.max())
 
     else:
-        assert segmentation_ids[0] != 0, "No objects were found."
+        msg = "No foreground objects were found."
+        if len(segmentation_ids) == 0:  # The list should not be completely empty.
+            raise AssertionError(msg)
```
constantinpape, line 998: I would raise a different error type.

micro_sam/util.py
```diff
@@ -990,7 +993,12 @@ def segmentation_to_one_hot(segmentation: np.ndarray, segmentation_ids: Optional
...
+
+        if segmentation_ids[0] == 0:  # The list should not have zero-only.
```
constantinpape, line 1000: I don't understand the logic here. Is zero allowed at all?

anwai98: Thanks for the feedback. I've updated this in a latest commit!

anwai98:
I suggest we leave this PR open and merge this only once the evaluations are done from my side. I'll ping here once this is ready. I'll convert this to a draft now!

anwai98:
Hi @constantinpape,

This PR is GTG from my side. I made a spontaneous decision of not going ahead with adding support for allowing training from scratch (as discussed from my preliminary experiments). We can add supports for it in future once we have a better idea of what's going on. Thanks!

anwai98:
I was talking to Martin yesterday about transformers, and something came to my mind. What if the `vit-t` training from scratch did not work for us at all because of the additional adapter modules MobileSAM has? (I think it would be worth a shot to try training `vit-b` from scratch). Does this make sense @constantinpape?

constantinpape:
> What if the `vit-t` training from scratch did not work for us at all because of the additional adapter modules MobileSAM has?

I didn't know they had additional adapters.

Giving it a try with `vit_b` def. would make sense.

anwai98:
This is why we have to install MobileSAM exclusively for `vit-t` models, as they make some changes to the backbone (I don't remember the details there, but I agree with going in the direction on trying `vit-b`) (we can follow up on this later). Thanks!

PR #823, opened 2024-12-30
Make minor code semantics changes in core module

anwai98:
This PR takes care of minor updates to docstrings and code semantics. GTG from my side!

### DIFF
--- a/micro_sam/models/sam_3d_wrapper.py
+++ b/micro_sam/models/sam_3d_wrapper.py
@@ -73,6 +73,12 @@ class Sam3DWrapper(nn.Module):
                 'image': The image as a torch tensor in 3xDxHxW format. Already transformed for the input to the model.
                 'original_size': The original size of the image (HxW) before transformation.
             multimask_output: Wheterh to predict with the multi- or single-mask head of the maks decoder.
+
+        Returns:
+            A list over input images, where each element is as dictionary with the following keys:
+                'masks': Mask prediction for this object.
+                'iou_predictions': IOU score prediction for this object.
+                'low_res_masks': Low resolution mask prediction for this object.
         """
         batched_images = torch.stack([inp["image"] for inp in batched_input], dim=0)
         original_size = batched_input[0]["original_size"]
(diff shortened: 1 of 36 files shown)


micro_sam/models/sam_3d_wrapper.py
```diff
@@ -75,12 +73,6 @@ def forward(
                 'image': The image as a torch tensor in 3xDxHxW format. Already transformed for the input to the model.
                 'original_size': The original size of the image (HxW) before transformation.
             multimask_output: Wheterh to predict with the multi- or single-mask head of the maks decoder.
-
-        Returns:
```
constantinpape, line 79: Shouldn't we leave a `Returns` in the doc string here?

micro_sam/models/simple_sam_3d_wrapper.py
```diff
@@ -154,10 +154,6 @@ def forward(
             batched_input: A list over input images, each a dictionary with the following keys.L
                 'image': The image as a torch tensor in 3xDxHxW format. Already transformed for the input to the model.
             multimask_output: Wheterh to predict with the multi- or single-mask head of the maks decoder.
-
-        Returns:
```
constantinpape, line 158: Shouldn't we leave a `Returns` here?

micro_sam/training/trainable_sam.py
```diff
@@ -67,9 +69,6 @@ def forward(
             batched_inputs: The batched input images and prompts.
             image_embeddings: The precompute image embeddings. If not passed then they will be computed.
             multimask_output: Whether to predict mutiple or just a single mask.
-
-        Returns:
```
constantinpape, line 71: Shouldn't we leave a `Returns` here?

constantinpape (commented):
Looks good overall. The only thing I noticed is that you're removing `Returns` in some doc-strings where I think they should be kept.

anwai98:
I thought of removing the return types for `nn.Module` as it kind of does not make a clear sense in terms of the forward pass. However, I see the value in keeping them as the cases above are a bit more complicated and document things well. I'll restore them!

micro_sam/training/trainable_sam.py
anwai98: This is quite important. I missed it. Thanks!

anwai98:
This is GTG from my side!

cc: @constantinpape

constantinpape (approved):

PR #825, opened 2025-01-05
Minor update to prompt-based segmentation

anwai98:
This PR takes care of a corner case where we avoid extracting prompts from labels where there are no masks. I encountered this in 3d interactive segmentation. And this comes across only in places where even the initial prompt does a pretty bad job and we try to extract the first round of prompts (eg. starting with point prompts with MedSAM).

GTG from my side!

### DIFF
--- a/micro_sam/prompt_based_segmentation.py
+++ b/micro_sam/prompt_based_segmentation.py
@@ -383,7 +383,7 @@ def segment_from_mask(
             raise ValueError("If points are passed you also need to pass labels.")
         point_coords, point_labels = points, labels
 
-    elif use_points and len(np.unique(mask)) > 1:
+    elif use_points and mask.sum() != 0:
         point_coords, point_labels = _compute_points_from_mask(
             mask, original_size=original_size, box_extension=box_extension,
             use_single_point=use_single_point,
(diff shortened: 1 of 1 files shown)


constantinpape (approved):

micro_sam/prompt_based_segmentation.py
```diff
@@ -383,7 +383,7 @@ def _to_tile(prompts, shape, tile_shape, halo):
             raise ValueError("If points are passed you also need to pass labels.")
         point_coords, point_labels = points, labels
 
-    elif use_points:
+    elif use_points and len(np.unique(mask)) > 1:
```
constantinpape, line 386: You are trying to check that the mask is not empty, right? In that case it would be better to check `mask.sum() != 0`. That is much more efficient.

micro_sam/prompt_based_segmentation.py
```diff
@@ -395,7 +395,7 @@ def _to_tile(prompts, shape, tile_shape, halo):
     if box is None:
         box = _compute_box_from_mask(
             mask, original_size=original_size, box_extension=box_extension
-        ) if use_box else None
+        ) if use_box and len(np.unique(mask)) > 1 else None
```
constantinpape, line 398: See above.

micro_sam/prompt_based_segmentation.py
anwai98: Ah I see. Thanks for mentioning! I've updated it in the latest commit!

constantinpape (approved):

PR #828, opened 2025-01-11
Move quantization after LoRA surgery

caroteu:
@anwai98 I noticed that the way we we doing quantization, we did not quantize the LoRA matrices (A and B matrices). Was that intended? If not this should be fixed by adding the quantization after doing the LoRA surgery.

I also added code to initialize the quantized model for inference.

### DIFF
--- a/micro_sam/util.py
+++ b/micro_sam/util.py
@@ -375,9 +375,46 @@ def get_sam_model(
         if abbreviated_model_type == "vit_t":
             raise ValueError("'micro-sam' does not support parameter efficient finetuning for 'mobile-sam'.")
 
+        _quantize = peft_kwargs.pop("quantize", None)
+
         sam = custom_models.peft_sam.PEFT_Sam(sam, **peft_kwargs).sam
 
+        if _quantize:
+            import bitsandbytes as bnb
+            for name, module in sam.image_encoder.named_modules():
+                if isinstance(module, torch.nn.Linear):
+                    *parent_path, layer_name = name.split(".")
+                    parent_module = sam.image_encoder
+
+                    for sub_module in parent_path:
+                        parent_module = getattr(parent_module, sub_module)
+
+                    # Extract weight and bias from the state_dict
+                    weight_data = model_state.pop(f"image_encoder.{'.'.join(parent_path)}.{layer_name}.weight")
+                    bias_data = model_state.pop(f"image_encoder.{'.'.join(parent_path)}.{layer_name}.bias", None)
+
+                    layer_state_dict = {
+                        k.split(f"image_encoder.{'.'.join(parent_path)}.")[1]: v
+                        for k, v in model_state.items() if k.startswith(f"image_encoder.{'.'.join(parent_path)}.{layer_name}")
+                    }
+
+                    # Recreate the Linear4bit layer and load weights
+                    linear_q4bit = bnb.nn.Linear4bit(
+                        module.in_features,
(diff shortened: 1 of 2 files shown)


micro_sam/util.py
```diff
@@ -375,9 +375,46 @@ def get_sam_model(
         if abbreviated_model_type == "vit_t":
             raise ValueError("'micro-sam' does not support parameter efficient finetuning for 'mobile-sam'.")
 
+        _quantize = peft_kwargs.pop("quantize", None)
+
         sam = custom_models.peft_sam.PEFT_Sam(sam, **peft_kwargs).sam
 
+        if _quantize:
```
caroteu, line 382: The initialisation is inspired by https://github.com/bitsandbytes-foundation/bitsandbytes/blob/main/tests/test_linear4bit.py#L52C1-L70C56

micro_sam/util.py
```diff
@@ -375,9 +375,46 @@ def get_sam_model(
...
+                        bias=True,
+                    )
+
+                    # Assign the quantized weights to the new layer
+                    linear_q4bit.weight = bnb.nn.Params4bit.from_prequantized(quantized_stats=layer_state_dict, data=weight_data)
+                    if bias_data is not None:
+                        linear_q4bit.bias = torch.nn.Parameter(bias_data)
+
+                    # Replace the original linear layer with the quantized one
+                    setattr(parent_module, layer_name, linear_q4bit)
```
caroteu, line 414: inspired by https://github.com/bitsandbytes-foundation/bitsandbytes/blob/main/tests/test_linear4bit.py#L52C1-L70C56

anwai98:
Hi @caroteu,

>  we did not quantize the LoRA matrices (A and B matrices). 

I see. I checked it out, and I realized there were some minor details I missed as well (eg. quantizing only the linear layers for A and B matrices). 

With the latest updates, the training with mixed precision does not work anymore for me! (let me know if that's not the case for you)

And regarding the inference loading scripts, we should probably discuss it on Monday. I am a bit skeptic about the current additions!

anwai98:
Okay. I think I figured it out. The issue is that the module layers for forward and backward pass should align with each other. And the safest way to do this is by taking care of using the right modules in the `__init__` section of the 'nn' modules.

And thanks to you, we have a remarkable improvement in memory efficiency now! 🎉 

I'll push my commits after confirming a few epochs!

PS. We should still discuss the inference changes you made because I think we don't need it. As mentioned, let's talk about the details on Monday!

finetuning/livecell_finetuning.py
```diff
@@ -56,18 +56,19 @@ def finetune_livecell(args):
...
 
     # vit_h
-    # freeze_encoder: ~36.05 GB
-    # QLoRA: ~ 65.68 GB
-    # LoRA: ~ 67.14 GB
-    # FFT: ~72.34 GB
+    # freeze_encoder: ~36.33 GB
+    # QLoRA: ~ 36.41 GB
+    # LoRA: ~ 67.52 GB
+    # FFT: ~72.79 GB
```
anwai98, line 71: And here goes the impressive improvement with QLoRA. Now we see it's true potential!

cc: @caroteu @constantinpape

constantinpape: Very cool!

anwai98:
I'll drop this, since we clarified a few confusions. We will work on finalizing QLoRA in another PR!

PR #829, opened 2025-01-11
Extend dice score metric for multi-dim segmentation

anwai98:
This PR takes care of a minor update for id-wise (or rather, class-wise) dice score evaluation for multi-dim segmentation results. I'll merge this once the tests pass!

### DIFF
--- a/micro_sam/evaluation/multi_dimensional_segmentation.py
+++ b/micro_sam/evaluation/multi_dimensional_segmentation.py
@@ -215,9 +215,23 @@ def segment_slices_from_ground_truth(
             segmentation=final_segmentation, groundtruth=curr_gt, return_accuracies=True
         )
         results = {"mSA": msa, "SA50": sa[0], "SA75": sa[5]}
-    elif evaluation_metric == "dice":
-        dice = dice_score(segmentation=final_segmentation, groundtruth=curr_gt)
+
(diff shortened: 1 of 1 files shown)


micro_sam/evaluation/multi_dimensional_segmentation.py
```diff
@@ -215,9 +215,23 @@ def segment_slices_from_ground_truth(
             segmentation=final_segmentation, groundtruth=curr_gt, return_accuracies=True
         )
         results = {"mSA": msa, "SA50": sa[0], "SA75": sa[5]}
-    elif evaluation_metric == "dice":
-        dice = dice_score(segmentation=final_segmentation, groundtruth=curr_gt)
+
+    elif evaluation_metric.startswith("dice"):
```
constantinpape, line 219: The if-else is too complicated. The following would work and doesn't require an additional level:
```python
elif evaluation_metric == "dice":
    ...
elif evaluation_metric == "dice_per_class":
   ...
else:
  raise ValueError("... Please choose one of 'sa'', 'dice' or 'dice_per_class'")
```

anwai98: Yeah, now that I think about it, it's wayyy too complicated! I'll create another PR to simplify this!

PR #846, opened 2025-02-04
Minor update to make rank optional in peft class

anwai98:
This PR takes care of a minor update to make rank optional (as it is not important for other selective peft methods). I'll wait for checks from @caroteu and we merge this then!


caroteu:
Looks good to me!

Issue #729, opened 2024-10-09
Extend automatic segmentation CLI to continue annotation


anwai98:
I came across a random thought: would it be interesting to integrate annotators to automatic segmentation CLI?

This would go in the direction of an intuitive segmentation workflow, where: user utilizes the CLI to perform automatic segmentation, and can move ahead with parsing an argument stating `--annotator`, which will activate the napari GUI, load the automatic segmentation to the `committed_objects` and a user then has the choice of improving segmentation using interactive mode.

PS. I also feel it might be an overkill, it's just a thought though (only makes sense if it's beneficial for a use-case). 

cc: @constantinpape

constantinpape:
I think that's a nice idea, and not much overhead. Before this we should make sure the `segmentation_result` argument works correctly in all annotators, as this would be how to implement this.

anwai98:
This is taken care of as well. I'll close this!

PR #858, opened 2025-02-07
Extend automatic segmentation CLI to allow continuing annotation

anwai98:
This PR closes https://github.com/computational-cell-analytics/micro-sam/issues/729, to (optionally) support continuing annotation / view automatic segmentations after automatic segmentation CLI process is completed. GTG from my side!

### DIFF
--- a/micro_sam/automatic_segmentation.py
+++ b/micro_sam/automatic_segmentation.py
@@ -75,6 +75,8 @@ def automatic_instance_segmentation(
     halo: Optional[Tuple[int, int]] = None,
     verbose: bool = True,
     return_embeddings: bool = False,
+    annotator: bool = False,
+    view: bool = False,
     **generate_kwargs
 ) -> np.ndarray:
     """Run automatic segmentation for the input image.
@@ -94,6 +96,8 @@ def automatic_instance_segmentation(
         halo: Overlap of the tiles for tiled prediction.
         verbose: Verbosity flag.
         return_embeddings: Whether to return the precomputed image embeddings.
+        annotator: Whether to activate the annotator for continue annotation process.
+        view: Whether to visualize the segmentations corresponding to the original input image.
         generate_kwargs: optional keyword arguments for the generate function of the AMG or AIS class.
 
     Returns:
@@ -140,6 +144,17 @@ def automatic_instance_segmentation(
             # if (raw) predictions provided, store them as it is w/o further post-processing.
             instances = masks
 
+        if annotator:  # Allow opening the automatic segmentation in the annotator for further annotation, if desired.
+            from micro_sam.sam_annotator import annotator_2d
+            annotator_2d(
+                image=image_data,
+                model_type=predictor.model_name,
+                embedding_path=embedding_path,
+                segmentation_result=instances,  # Initializes the automatic segmentation to the annotator.
+                tile_shape=tile_shape,
+                halo=halo,
+            )
+
     else:
         if (image_data.ndim != 3) and (image_data.ndim != 4 and image_data.shape[-1] != 3):
             raise ValueError(f"The inputs does not match the shape expectation of 3d inputs: {image_data.shape}")
@@ -161,10 +176,29 @@ def automatic_instance_segmentation(
         else:
             instances = outputs
 
+        if annotator:  # Allow opening the automatic segmentation in the annotator for further annotation, if desired.
+            from micro_sam.sam_annotator import annotator_3d
+            annotator_3d(
+                image=image_data,
+                model_type=predictor.model_name,
+                embedding_path=embedding_path,
+                segmentation_result=instances,  # Initializes the automatic segmentation to the annotator.
+                tile_shape=tile_shape,
+                halp=halo,
+            )
+
     # Save the instance segmentation, if 'output_path' provided.
     if output_path is not None:
         output_path = Path(output_path).with_suffix(".tif")
         imageio.imwrite(output_path, instances, compression="zlib")
+
+    # Whether to visualize the input image and corresponding segmentation.
+    if view:
+        import napari
+        v = napari.Viewer()
+        v.add_image(image_data, name="Image")
+        v.add_labels(instances.astype(int), name="Instance Segmentation")
+        napari.run()
 
     if return_embeddings:
         return instances, image_embeddings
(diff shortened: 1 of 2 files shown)


micro_sam/automatic_segmentation.py
```diff
@@ -75,6 +75,8 @@ def automatic_instance_segmentation(
     halo: Optional[Tuple[int, int]] = None,
     verbose: bool = True,
     return_embeddings: bool = False,
+    annotator: bool = False,
+    view: bool = False,
```
constantinpape, line 79: I am not sure if I would add both `view` and `annotator`. If we do, I would use a consistent naming and call it `annotate`.

micro_sam/automatic_segmentation.py
```diff
@@ -161,11 +176,30 @@ def automatic_instance_segmentation(
...
+        if annotator:  # Allow opening the automatic segmentation in the annotator for further annotation, if desired.
+            from micro_sam.sam_annotator import annotator_3d
+            annotator_3d(
+                image=image_data,
+                model_type=predictor.model_name,
+                embedding_path=embedding_path,
+                segmentation_result=instances,  # Initializes the automatic segmentation to the annotator.
+                tile_shape=tile_shape,
+                halp=halo,
+            )
```
constantinpape, line 188: The combination of `annotator=True` and `output_path is not None` yields to a bit counter intuitive behavior:
the viewer opens, the user can annotate the image, but after closing it the original instance segmentation get saved and not the version corrected by the user.

constantinpape: It's probably easy to updated `annotator_2d`, `annotator_3d` (and `annotator_tracking`) so that they return the segmentation after annotation to fix this.

micro_sam/automatic_segmentation.py
anwai98: Now that I think about it, I agree with your first suggestions. It's unnecessary to have `view`, when `annotator` can just do the same -> user can just close the annotator -> and the results are saved anyways as desired (goes in the direction of the supports which I'll extend now to get `committed_objects` from the `viewer`). Thanks for spotting!

micro_sam/automatic_segmentation.py
anwai98: For the CLI, it's mandatory for a user to provide `output_path` to store their segmentations. Now, we adapt it in such a way that the `committed_objects` is the final target which is saved.

micro_sam/sam_annotator/annotator_2d.py
```diff
@@ -98,11 +98,11 @@ def annotator_2d(
         device=device, tile_shape=tile_shape, halo=halo
     )
 
+    napari.run()
+
     if return_viewer:
         return viewer
```
anwai98, line 104: I had to take care of this update such that the behaviour is as follows:

```python
# OLD
if return_viewer:
    return viewer

napari.run()

# NEW
napari.run()

if return_viewer:
    return viewer
```

With the old version, the viewer just did not open, and does not issue any errors either. Whereas in the new setup, once the annotation process ends, the viewer is returned and the desired parts from the layers can be extracted.

I checked and this is a feature which is used in image series annotator and CIs for 2d, 3d and tracking annotators. Let me know if this changes behavior for some other functionality.

cc: @constantinpape

micro_sam/sam_annotator/annotator_2d.py
```diff
@@ -98,11 +98,11 @@ def annotator_2d(
         device=device, tile_shape=tile_shape, halo=halo
     )
 
+    napari.run()
```
constantinpape, line 101: The viewer is not started on purpose if `return_viewer=True`. Your changes would not be compatible with the image series annotator. There is no reason to start the viewer here, you can just do it in the code that calls the annotator.

micro_sam/sam_annotator/annotator_2d.py
constantinpape: Please change this back, this will not be compatible with the image series annotator.

micro_sam/automatic_segmentation.py
```diff
@@ -140,6 +142,22 @@ def automatic_instance_segmentation(
             # if (raw) predictions provided, store them as it is w/o further post-processing.
             instances = masks
 
+        if annotate:  # Allow opening the automatic segmentation in the annotator for further annotation, if desired.
```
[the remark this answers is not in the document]
anwai98: Ah yeah right. That reduces a lot of lines. Thanks!

micro_sam/automatic_segmentation.py
```diff
@@ -140,6 +142,22 @@ def automatic_instance_segmentation(
...
+            from micro_sam.sam_annotator import annotator_2d
+            viewer = annotator_2d(
+                image=image_data,
+                model_type=predictor.model_name,
+                embedding_path=embedding_path,
+                segmentation_result=instances,  # Initializes the automatic segmentation to the annotator.
+                tile_shape=tile_shape,
+                halo=halo,
+                return_viewer=True,  # Returns the viewer, which allows the user to store the updated segmentations.
+            )
```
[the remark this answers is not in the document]
anwai98: That was the missing piece in the puzzle. It works as expected now. Thanks for the hint!

micro_sam/automatic_segmentation.py
```diff
@@ -161,6 +163,30 @@ def automatic_instance_segmentation(
...
+        )
+
+        # Start the GUI here
+        import napari
+        napari.run()
+
+        # We extract the segmentation in "committed_objects" layer, where the user either:
+        # a) Performed interactive segmentation / corrections and committed them, OR
+        # b) Did not do anything and closed the annotator, i.e. keeps the segmentations as it is.
+        instances = viewer.layers["committed_objects"].data
```
constantinpape, line 188: Have you tested once that this actually works? I am never sure about the actual state logic of napari.

constantinpape (commented):
The code looks good now! Have you checked that the instance segmentation is fetched from the viewer correctly though?

micro_sam/automatic_segmentation.py
anwai98: Yes, I tried this out, by removing labels, adding segmentations and closing the annotator. It works for all cases!

constantinpape (approved):

Issue #765, opened 2024-10-27
Extend support for RGB images in training GUI


anwai98:
This should be an easy fix in `default_sam_loader` under `TrainingWidget` (now that we support all arguments from `torch_em.default_segmentation_dataset` (currently since RGB images would inherently have channels last, the segmentation dataset throws an error).

I'll take care of it this week!

anwai98:
This is taken care of as well. I'll close this!

PR #859, opened 2025-02-07
Extend support for GUI training on RGB images

anwai98:
Closes https://github.com/computational-cell-analytics/micro-sam/issues/765.

This PR makes our backbone a bit more flexible to adapt to RGB images (it did not work in it's current state for several reasons), the most important one being the segmentation dataset being chosen for RGB images.

TLDR fix: I make the judgement based on provided images, and allow switching to `ImageCollectionDataset` for RGB images.

GTG from my side!

### DIFF
--- a/micro_sam/training/training.py
+++ b/micro_sam/training/training.py
@@ -381,7 +381,7 @@ def default_sam_dataset(
     label_key: Optional[str],
     patch_shape: Tuple[int],
     with_segmentation_decoder: bool,
-    with_channels: bool = False,
+    with_channels: Optional[bool] = None,
     sampler: Optional[Callable] = None,
     raw_transform: Optional[Callable] = None,
     n_samples: Optional[int] = None,
@@ -403,7 +403,7 @@ def default_sam_dataset(
             or a glob pattern for selecting multiple files.
         patch_shape: The shape for training patches.
         with_segmentation_decoder: Whether to train with additional segmentation decoder.
-        with_channels: Whether the image data has RGB channels.
+        with_channels: Whether the image data has channels. By default, it makes the decision based on inputs.
         sampler: A sampler to reject batches according to a given criterion.
         raw_transform: Transformation applied to the image data.
             If not given the data will be cast to 8bit.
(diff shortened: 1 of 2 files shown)


micro_sam/training/training.py
```diff
@@ -354,13 +355,12 @@ def _update_patch_shape(patch_shape, raw_paths, raw_key, with_channels):
     if raw_key is None:  # If no key is given then we assume it's an image file.
         ndim = imageio.imread(path).ndim
     else:  # Otherwise we try to open the file from key.
-        try:  # First try to open it with elf.
-            with open_file(path, "r") as f:
-                ndim = f[raw_key].ndim
-        except ValueError:  # This may fail for images in a folder with different sizes.
-            # In that case we read one of the images.
+        if "*" in raw_key:  # In that case we read one of the images.
```
constantinpape, line 358: The code before should have been correct: if `*` is in the key then [ImageStackDataset](https://github.com/constantinpape/elf/blob/master/elf/io/image_stack_wrapper.py#L91) can open the data, unless the images are of different size. In that case we get a `ValueError`, which gets caught, and read an individual image.

anwai98: I switched this up as `ImageStackDataset` currently does not support 3 channel (RGB) images (https://github.com/constantinpape/elf/blob/master/elf/io/image_stack_wrapper.py#L99). Can you suggest something for this use-case?

anwai98: Should be fixed at https://github.com/constantinpape/elf/pull/115.

anwai98:
Hi @constantinpape,

I removed the hard-coded dimension check for `ImageStackDataset` in `elf` and restored the previous structure of checking for stacks in our training GUI for prepping dataloaders. This is GTG from my side now.

micro_sam/training/training.py
```diff
@@ -416,6 +417,24 @@ def default_sam_dataset(
         The segmentation dataset.
     """
 
+    # By default, let the 'default_segmentation_dataset' heuristic decide for itself.
+    is_seg_dataset = None
```
constantinpape, line 421: Have you checked if we actually need any of this now? In principle it should not matter since the `SegmentationDataset` can also read RGB data.

constantinpape:
I created a new elf release so that the changes to `ImageStackDataset` will be available soon: https://github.com/constantinpape/elf/releases/tag/0.6.1

micro_sam/training/training.py
anwai98: This won't work with `SegmentationDataset` still as it expects inputs with channels first: https://github.com/constantinpape/torch-em/blob/main/torch_em/data/segmentation_dataset.py#L101-L103

constantinpape: Makes sense! Thanks for clarifying.

micro_sam/training/training.py
```diff
@@ -416,6 +417,24 @@ def default_sam_dataset(
...
+
+    # Check if the raw inputs are RGB or not. If yes, use 'ImageCollectionDataset'.
+    # Get valid raw paths to make checks possible.
+    if raw_key is None:
+        rpath = raw_paths if isinstance(raw_paths, str) else raw_paths[0]
+    else:
+        rpath = glob(os.path.join(raw_paths if isinstance(raw_paths, str) else raw_paths[0], raw_key))[0]
```
constantinpape, line 428: This only makes sense if raw_key encodes a glob string. But if we have a `raw_key`  for a container file format (hdf5, zarr etc.) this will fail.

anwai98: You are right. The logic was incorrect from my side. This should be fixed now!

micro_sam/training/training.py
```diff
@@ -416,6 +417,27 @@ def default_sam_dataset(
...
+    # Load one of the raw inputs to validate whether it is RGB or not.
+    test_raw_inputs = load_data(path=rpath, key=raw_key if raw_key and "*" not in raw_key else None)
+    if test_raw_inputs.ndim == 3:
+        if test_raw_inputs.shape[-1] == 3:  # i.e. if it is an RGB image and has channels last.
+            is_seg_dataset = False
+            # Provide list of inputs to 'ImageCollectionDataset' in this case.
+            raw_paths = [raw_paths] if isinstance(raw_paths, str) else raw_paths
+            label_paths = [label_paths] if isinstance(label_paths, str) else label_paths
+        elif test_raw_inputs.shape[0] == 3:  # i.e. if it is an RGB image and has channels first.
+            with_channels = True
```
constantinpape, line 439: We should be careful with over-writing a parameter passed by the user like this. The behavior is now quite inconsistent, e.g. why not set `with_channels = True` in the previous condition? 
What about images with three slices or timepoints? They trigger `with_channels=True` although this is wrong.
I would suggest to change the default of `with_channels` to `None` and then to derive its value from the data -- only if the user did not pass an exlicit value (i.e. `True` or `False`).

anwai98: I see the problem. I clarified the docstring for `with_channels` now and make sure that it priorities the user's choice over the internal data heuristics. Let me know if it looks better now!

constantinpape (approved):

Issue #836, opened 2025-01-17
Add CLI for finetuning Segment Anything


anwai98:
I think it would be nice to have a simple CLI for finetuning SAM / `micro-sam` models (and have heard of some requests from other users as well)! What do you think @constantinpape?

PS. I will only come back to this early next month!

constantinpape:
Yes, I agree that this would be good to have.

anwai98:
Another feature request: (I've also been thinking about this for a while now)

- It would make sense to not store model checkpoints in the current working directory, and allow users to specify their directory (using `save_root`). From what I understand, this should be updated! (https://github.com/computational-cell-analytics/micro-sam/blob/master/micro_sam/sam_annotator/training_ui.py#L244-L250)
- Review the advanced parameters:
    - `Initial model`: we can fetch the value from `Configuration` and set it to default, or display a template. I've seen users get confused about it!
    - `Checkpoint`: Is there a reason to put a default value there? I would call it something like `Checkpoint path` and provide a template for a filepath. 
    - Can we support an export script which stores the SAM model state and `decoder_state` in one filepath? (this would simplify our design to export all relevant components for the finetuned model)

anwai98:
On a second thought, we can do the above mentioned engineering in training GUI later. I have some ideas on it.

For now, the training CLI is merged to `dev`. I'll close this!

PR #860, opened 2025-02-08
Add CLI for training SAM

anwai98:
This PR adds a new CLI for finetuning SAM on custom data. This is still a WIP, and can be fully tested once https://github.com/computational-cell-analytics/micro-sam/pull/859 is merged! (I will work on finalizing that first)

Once finished, this resolves https://github.com/computational-cell-analytics/micro-sam/issues/836

PS. 1) I moved a few functions around files to dodge circular imports, and 2) I made an optional change in export scripts, where the exported model checkpoints would now also have the decoder state, allowing users to run AIS with exported models.

### DIFF
--- a/micro_sam/training/training.py
+++ b/micro_sam/training/training.py
@@ -10,8 +10,9 @@ import imageio.v3 as imageio
 
 import torch
 from torch.optim import Optimizer
-from torch.optim.lr_scheduler import _LRScheduler
+from torch.utils.data import random_split
 from torch.utils.data import DataLoader, Dataset
+from torch.optim.lr_scheduler import _LRScheduler
 
 import torch_em
 from torch_em.data.datasets.util import split_kwargs
@@ -23,10 +24,10 @@ try:
 except Exception:
     QObject = Any
 
-from ..util import get_device
 from . import sam_trainer as trainers
 from ..instance_segmentation import get_unetr
 from . import joint_sam_trainer as joint_trainers
+from ..util import get_device, get_model_names, export_custom_sam_model
 from .util import get_trainable_sam_model, ConvertToSamInputs, require_8bit
 
 
@@ -426,9 +427,7 @@ def default_sam_dataset(
             foreground=True, instances=True, min_size=min_size,
         )
     else:
-        label_transform = torch_em.transform.label.MinSizeLabelTransform(
-            min_size=min_size
-        )
+        label_transform = torch_em.transform.label.MinSizeLabelTransform(min_size=min_size)
 
     # Set a default sampler if none was passed.
     if sampler is None:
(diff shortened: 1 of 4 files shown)


micro_sam/util.py
```diff
@@ -474,7 +478,18 @@ def export_custom_sam_model(
...
     )
-    torch.save(model_state, save_path)
+
+    # Store the 'decoder_state' as well, if desired.
+    if with_segmentation_decoder:
+        if "decoder_state" not in state:
+            raise RuntimeError(f"'decoder_state' is not found in the model at '{checkpoint_path}'.")
+
+        decoder_state = state["decoder_state"]
+        save_state = {"model_state": model_state, "decoder_state": decoder_state}
```
constantinpape, line 488: Have you checked that this can be correctly loaded again by micro_sam?

micro_sam/util.py
```diff
@@ -465,6 +468,7 @@ def export_custom_sam_model(
         checkpoint_path: The path to the corresponding checkpoint if not in the default model folder.
         model_type: The Segment Anything Model type corresponding to the checkpoint (vit_h, vit_b, vit_l or vit_t).
         save_path: Where to save the exported model.
+        with_segmentation_decoder: Whether to store the decoder state in the model checkpoint as well.
```
constantinpape, line 471: We should add that this means that the exported model will be incompatible with other tools using SAM.

anwai98: I have added a mention in the docstring as discussed!

micro_sam/training/training.py
```diff
@@ -589,3 +613,243 @@ def train_sam_for_configuration(
...
+    # Export the model if an output path has been given.
+    if output_path:
+
+        # If the filepath has a pytorch-specific ending, then we just export the checkpoint.
+        if os.path.splitext(output_path)[1] in (".pt", ".pth"):
+            export_custom_sam_model(
+                checkpoint_path=best_checkpoint,
+                model_type=model_type[:5],
+                save_path=output_path,
+                with_segmentation_decoder=with_segmentation_decoder,
```
anwai98, line 822: After our discussion today @constantinpape, I thought about this a bit. The recommendation now would be: if you provide a filepath (eg. `output_model.pt`), then you get a `micro-sam` suitable model with both backbones (SAM + IS) for running via `micro-sam` compatible tools. Otherwise, users should use the `bioimageio` export scripts (which I fixed a few weeks ago) to have the cross-tool compatible model!

constantinpape: I agree with setting the same flag here. I.e. if a user sets `with_segmentation_decoder` to `True` in the CLI, we should export the model with it as well.
We should actually do the same in the training plugin. (I will take care of it).

constantinpape (approved):

PR #869, opened 2025-02-19
Update win installation WIP

constantinpape:
cc @anwai98


constantinpape:
Hi @hmaarrfk ,
we found that pytorch is now on conda-forge also for windows (yay!!!) and so try to update the dependencies for the installation here. However, the installation with this dependency consistently times out, because installing it takes so long.

See
https://github.com/computational-cell-analytics/micro-sam/actions/runs/13410630608/job/37467791597?pr=869

Do you have any idea why this might be?

Thanks!

hmaarrfk:
The timeout is set at 60 mins. Increase it to 120?

constantinpape:
We could increase the time limit, but before updating the environment (when we still used pytorch from the pytorch channel) it was resolved in a few minutes.

hmaarrfk:
I’m sorry. I don’t use use windows myself. I know it is slower with conda. PyTorch and Cuda are heavy.

environment_win.yaml
```diff
@@ -1,9 +1,7 @@
 name: sam
 channels:
-    - pytorch
     - conda-forge
 dependencies:
-    - cpuonly
     # This pin is necessary because later nifty versions have import errors on windows.
     - nifty =1.2.1=*_4
```
hmaarrfk, line 6: remove this.

hmaarrfk: or at leas the "pin"

hmaarrfk: well its good we left ourselves that note such a long time ago ^_^

.github/workflows/test.yaml
```diff
@@ -31,7 +31,7 @@ jobs:
       - name: Setup micromamba
         uses: mamba-org/setup-micromamba@v1
         with:
-          environment-file: ${{ runner.os == 'Windows' && 'environment_cpu_win.yaml' || 'environment.yaml' }}
+          environment-file: ${{ runner.os == 'Windows' && 'environment_win.yaml' || 'environment.yaml' }}
```
hmaarrfk, line 34:
```suggestion
          environment-file: 'environment.yaml'
```

you might as well simplify your CI file altogether.

hmaarrfk: and your instructions!

constantinpape:
Thanks for the quick help @hmaarrfk ! We will double check that the installation works now without the nifty pin, but then this should work!

constantinpape:
I merged this now, but we need to check locally that it works. I have a spare laptop with windows now, which we can use @anwai98

anwai98:
Nice. That sounds cool. I'll take the windows laptop from you and test it out over there. Thanks!

environment_win.yaml
hmaarrfk, line 1: I’m pretty sure this file was meant to be deleted

PR #879, opened 2025-02-20
Add support for preprocessing inputs in training CLI

anwai98:
This PR adds support for preprocessing inputs for training SAM. I tested this on a data, and everything works as expected. GTG from my side!

### DIFF
--- a/micro_sam/training/util.py
+++ b/micro_sam/training/util.py
@@ -1,6 +1,7 @@
 import os
 from math import ceil, floor
-from typing import Dict, List, Optional, Union, Tuple
+from functools import partial
+from typing import Dict, List, Optional, Union, Tuple, Callable
 
 import numpy as np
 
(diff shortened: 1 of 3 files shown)


micro_sam/training/util.py
```diff
@@ -39,6 +40,40 @@ def require_8bit(x):
     return x
 
 
+def _raw_transform(image: np.ndarray, raw_trafo: Callable) -> np.ndarray:
+    return raw_trafo(image) * 255
+
+
+def _normalize_percentile(image: np.ndarray) -> np.ndarray:
+    image = normalize_percentile(image)  # Use 1st and 99th percentile values for min-max normalization.
+    image = normalize(image)  # Normalize the rescaled values for ranging them under (0, 1)
```
constantinpape, line 49: I am not sure if this is a good idea, since you are using a `min/max` after the percentile. This undoes the effect.
You can do a `clip` to `[0, 1]` here instead.

micro_sam/training/util.py
```diff
@@ -39,6 +40,40 @@ def require_8bit(x):
...
+    Args:
+        preprocess: By default, the transformation function is set to 'None'.
+            The user can choose from 'normalize_minmax' / 'normalize_percentile'.
+
+    Returns:
+        The transformation function.
+    """
+
+    if preprocess is None:
+        return None
```
constantinpape, line 65: What actually happens when we run into this code path? Is no preprocessing used, or do we run into the defaults from some other preprocessing? In the latter case, we should refactor this, so that the preprocessing for training is all handled in one place. Otherwise it becomes confusing.

constantinpape (commented):
The changes look good overall, but I would change the percentile normalization logic, and maybe refactor the normalization, so that all of this happens in one place.

micro_sam/training/util.py
anwai98: I see. When this is set to `None`, we utilize the default identity transform. I agree, it's best to return the `identity` raw transforms here to allow users to understand what's going on. I will take care of this!

anwai98: This is fixed in the latest commit now. Thanks!

micro_sam/training/util.py
constantinpape: It's fine to clip in this case, there just won't be any differences in the intensity for these values.
On the other hand, if you do a min-max norm, then you pretty much redo any effect of the percentile norm and squeeze the dynamic range if you have any outliers.

anwai98: Ah I see. So the target is to moderate the overall dynamic range. Got it. Thanks!

I'll take care of this!

anwai98: I fixed it in a latest commit now. Thanks for the detailed explanation!

constantinpape (approved):

PR #882, opened 2025-02-21
Minor fix to automatic segmentation in 3d

anwai98:
This PR takes care of a minor fix to automatic segmentation post-processing in 3d. GTG from my side!

### DIFF
--- a/micro_sam/multi_dimensional_segmentation.py
+++ b/micro_sam/multi_dimensional_segmentation.py
@@ -397,7 +397,9 @@ def automatic_3d_segmentation(
         The segmentation.
     """
     offset = 0
-    segmentation = np.zeros(volume.shape, dtype="uint32")
+    segmentation = np.zeros(
+        volume.shape if volume.ndim == 3 else volume.shape[:3], dtype="uint32"
+    )
 
     min_object_size = kwargs.pop("min_object_size", 0)
     image_embeddings = util.precompute_image_embeddings(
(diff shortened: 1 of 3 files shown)


micro_sam/multi_dimensional_segmentation.py
```diff
@@ -397,7 +397,9 @@ def automatic_3d_segmentation(
         The segmentation.
     """
     offset = 0
-    segmentation = np.zeros(volume.shape, dtype="uint32")
+    segmentation = np.zeros(
+        volume.shape if volume.ndim == 3 else volume.shape[:3], dtype="uint32"
+    )
```
constantinpape, line 402: You don't even need the `if else` here, you can just use `volume.shape[:3]`.

micro_sam/multi_dimensional_segmentation.py
```diff
@@ -416,7 +418,16 @@ def automatic_3d_segmentation(
         if len(seg) == 0:
             continue
         else:
-            seg = mask_data_to_segmentation(seg, with_background=with_background, min_object_size=min_object_size)
+            if isinstance(seg, list):
```
constantinpape, line 421: I am bit confused by this change. We already have the `if len(seg) == 0: continue` on top. So why do we need to repeat this here?

micro_sam/multi_dimensional_segmentation.py
anwai98: Ah yeah right hehe. Makes sense. Thanks for spotting!

micro_sam/multi_dimensional_segmentation.py
anwai98: You are right.

I thought about this a bit, and I simplified the post-processing logic combination. I tested it on volumetric segmentation (with and without channels), and works as expected.

Thanks for spotting this!

constantinpape (approved):

Issue #883, opened 2025-02-21
Raise informed errors for mismatching model type and model weight combination


anwai98:
Let's say a user uses `vit_l` to finetune on their data. And then opens napari, does not switch to `vit_l` model and passes the filepath to their finetuned model in our framework, the model just raises a huge incompatible key-value errors (the usual state-dict mismatch stuff)

It would be nice to make the user aware about this in either of the two ways:
- Either raise an error by informing the user that the model type and model checkpoint do not match (we can do this by checking the depth/ embedding dimension of the image encoder (I'll call this IEF to refer later) vs incoming state dict)
- Or, allow flexible compatibility of user-provided checkpoint as a priority and switch the model type internally subjected to the IEFs existing in the provided model weights, and alert the user with a validation window in GUI and raise a warning about the behavior.

I've seen this before with a lot of users, and I think this would help users understand this a bit better. What do you think about this @constantinpape?

constantinpape:
I think the second option would be the ideal solution (if it's easy to implement, I haven't thought about that part though).
In that case, we would just raise a warning if the specified model type does not fit the checkpoint, and then switch the model type.

anwai98:
The core function is supported now. I will go ahead and close this!

PR #886, opened 2025-02-22
Allow flexible loading of mismatching checkpoint-model type combination

anwai98:
This PR allows flexible loading of mismatching `model_type` and `checkpoint_path` combinations. The behavior is set to prioritize the provided model weights over the model type name. We make this more flexible and informed here for the users to understand about this behavior.

I still need to work on switching the annotator's model choice. I think the easiest would be to rely on `predictor._model_type`, as we assign the latest `model_type` to this object anyways. I will look into it.

Once finalized, it resolves https://github.com/computational-cell-analytics/micro-sam/issues/883.


anwai98:
Okay, I managed to figure out the GUI triggers via CLI. I will look into checking for updating the model choice automatically when the user uses GUI (it works functionally across all modes, I am just missing the `model_type` upload in widgets).

anwai98:
I think I figured out the spot for updating the `model_type` to sync the new embeddings, however I think I am still missing something minor (the drop-down default for choosing the model still does not change). Once you have time @constantinpape, could you take a look at this and let me know if you spot what I am missing?

constantinpape:
I had a quick look, but I think it's better if we discuss the logic and go over this together.

constantinpape:
fyi, I moved the base branch of this PR to `dev`, since we will only include it in the next release (1.4)

anwai98:
Okay, this was pretty easy to combine hehe.

I will now figure out how to make the `model` dropdown's "Default" text gone now and then should be good to go!

anwai98:
So the checkpoint passing (in an arbitrary `model_type` - `checkpoint_path` combination) works via both CLI and GUI now.

However, I made a small change in plan: I decided to change the dropdown display text to `Default` and make the `model size:` parameter flexible to the provided model checkpoint (i.e. it does change depending on the true model size subjected to checkpoint). Removing the model family to anything else or removing the text makes the GUI much more complicated imo.

What do you think about this? @constantinpape 

Otherwise, everything is GTG from my side!

constantinpape (approved):

constantinpape:
> What do you think about this? @constantinpape

Let's discuss this tomorrow. I will go ahead and merge this one anyways, as this could be done in a separate PR.
There are also a few other minor points to take care of. (E.g. update the message when pressing compute embeddings.)

PR #890, opened 2025-02-24
Update documentation with latest features

anwai98:
This PR is a WIP to document all new features from our side!

### DIFF
--- a/doc/annotation_tools.md
+++ b/doc/annotation_tools.md
@@ -133,7 +133,9 @@ Check out [the video tutorial](https://youtu.be/HqRoImdTX3c) for an in-depth exp
 
 ## Finetuning UI
 
-We also provide a graphical inferface for fine-tuning models on your own data. It can be started by clicking `Finetuning` in the plugin menu after starting `napari`.
+We also provide a graphical inferface for fine-tuning models on your own data. It can be started by
+- clicking `Finetuning` in the plugin menu after starting `napari`.
+- running `$ micro_sam.train` in the command line.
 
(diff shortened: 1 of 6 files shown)


doc/annotation_tools.md
```diff
@@ -133,7 +133,9 @@ Check out [the video tutorial](https://youtu.be/HqRoImdTX3c) for an in-depth exp
 
 ## Finetuning UI
 
-We also provide a graphical inferface for fine-tuning models on your own data. It can be started by clicking `Finetuning` in the plugin menu after starting `napari`.
+We also provide a graphical inferface for fine-tuning models on your own data. It can be started by
+- clicking `Finetuning` in the plugin menu after starting `napari`.
+- running `$ micro_sam.train` in the command line.
```
constantinpape, line 138: I don't think that is quite correct. `micro_sam.train` does not start the UI, but runs training from the command line. I suggest we explain it somewhere else, and here also recommend to use either this CLI function or a script / notebook for finetuning, since it is more robust compared to the UI.

constantinpape (commented):
One minor comment, looks good otherwise.

doc/annotation_tools.md
anwai98: Oh yeah you are right. I didn't think about this. I refactored it in such a way that it now conveys the message as expected (i.e. finetuning via GUI is supported, however we recommend either the CLI or scripts / notebook for this purpose). Thanks!

constantinpape (approved):

PR #892, opened 2025-02-25
Minor fix to handle is_seg_dataset argument

anwai98:


anwai98:
Btw, I double checked it locally and the dataloaders work with this fix now!

cc: @constantinpape

PR #898, opened 2025-02-27
Move automatic segmentation mode

constantinpape:
This moves the determination of the automatic segmentation mode from the CLI function to `automatic_instance_segmentation`. This should address https://forum.image.sc/t/micro-sam-fine-tuning/109293 , which is very likely due to the fact that `with_background` is set to `True`; which currently cannot be over-written.

This change addresses the issue by not using the `binary_mask` segmentation mode for AIS at all, so the `with_background` parameter will not play a role.


constantinpape:
cc @anwai98

PR #899, opened 2025-02-27
Add evaluation CLI for segmentations

anwai98:
This PR adds a new CLI for quantitative evaluation of predicted instances w.r.t. ground-truth instance labels (using mean segmentation accuracy). The default behavior returns the metrics reported in our paper!

GTG from my side!


micro_sam/evaluation/evaluation.py
```diff
@@ -125,3 +151,93 @@ def run_evaluation_for_iterative_prompting(
...
+
+def main():
+    """@private"""
+    import argparse
+
+    parser = argparse.ArgumentParser(description="Evaluating segmentations from Segment Anything model on custom data.")
+
+    # labels and predictions for quantitative evaluation.
+    parser.add_argument(
+        "--labels", required=True, type=str, nargs="*",
```
constantinpape, line 164: Shouldn't this be `nargs="+"`? With `nargs="*"` you accept zero arguments. And I don't think that works.

micro_sam/evaluation/evaluation.py
```diff
@@ -125,3 +151,93 @@ def run_evaluation_for_iterative_prompting(
...
+        help="Filepath(s) to ground-truth labels or the directory where the label data is stored."
+    )
+    parser.add_argument(
+        "--predictions", required=True, type=str, nargs="*",
```
constantinpape, line 168: See above.

micro_sam/evaluation/evaluation.py
anwai98: Yes, you are right. I overlooked it due to the `required` argument. I fixed it in the latest commit. Thanks for spotting!

constantinpape (approved):

PR #910, opened 2025-02-28
Update embedding widget to pop dialog box for no images

### LINKED ISSUE #908: Open validation window when no image found for computing embeddings

anwai98:
Closes #908 

GTG from my side! (tested it on 2d annotator via `napari` Plugin option)

### DIFF
--- a/micro_sam/sam_annotator/_annotator.py
+++ b/micro_sam/sam_annotator/_annotator.py
@@ -150,25 +150,27 @@ class _AnnotatorBase(QtWidgets.QScrollArea):
     def _update_image(self, segmentation_result=None):
         state = AnnotatorState()
 
-        # Update the image shape if it has changed.
-        if state.image_shape != self._shape:
-            if len(state.image_shape) != self._ndim:
-                raise RuntimeError(
-                    f"The dim of the annotator {self._ndim} does not match the image data of shape {state.image_shape}."
-                )
-            self._shape = state.image_shape
+        if state.image_shape is not None:
+            # Update the image shape if it has changed.
+            if state.image_shape != self._shape:
+                if len(state.image_shape) != self._ndim:
+                    raise RuntimeError(
+                        f"The dim of the annotator {self._ndim} does not match "
+                        "the image data of shape {state.image_shape}."
+                    )
+                self._shape = state.image_shape
 
-        # Reset all layers.
-        self._viewer.layers["current_object"].data = np.zeros(self._shape, dtype="uint32")
-        self._viewer.layers["current_object"].scale = state.image_scale
-        self._viewer.layers["auto_segmentation"].data = np.zeros(self._shape, dtype="uint32")
-        self._viewer.layers["auto_segmentation"].scale = state.image_scale
-        if segmentation_result is None or segmentation_result is False:
-            self._viewer.layers["committed_objects"].data = np.zeros(self._shape, dtype="uint32")
-        else:
-            assert segmentation_result.shape == self._shape
-            self._viewer.layers["committed_objects"].data = segmentation_result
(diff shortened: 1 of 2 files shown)


micro_sam/sam_annotator/_widgets.py
```diff
@@ -893,11 +893,12 @@ def __init__(self, parent=None):
     def _initialize_image(self):
         state = AnnotatorState()
         layer = self.image_selection.get_value()
-        image_shape = layer.data.shape
-        image_scale = tuple(layer.scale)
-        state.image_shape = image_shape
-        state.image_scale = image_scale
-        state.image_name = layer.name
+        if layer is not None:
```
constantinpape, line 896: Why is this necessary? The case that the image layer is None should be caught by your error message in `_validate_inputs`, no?

micro_sam/sam_annotator/_annotator.py
```diff
@@ -150,25 +150,27 @@ def __init__(self, viewer: "napari.viewer.Viewer", ndim: int) -> None:
...
-        if segmentation_result is None or segmentation_result is False:
-            self._viewer.layers["committed_objects"].data = np.zeros(self._shape, dtype="uint32")
-        else:
-            assert segmentation_result.shape == self._shape
-            self._viewer.layers["committed_objects"].data = segmentation_result
-        self._viewer.layers["committed_objects"].scale = state.image_scale
-        self._viewer.layers["point_prompts"].scale = state.image_scale
-        self._viewer.layers["prompts"].scale = state.image_scale
-        vutil.clear_annotations(self._viewer, clear_segmentations=False)
+        if state.image_shape is not None:
```
constantinpape, line 153: I am also not sure why this is necessary. Going here with `image_shape = None` should not be possible.

constantinpape (approved):
I am not sure why you need to change so much? Wouldn't it be necessary to just add the error message you did in `_validate_inputs`?

micro_sam/sam_annotator/_widgets.py
anwai98: You are right. This shouldn't really be an issue, however the click trigger has the first step for `self._initialize_image` and then executes the `__call__` method, where we validate the inputs.

Which is why we need to have this part!

constantinpape: Can you change this to
```python
if layer is None:
    return
```
instead, and add a comment why it's needed? Then we have fewer changes in the code and it's easier to understand.

anwai98: On it.

micro_sam/sam_annotator/_annotator.py
anwai98: There's one tiny detail in this part as well: https://github.com/computational-cell-analytics/micro-sam/blob/master/micro_sam/sam_annotator/_annotator.py#L127

When the `_AnnotatorBase` runs the `__init__` method, then it executes `self._create_widgets`, which basically checks for the image shape and resets layers (which is tricky if the image layer does not exist).

Would you suggest the same as your feedback above?

constantinpape: Yep, I would suggest the same as above.

anwai98: On it.

anwai98: This is taken care of in the new commit. Thanks!

micro_sam/sam_annotator/_widgets.py
anwai98: Aaaaand done!

constantinpape (approved):

PR #916, opened 2025-03-03
Add implementation for Late PEFT

caroteu:
I made the most straight forward implementation of 'late lora'  by simply applying lora as it is (no additional adapters) to a selection of encoder blocks, specified in peft_kwargs by 'start_layer'. 

Late LoRA Tests (for convenience on OrgaSegment with n_objects set to 5):

regular lora: 12202MiB, ~16.27M Params
apply lora to layer 6-11: 10268MiB, ~15.68M Params
apply lora to layer 8-11: 8754MiB, ~15.49M Params
apply lora to only the last layer: 7924MiB, ~15.19M Params

### DIFF
--- a/finetuning/livecell_finetuning.py
+++ b/finetuning/livecell_finetuning.py
@@ -72,7 +72,9 @@ def finetune_livecell(args):
         save_root=args.save_root,
         scheduler_kwargs=scheduler_kwargs,
         save_every_kth_epoch=args.save_every_kth_epoch,
-        peft_kwargs={"rank": args.lora_rank, "start_layer": 10} if args.lora_rank is not None else None,
+        peft_kwargs={
+            "rank": args.lora_rank, "attention_layers_to_update": [9, 10, 11]
+        } if args.lora_rank is not None else None,
     )
 
     if args.export_path is not None:
--- a/finetuning/livecell_finetuning.py
+++ b/finetuning/livecell_finetuning.py
@@ -56,6 +56,12 @@ def finetune_livecell(args):
     train_loader, val_loader = get_dataloaders(patch_shape=patch_shape, data_path=args.input_path)
     scheduler_kwargs = {"mode": "min", "factor": 0.9, "patience": 10, "verbose": True}
 
+    # freeze encoder: 35.02 GB
+    # late lora 50% classic: 43.142 GB
+    # late FT 50%: 42.87 GB
+    # lora classic: 48.46 GB
+    # full ft: 49.35
+
     # Run training.
     sam_training.train_sam(
         name=checkpoint_name,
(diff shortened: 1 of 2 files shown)


caroteu:
@anwai98

anwai98:
Hi @caroteu,

This looks great to me. Thanks!

I made two tiny changes:
- I changed the attribution for `start_layer`, such that it matches the mathematical counts for layers, i.e. 10 would signify that out of 1 to 12 layers, the PEFT methods would be applied to 10, 11 and 12. I think it's simpler to understand this way.
- And added a check on top to make sure that `start_layer` stays in expected bounds!

I say let's fire up the GPUs now! ;)

anwai98:
On a second thought on design detail, we already have a parameter `attention_layers_to_update`, to which if we pass `[10, 11, 12]`, it would be the same effect as `start_layer`.

Let's visit the implementation details later to finalize this. For now, go ahead with the experiments with `start_layer`!

micro_sam/models/peft_sam.py
```diff
@@ -375,7 +379,16 @@ def __init__(
...
-        for t_layer_i, blk in enumerate(model.image_encoder.blocks):
+        # The starting layer cannot be greater than the total number of blocks.
+        if start_layer and start_layer > len(model.image_encoder.blocks):
+            raise ValueError("The chosen layer to start PEFT method is greater than the total transformer blocks.")
+
+        for t_layer_i, blk in enumerate(model.image_encoder.blocks, start=1):
+
+            # For late lora, we only apply lora on the layers after a certain "start_layer".
+            if start_layer and t_layer_i < start_layer:
+                continue
```
caroteu, line 390: @anwai98 I think we are never changing the last block this way. We should remove start=1 from the enumerate expression and instead just change `t_layer_i < start_layer` to `t_layer_i < start_layer`. That should do the job imo

anwai98:
Hi @caroteu,

I agree with your comment above. However, I do see the obvious confusion where it came from. If we are removing the starting id for enumerate, I think it's best to stick with `attention_layers_to_update` in this case anyways then! (as we have supports for it and gives us better control over our choice of layers)

I kept the checks for matching if the passed `attention_layers_to_updates` matches the expectation or not.

Let me know if this works as expected for you as well! (I'll point to an example script I tried in a following comment)

finetuning/livecell_finetuning.py
```diff
@@ -85,7 +72,9 @@ def finetune_livecell(args):
         save_root=args.save_root,
         scheduler_kwargs=scheduler_kwargs,
         save_every_kth_epoch=args.save_every_kth_epoch,
-        peft_kwargs={"rank": args.lora_rank, "quantize": True} if args.lora_rank is not None else None,
+        peft_kwargs={
+            "rank": args.lora_rank, "attention_layers_to_update": [9, 10, 11]
```
anwai98, line 76: Here's how I tried it and worked for me.

anwai98:
Okay, the implementation looks GTG from my side!

@caroteu let us know if there's anything to add from your side. Meanwhile, I'll put this PR for a review by @constantinpape !

constantinpape (approved):
Looks good to me. Feel free to merge if everything is added from your side.

PR #917, opened 2025-03-04
Support low rank adaptation for all matrices in attention blocks

caroteu:
@anwai98 
Adding adapters to the key matrices and feed forward layers does not add any significant overload in the memory while increasing the number of trainable parameters. For example if we add late lora to blocks 6-11:

Only on Q, V: 10268MiB, ~15.68M Params
On Q, V, K, MLP: 10296MiB, ~17.45M Params

### DIFF
--- a/finetuning/livecell_finetuning.py
+++ b/finetuning/livecell_finetuning.py
@@ -56,6 +56,12 @@ def finetune_livecell(args):
     train_loader, val_loader = get_dataloaders(patch_shape=patch_shape, data_path=args.input_path)
     scheduler_kwargs = {"mode": "min", "factor": 0.9, "patience": 10, "verbose": True}
 
+    # freeze encoder: 35.02 GB
+    # late lora 50% classic: 43.142 GB
+    # late FT 50%: 42.87 GB
+    # lora classic: 48.46 GB
+    # full ft: 49.35
+
     # Run training.
     sam_training.train_sam(
         name=checkpoint_name,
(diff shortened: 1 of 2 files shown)


micro_sam/models/peft_sam.py
```diff
@@ -365,6 +365,20 @@ def __init__(self, block: nn.Module):
...
+
+    def __init__(self, block: nn.Module):
+        super().__init__(block=block)
+        self.block = block
+
+        for k, v in self.block.named_parameters():
+            v.requires_grad = True
+
+    def forward(self, x):
+        return x
```
anwai98, line 379: Just to be sure @caroteu, we need this so that we allow gradient updates on all layers in a block and not just the attention matrices, right?

anwai98: I see. So in theory, we desire to "unfreeze" all parts of the attention block?

anwai98: Nice. Thanks for clarifying!

finetuning/livecell_finetuning.py
```diff
@@ -56,6 +56,12 @@ def finetune_livecell(args):
     train_loader, val_loader = get_dataloaders(patch_shape=patch_shape, data_path=args.input_path)
     scheduler_kwargs = {"mode": "min", "factor": 0.9, "patience": 10, "verbose": True}
 
+    # freeze encoder: 35.02 GB
+    # late lora 50% classic: 43.142 GB
+    # late FT 50%: 42.87 GB
+    # lora classic: 48.46 GB
+    # full ft: 49.35
+
```
anwai98, line 64: I can reproduce the memory efficiency on LIVECell atleast. I am very confident now that there is some stochastic memory issues you are facing on HPA.

cc: @caroteu

anwai98:
I think this is in a good state. I'll go ahead and merge this in the running Late PEFT main PR, once the tests pass!

cc: @caroteu

PR #919, opened 2025-03-06
Add support for avoiding to overwrite trained model

anwai98:
This PR adds support to avoid overwriting a trained model, if the user specifies it. By default, it is the same behavior as without this PR, i.e. a model will be overwritten regardless. GTG from my side!


anwai98:
I'll go ahead and merge this for now, as this does not functionally affect the default behavior for our training setup!

PR #926, opened 2025-03-11
Implement batched embedding computation

constantinpape:


anwai98 (commented):
The changes look good to me. Thanks @constantinpape!

I made a minor bug fix to ensure that the correct `original_size` is extracted.

Two mentions which might be helpful:
- For 2d (eg. tried in on our `patho-sam` WSI): I do see an advantage of using a higher batch size.
- For 3d (eg. tried it on `lucchi` test volume): There is almost no difference for a higher batch size without tiling (but it runs fast anyways due to very few z-slices), and it is slightly slow with tiling involved for higher batch size.

NOTE: Both mentions above were tried via automatic segmentation CLI, which now supports an argument `--batch_size`!

constantinpape:
> I made a minor bug fix to ensure that the correct `original_size` is extracted.

Thanks for catching this!



> * For 2d (eg. tried in on our `patho-sam` WSI): I do see an advantage of using a higher batch size.
>     * For 3d (eg. tried it on `lucchi` test volume): There is almost no difference for a higher batch size without tiling (but it runs fast anyways due to very few z-slices), and it is slightly slow with tiling involved for higher batch size.

Thanks for checking! I think 3D might not profit because some other part of the computation has overhead and is limiting here. I will merge it now, and we can discuss further steps on this tomorrow.

PR #927, opened 2025-03-11
Update model choice design in annotator tools

anwai98:
WIP!

This PR updates the GUI structure of model choice, by exposing the 'base' variants for all domains, and moving other size options to the advanced embedding settings. I am missing two thing atm:
- Dynamic change of model sizes, eg. `Default` should show tiny-huge (all 4), and `Medical Imaging` should show only base.
- The `model_type` is not getting updated somehow. I will investigate closely what's going on!

### DIFF
--- a/micro_sam/sam_annotator/_widgets.py
+++ b/micro_sam/sam_annotator/_widgets.py
@@ -953,6 +953,9 @@ class EmbeddingWidget(_WidgetBase):
         if "segment_nd" in state.widgets:
             vutil._sync_ndsegment_widget(state.widgets["segment_nd"], self.model_type, self.custom_weights)
 
+    def _update_model_type(self):
+        self.model_type = "vit_" + self.model_size[0] + self.supported_dropdown_maps[self.model_family]
+
     def _create_model_section(self):
         self.model_family = "Default"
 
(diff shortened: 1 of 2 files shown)


anwai98:
> The `model_type` is not getting updated somehow. I will investigate closely what's going on!

Okay, this is fixed now. Works as expected! (love the design now)

> Dynamic change of model sizes, eg. `Default` should show tiny-huge (all 4), and `Medical Imaging` should show only base.

This one's messing with me bad. I would really love to make this one work, but this cross-connect triggers via changing drop-down indices is confusing me a lot. I will check this with a fresh mind tomorrow, maybe third times' the charm!

anwai98:
Okay, I think everything should work as expected now! Let me know how it looks, @constantinpape!

constantinpape (approved):
I think this all looks good. We should also take care of:
- Better GUI when using custom models (so that this is reflected both in the UI and in the window that pops up when computing embeddings)
    - Ideally, we also write this to the embedding metadata 
- Integration of https://github.com/computational-cell-analytics/micro-sam/pull/886

I think it makes sense to work on this in a follow-up PR, we can quickly discuss the strategy.

Issue #921, opened 2025-03-06
Store automatic segmentation CLI outputs in stages


anwai98:
i.e. before continuing to annotate, and once after to overwrite it.

This way, the user would not loose their progress if some process breaks!

PR #928, opened 2025-03-11
Save outputs of automatic segmentation CLI in stages

anwai98:
GTG from my side. And closes https://github.com/computational-cell-analytics/micro-sam/issues/921

### DIFF
--- a/micro_sam/automatic_segmentation.py
+++ b/micro_sam/automatic_segmentation.py
@@ -177,6 +177,11 @@ def automatic_instance_segmentation(
             **generate_kwargs
         )
 
+    # Before starting to annotate, if at all desired, store the automatic segmentations in the first stage.
+    if output_path is not None:
+        imageio.imwrite(output_path, instances, compression="zlib")
+        print(f"The automatic segmentation results are stored at '{os.path.abspath(output_path)}'.")
+
     # Allow opening the automatic segmentation in the annotator for further annotation, if desired.
     if annotate:
         from micro_sam.sam_annotator import annotator_2d, annotator_3d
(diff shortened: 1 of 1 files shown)


micro_sam/automatic_segmentation.py
```diff
@@ -201,10 +206,10 @@ def automatic_instance_segmentation(
...
         instances = viewer.layers["committed_objects"].data
 
-    # Save the instance segmentation, if 'output_path' provided.
-    if output_path is not None:
-        imageio.imwrite(output_path, instances, compression="zlib")
-        print(f"The segmentation results are stored at '{os.path.abspath(output_path)}'.")
+        # Save the instance segmentation, if 'output_path' provided.
+        if output_path is not None:
+            imageio.imwrite(output_path, instances, compression="zlib")
+            print(f"The final segmentation results are stored at '{os.path.abspath(output_path)}'.")
```
constantinpape, line 212: This way, we over-write the previous saved results. Or am I missing something? If yes, I would maybe update this to not over-write them, e.g. by changing the output_path in line 181, if annotate is set.

anwai98: You are right. This is exactly the expectation here. I'll explain my thought for the possible workflows:
- Say a user uses the CLI for automatic segmentation only, then the first stage of saving does the job for them and that's it.
- Say a user uses the CLI for automatic segmentation and chooses to continue annotation, we store the automatic segmentation first and then the interactive segmentation updates get overwritten in the same filepath.

Now that you mention it, I see your vision. Would you suggest to create clones for different stage of outputs with different names without overwriting the files one-to-one?

eg. for `-o output.tif`, we store `output.tif` for the final version,  and `output_automatic.tif` as the first stage of automatic segmentation outputs? (so that users have access to both)

constantinpape: > eg. for `-o output.tif`, we store `output.tif` for the final version, and `output_automatic.tif` as the first stage of automatic segmentation outputs? (so that users have access to both)

Yes, that is exactly what I have in mind :)

PR #933, opened 2025-03-13
Add download progress bar to napari info log

anwai98:
WIP!

### DIFF
--- a/micro_sam/sam_annotator/_annotator.py
+++ b/micro_sam/sam_annotator/_annotator.py
@@ -160,6 +160,10 @@ class _AnnotatorBase(QtWidgets.QScrollArea):
     def _update_image(self, segmentation_result=None):
         state = AnnotatorState()
 
+        # Whether embeddings already exist and avoid clearing objects in layers.
+        if state.skip_recomputing_embeddings:
+            return
+
         # This is encountered when there is no image layer available / selected.
         # In this case, we need not update the image shape or check for changes.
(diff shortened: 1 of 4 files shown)


micro_sam/util.py
```diff
@@ -345,7 +347,15 @@ def get_sam_model(
...
-        checkpoint_path = model_registry.fetch(model_type, progressbar=True)
+
+        progress_bar = True
+        # Check if we have to download the model.
+        # If we do and have a progress bar factory, then we over-write the progress bar.
+        if not os.path.exists(os.path.join(get_cache_directory(), model_type)) and progress_bar_factory is not None:
+            progress_bar = progress_bar_factory(model_type)
+
+        checkpoint_path = model_registry.fetch(model_type, progressbar=progress_bar)
+        progress_bar.close()  # Close the progress bar when the task finishes.
```
constantinpape, line 358: This will not work if you have a bool value (i.e. the default). You need to add a check to only do this if this is not bool.

anwai98: Ah yeah right. Thanks for spotting this!

anwai98:
And FYI @constantinpape, `show_info` has (or rather had) a bug on the latest stable version, but has now been fixed at https://github.com/napari/napari/issues/7654 (and I think the fix will be a part of napari v0.6).

To reproduce it, just press "x" on the popped info window, and it returns "segmentation core (dumped)" for me!

We can stick to keeping it on `dev` for now, but should keep this in mind that it can happen to users!

PR #936, opened 2025-03-14
Fix issues in the image series annotator and load existing segmentati…

constantinpape:
…ons if skip_segmented=False


constantinpape:
The mac timeout in the test is known and fixed by skipping the respective test on dev already.

Issue #942, opened 2025-03-17
Recreate layers that users have (accidentally) removed


constantinpape:
E.g. recreate the auto segmentation layer if it was removed.

cc @anwai98

anwai98:
This is taken care of now. I'll go ahead and close this!

PR #943, opened 2025-03-17
Ensure valid layers exist for each annotator click menu

anwai98:
This PR ensures if all layers exist or not. If not, it creates dummy layers and avoids throwing ambiguous errors. I checked it for 2d annotator only, but should work for all annotators, as they share the same components!

PS. One thing that is still tricky is the `commit` option. It can't account for missing layers. I am thinking if I can raise a error window stating that the layers do not exist, hence commit won't work, and once the user closes the window, create dummy layers. (I'll check this out now)

WIP!


anwai98:
Aaaaaaaand this works all as expected now! ;) GTG from my side! @constantinpape

anwai98:
The tracking annotator needs some specific changes for the layers, as it is hard-coded at the moment in the annotator tracking (which also makes sense as it needs specific label layers). I will look into fixing this tomorrow! (PS. I can reproduce the issue on my laptop which is easy to debug)

constantinpape:
Ah I see, this is for fixing #942 and not for the current dev. I didn't understand that when we discussed this during breakfast. Let's wait till after the last TiM workshop to fix this. I would avoid making live changes, unless they fix genuine bugs.

micro_sam/sam_annotator/_widgets.py
```diff
@@ -715,7 +719,69 @@ def _validate_embeddings(viewer: "napari.viewer.Viewer"):
     #     return False
 
 
+def _validate_layers(viewer, layer_choice=None):
+
+    # Let's find the first image layer to use as our reference for getting the shape.
```
constantinpape, line 724: This is not a good idea. We can't just assume that the first image layer we have is the image we annotate.

constantinpape: Instead, we can either use [state.image_shape](https://github.com/computational-cell-analytics/micro-sam/blob/master/micro_sam/sam_annotator/_state.py#L45), or maybe even better, we store the name of the reference image layer in the state as well, and update it where we also update `state.image_shape` and `state.image_scale`.

micro_sam/sam_annotator/_widgets.py
```diff
@@ -715,7 +719,69 @@ def _validate_embeddings(viewer: "napari.viewer.Viewer"):
     #     return False
 
 
+def _validate_layers(viewer, layer_choice=None):
```
constantinpape, line 722: This function is almost identical to [annotator._create_layers](https://github.com/computational-cell-analytics/micro-sam/blob/master/micro_sam/sam_annotator/_annotator.py#L18). We should avoid that and rather refactor this into a common function.

constantinpape: We should possibly also unify this with [annotator._update_image](https://github.com/computational-cell-analytics/micro-sam/blob/master/micro_sam/sam_annotator/_annotator.py#L150).
In any case, note that we have to take care of the scale here as well.

constantinpape:
The test error occurs because `AnnotatorTracking._create_layers` is now called twice. I don't quite know why, but we should overhaul the design a bit, because having three very similar functions: `_create_layers`, `_update_image` and `_validate_layers` is not a good design. Let's discuss this on Monday.

micro_sam/sam_annotator/_widgets.py
anwai98: This is slightly complicated if we seek for image shape from state (we then would have to add the state in several tiny parts which do not really make use of it). I will think about this and discuss the details with you on Monday!

micro_sam/sam_annotator/_widgets.py
anwai98: Good point. Would make sense to unify them all. I'll think about it!

constantinpape:
We rename `Annotator._create_layers` to `Annotator._require_layers`, so that it unifies the functionality of `_validate_layers` and `_create_layers`. The tracking annotator over-writes this implementation (as it does right now for `_create_layers` anyways). In order to have access to the annotator everywhere we also add it to the `AnnotatorState`.

constantinpape:
We should also call a `AnnotatorState.reset_state` in the Annotator inits.

anwai98:
Okay, took me a bit longer to sort things out in an elegant way, but we have it all now (hopefully)! Let me know how it looks @constantinpape!

PS. I tested both the aforementioned suggestions on annotator 2d.

micro_sam/sam_annotator/_annotator.py
```diff
@@ -16,15 +16,21 @@ class _AnnotatorBase(QtWidgets.QScrollArea):
     The annotators differ in their data dimensionality and the widgets.
     """
 
-    def _create_layers(self):
+    def _require_layers(self, layer_choice=None, shape=None):
         # Add the label layers for the current object, the automatic segmentation and the committed segmentation.
-        dummy_data = np.zeros(self._shape, dtype="uint32")
+        dummy_data = np.zeros(self._shape if shape is None else shape, dtype="uint32")
```
constantinpape, line 21: What happens if `shape` and `self._shape` disagree? I think we should be careful here, and either:
i) not pass `shape` as an argument at all and rely on `self._shape`
ii) update `self._shape` if it differs from `shape` (and maybe raise a warning)?

micro_sam/sam_annotator/_annotator.py
```diff
@@ -16,15 +16,21 @@ class _AnnotatorBase(QtWidgets.QScrollArea):
...
 
         # Before adding new layers, we always check whether a layer with this name already exists or not.
         if "current_object" not in self._viewer.layers:
+            if "current_object" == layer_choice:  # Check at 'commit' call button.
```
constantinpape, line 25: Is the layer choice necessary? Should we even raise this window?

micro_sam/sam_annotator/_state.py
```diff
@@ -69,6 +70,9 @@ class AnnotatorState(metaclass=Singleton):
     # z-range to limit the data being committed in 3d / tracking.
     z_range: Optional[Tuple[int, int]] = None
 
+    # annotator_class
+    annotator_class: Optional["micro_sam.sam_annotator._annotator._AnnotatorBase"] = None
```
constantinpape, line 74: This should just be called `annotator`

constantinpape: `annotator_class` implies that you hold the actual class and not an instantiated object.

micro_sam/sam_annotator/annotator_tracking.py
```diff
@@ -101,7 +101,7 @@ class AnnotatorTracking(_AnnotatorBase):
     # The tracking annotator needs different settings for the prompt layers
     # to support the additional tracking state.
     # That's why we over-ride this function.
-    def _create_layers(self):
+    def _require_layers(self):
```
constantinpape, line 104: We need to add the logic to prevent adding layers multiple times here as well. And you should have the same function signature as in the other case.

constantinpape (commented):
The design looks good overall, only some detailed comments.

micro_sam/sam_annotator/_annotator.py
anwai98: Okay yes, you are right. We don't really need it explicitly besides the correct `self._shape` value. I will remove exposing this parameter!

micro_sam/sam_annotator/_state.py
anwai98: Ahha yes, makes sense. I will update this. Thanks!

micro_sam/sam_annotator/_annotator.py
constantinpape: But, taking the example of `commit`: don't you need to also have `commited_objects` in that case? It feels a bit arbitrary to just single out one layer.

anwai98: The thing is, the `committed_objects` is the target layer, and the dropdown choices only allow `current_objects` or `auto_segmentation`, which is where wrapping this was easy.

Can you suggest how can I look for the missing layer and pop the window for the `commit` call button setting?

micro_sam/sam_annotator/_widgets.py
```diff
@@ -509,6 +509,11 @@ def commit(
         commit_path: Select a file path where the committed results and prompts will be saved.
             This feature is still experimental.
     """
+
+    # Check whether all layers exist as expected or create new ones automatically.
+    state = AnnotatorState()
+    state.annotator._require_layers(layer_choice=layer)
```
constantinpape, line 515: You could pass a list here: `layer_choices=["committed_objects", layer]`

micro_sam/sam_annotator/_widgets.py
```diff
@@ -551,6 +556,10 @@ def commit_track(
     # Commit the segmentation layer.
     id_offset, seg, mask, bb = _commit_impl(viewer, layer, preserve_committed)
 
+    # Check whether all layers exist as expected or create new ones automatically.
+    if _validate_layers(viewer):
+        return None
```
constantinpape, line 561: Why don't we use the same pattern as in 514-515 here? In fact, we could just move this to `_commit_impl` to avoid code duplication.

micro_sam/sam_annotator/annotator_tracking.py
```diff
@@ -101,7 +101,43 @@ class AnnotatorTracking(_AnnotatorBase):
...
+        if "committed_objects" not in self._viewer.layers:
+            self._viewer.add_labels(data=dummy_data, name="committed_objects")
+            # Randomize colors so it is easy to see when object committed.
+            self._viewer.layers["committed_objects"].new_colormap()
+            if image_scale is not None:
+                self.layers["committed_objects"].scale = image_scale
+
+        # Add the point prompts layer.
+        # NOTE: The lines below ensure that there is no existing 'point_prompts' layer with same name, and remove them.
+        if "point_prompts" in self._viewer.layers:
```
constantinpape, line 138: Why do we change the pattern here? Above we check if the layer exists and add them otherwise. Here we first remove the layer and then always add. Why don't we use the same pattern as before?

micro_sam/sam_annotator/annotator_tracking.py
```diff
@@ -165,14 +198,23 @@ def _get_widgets(self):
             "clear": widgets.clear_track(),
         }
 
-    def __init__(self, viewer: "napari.viewer.Viewer") -> None:
+    def __init__(self, viewer: "napari.viewer.Viewer", reset_state: bool = True) -> None:
```
constantinpape, line 201: You need to set `reset_state=False` when this is called in the function `annotator_tracking` below.

micro_sam/sam_annotator/_annotator.py
constantinpape: Why don't you make it a list `layer_choices`? Then you can pass both the current layer and `committed_objects` in `commit`.

constantinpape (commented):
I think the core logic is there now, only a few minor things to address.
Besides the things I mentioned in the comments: you also need to set `reset_state=False` here: https://github.com/computational-cell-analytics/micro-sam/blob/master/micro_sam/sam_annotator/image_series_annotator.py#L100-L107

micro_sam/sam_annotator/_widgets.py
anwai98: Oh yeah good point. That would make it robust to the point you raised. Nice. I will take care of it. Thanks! :D

micro_sam/sam_annotator/_widgets.py
anwai98: Good point. Then atleast for the commit call, we have it unified. Thanks!

micro_sam/sam_annotator/annotator_tracking.py
anwai98: For this step, we have a different set of properties for the `point_prompts` layer in the tracking annotator, which is why we can't search for it and consider it to be valid. This was a critical step to make the tracking annotator work as expected and ensure that we have all desired layers only!

micro_sam/sam_annotator/annotator_tracking.py
anwai98: Ah missed that. Thanks for spotting!

micro_sam/sam_annotator/annotator_tracking.py
constantinpape: Ah, I see! Makes sense. Please make sure to test that the annotator still works with these changes.

anwai98: I tested it on my laptop, and the tests pass! ;)
(and the MacOS tests for annotator tracking are enabled, which now pass as well)

constantinpape: Can you actually run the annotator tool and check that it works? The tests are a good starting point, but they don't cover the full interactive functionality. For example, use https://github.com/computational-cell-analytics/micro-sam/blob/master/examples/annotator_tracking.py

anwai98: Okay, something is broken (interactive segmentation in the tracking annotator does not work atm) (double checked on `dev` , and all is good, so I broke something here).

I will investigate this and come back to you!

anwai98: Okay, so things are slightly tricky, but the latest commit pushes a stable fix for the mess:

1) > Above we check if the layer exists and add them otherwise.

So your mention above was actually in the right direction. I need to check a "bit smartly" if the point prompts layer is connected to the tracking widget or not. Recreating it everytime (which is what I did) does basically nothing (as you are in the loop of always updating the layer and not doing anything).

2) In the current solution I pushed, if the user stays on one annotator experiments (eg. not mix widgets up without manually removing layers), everything works as expected.

I have a plan to handle this better: check for `property_choices`: this I think is available in the metadata. I just need to extract it and validate properly. I will give this a shot now!

anwai98: That was an adventure, but now all is figured out now.

I tested the CIs and the example scripts for tracking as well.

anwai98:
I would say everything works as expected now. Let me know how it looks @constantinpape

Also, once you have time, it would be nice to have another pair of eyes testing the annotators!

constantinpape:
I checked the changes and they look good. However, one thing is not working in the tracking annotator:
after removing a point layer the division state doesn't work anymore. See the video:

https://github.com/user-attachments/assets/e01eeeb6-b661-4d8c-95b0-6d779c7ea13c


(the point should change the color, and also the division is not taken into account when running segmentation)

anwai98:
I have a suspicion why this might be not working. I will check back on this later in the evening!

anwai98:
Hmm strange that I cannot reproduce the issue you encountered. I will check back once again after lunch and post you with a video (with the latest state of `ensure-valid-layers` branch)

constantinpape:
> Edit 1: Okay I spoke too soon, sorry. I can reproduce the exact same steps you showed.
> 
> Edit 2: The video behavior only happens when the `point_prompts` layer is deleted. I'll check if I can debug this!

Yes, indeed it only happens when the point prompts are removed.

I am pretty sure the problem is because the tracking menu (created [here](https://github.com/computational-cell-analytics/micro-sam/blob/master/micro_sam/sam_annotator/annotator_tracking.py#L24)) only works for the specific point prompt layer, and the connected events are not working anymore once the layer is removed and added again.

It should be possible to fix this by reconnecting the events:
- You can get the menu via `AnnotatorState.widgets["tracking"]` (see [here](https://github.com/computational-cell-analytics/micro-sam/blob/master/micro_sam/sam_annotator/annotator_tracking.py#L159))
- Then you would need to redo these event connections: https://github.com/computational-cell-analytics/micro-sam/blob/master/micro_sam/sam_annotator/annotator_tracking.py#L34-L96

It's probably easiest to update the `create_tracking_menu` function, so that you can optionally pass the `tracking_widget` as well, and in that case just skip the first few lines. You probably wouldn't need to touch the rest of the function and it should correctly re-initialize the event connections.

anwai98:
Okay wow your suggestion worked (thank you so very much @constantinpape <3)
Everything should be working now (tried several combinations with the major fixes).
GTG from my side (finally)!

constantinpape:
I still get an error when switching to division after the point prompts have been removed: (only a short snippet of the full error message)
```
EmitLoopError: 

While emitting signal 'magicgui.widgets.ComboBox.changed', a KeyError occurred in a callback:

  Signal emitted at: /home/pape/micromamba/envs/main/lib/python3.11/site-packages/magicgui/widgets/bases/_value_widget.py:71, in _on_value_change
    >  self.changed.emit(value)

  Callback error at: /home/pape/micromamba/envs/main/lib/python3.11/site-packages/pandas/core/indexes/base.py:6249, in _raise_if_missing
    >  raise KeyError(f"None of [{key}] are in the [{axis_name}]")

    Local variables:
       key = Index([0], dtype='int64')
       indexer = array([-1])
       axis_name = 'index'
       missing_mask = array([ True])
       nmissing = 1

See KeyError above for original traceback.
```

constantinpape:
I checked this out again, and it only happens when removing the point layer multiple times. I don't quite now how to get to the root cause, but I think this is an acceptable limitation since:
i) It's still much better than the current state, where nothing works anymore if the layer is removed at all.
ii) I think it's quite unlikely that a user removes a layer twice accidentally in the same session.

I will go ahead and merge it once the tests pass.

constantinpape (approved):

anwai98:
Thanks for confirming this @constantinpape!

> I checked this out again, and it only happens when removing the point layer multiple times. 

This is super useful to know. I will open an issue on this so that we are aware of this behaviour on current `dev` and see if we can figure it out in future!

PR #954, opened 2025-03-20
Update windows installation instruction doc

anwai98:
This PR removes the `pytorch` channel in our official doc for installing `micro-sam` on Windows. GTG from my side!


anwai98:
Something is broken in `xarray` package now (they made a new release today)

constantinpape:
This needs to be fixed in `bioimageio.core` (or we pin xarray)

constantinpape:
I don't think this will fix the issue. We should wait on the fix of nifty before addressing this. Everything else will be a hotfix that likely makes this worse.

PR #962, opened 2025-03-24
Validate model type with model state instead of state

caroteu:
@anwai98 I think we need to use model state instead of state here or else the validation won't work as desired. For instance with a custom checkpoint for a micro-sam model the keys for the state would be "model_state", "decoder_state"... (Same goes for any finetuned checkpoint where the keys include optimizer state, training times, etc. )


anwai98 (approved):
Good catch ;)

Thanks @caroteu. I will go ahead and merge this once the tests pass!

PR #964, opened 2025-03-25
Update model combination tooltips and adapt other GUIs

anwai98:
WIP!

This PR adapts the image series annotator and finetuning GUI to the latest model selection logic.

### DIFF
--- a/micro_sam/sam_annotator/image_series_annotator.py
+++ b/micro_sam/sam_annotator/image_series_annotator.py
@@ -370,7 +370,28 @@ class ImageSeriesAnnotator(widgets._WidgetBase):
         self.run_button.clicked.connect(self.__call__)
         self.layout().addWidget(self.run_button)
 
-    # model_type: str = util._DEFAULT_MODEL,
+    def _get_model_size_options(self):
+        # We store the actual model names mapped to UI labels.
+        self.model_size_mapping = {}
+        if self.model_family == "Natural Images (SAM)":
+            self.model_size_options = list(self._model_size_map .values())
+            self.model_size_mapping = {self._model_size_map[k]: f"vit_{k}" for k in self._model_size_map.keys()}
+        else:
+            model_suffix = self.supported_dropdown_maps[self.model_family]
+            self.model_size_options = []
+
+            for option in self.model_options:
+                if option.endswith(model_suffix):
+                    # Extract model size character on-the-fly.
+                    key = next((k for k in self._model_size_map .keys() if f"vit_{k}" in option), None)
+                    if key:
+                        size_label = self._model_size_map[key]
+                        self.model_size_options.append(size_label)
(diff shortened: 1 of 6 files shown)


anwai98:
Okay, so I synced up the image series annotator to match our new model combination style (and added specific tooltips for model family and model size dropdowns)

My next question would be: what's the best strategy to go ahead with `Initial model` choice? (current expectation: `vit_b_em_organelles`)

There are two ways to go with this:
- Have another string parameter called `Model size`, such that the users can now pass: `Initial model: "Electron Microscopy"` and `Model size: Large`
- Or, use drop-down options for model family and size here. However, with this option, it needs a bit more work as I would have to duplicate a few things from the embedding widgets (or maybe find a nice common solution) and find a way to combine them all, subject to the `Configuration` parameter!

What do you suggest here @constantinpape?

micro_sam/sam_annotator/image_series_annotator.py
constantinpape, line 1: Good catch, we should definitely update this one! Would it make sense to unify the code here with the respective code from the Embedding widget? If they are very similar it could be a good idea to refactor some parts into common functions.

anwai98: My quick answer for this would be yes, this is def. possible and something we should do.

(Because if we can unify this, that would make maintaining stuff so much convenient in future)

I will give this a shot today evening, if it's not too much work. Otherwise, I bring it up to the state where the GUI matches others, and then we do the functional unification later!

anwai98:
>  In this case, I would just expose the options of the model family widget and I would hard-code the size to vit_b. Given our experiences this is the best choice, so I would just keep it fixed in the GUI.

That's exactly what I wanted to know. Thanks ;)

anwai98:
Before signing off now, I wanted to bring all GUI stuff in sync (i.e. all GUIs now have the model family and model size structure).

Unfortunately, I didn't manage to unify the common stuff into the base widget class. I will give it a quick shot in the morning (I think it's overall easy to do, it's just a lot of sanity checks to be done), otherwise we can probably go ahead with this version for now, and focus on unification in the next release!

PS. It's hard to test the training GUI on my laptop, but I did a super quick test, and until loader checks on 1 image, training works.

cc: @constantinpape

constantinpape:
Sounds good! I briefly looked at the current changes and they look good. Let me know once things are unified, then I will also check them out.

anwai98:
Okay, all looks good from my side (did a super simple and quick sanity check across all GUI stuff and all seems to work for me).

Let me know how everything looks!

PS. A quick TLDR for the idea is that I moved all model choice-related stuff for GUI in the `_WidgetBase` class ;) (since all widgets rely on it, it's now super easy to control stuff from there for all GUIs)

constantinpape (approved):
This looks exactly like what I had in mind!

PR #967, opened 2025-03-27
Better preservation for committed objects

constantinpape:


anwai98 (approved):
Looks great to me! (I quickly tested it out on 2d annotator)

PR #972, opened 2025-03-28
Minor update to environment file and other docs

anwai98:
This PR addresses super minor changes to docs and an important thing for the environment file.

The important thing here probably is to pin `napari` for the working version (so that we don't have another round of unforeseen troubles if users install environments in the new release period of `napari`). What do you think @constantinpape?

Besides that, GTG from my side!

### DIFF
--- a/environment.yaml
+++ b/environment.yaml
@@ -20,7 +20,7 @@ dependencies:
     # - libtorch=*=cuda11*
     # or, to enforce a CPU installation, change to
     # - "pytorch=*=cpu*"
-    - pytorch >=2.6
+    - pytorch >=2.5
     - segment-anything
     - torchvision
     - torch_em >=0.7.8
--- a/environment.yaml
+++ b/environment.yaml
@@ -5,7 +5,7 @@ dependencies:
     - nifty >=1.2.3
     - imagecodecs
     - magicgui
-    - napari >=0.5.0
+    - napari >=0.5.0,<=0.5.6 
     - natsort
     - pip
     - pooch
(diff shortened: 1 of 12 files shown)


environment.yaml
```diff
@@ -5,7 +5,7 @@ dependencies:
     - nifty >=1.2.3
     - imagecodecs
     - magicgui
-    - napari >=0.5.0
+    - napari >=0.5.0,<=0.5.6
```
constantinpape, line 8:
```suggestion
    - napari >=0.5.0,<0.6.0
```

environment.yaml
```diff
@@ -20,10 +20,10 @@ dependencies:
     # - libtorch=*=cuda11*
     # or, to enforce a CPU installation, change to
     # - "pytorch=*=cpu*"
-    - pytorch >=2.5
+    - pytorch >=2.6
```
constantinpape, line 23: Is this necessary? Do we really need 2.6?

environment.yaml
```diff
@@ -20,10 +20,10 @@ dependencies:
...
     - segment-anything
     - torchvision
-    - torch_em >=0.7.7
+    - torch_em >=0.7.8
```
constantinpape, line 26: Do we really need 0.7.8?

constantinpape:
> The important thing here probably is to pin `napari` for the working version (so that we don't have another round of unforeseen troubles if users install environments in the new release period of `napari`).

Makes sense, but it is better to have a more lax pin that just demands <0.6 (I changed this already).

Besides this, is it really necessary to change the pins for pytorch and torch-em? It's best to only change those if it's really necessary.

environment.yaml
anwai98: We don't need to bump this, you are right. I just pushed it to the latest `pytorch` from conda forge!

constantinpape: Then let's undo this. The rule for pins is to only update them if necessary. They should express real constraints.

anwai98: Fixed it in the latest commit. Thanks Constantin!

anwai98:
> Makes sense, but it is better to have a more lax pin that just demands <0.6 (I changed this already).

I see. Good to know that it works for not released version numbers as well. Thanks for the hint!

PR #975, opened 2025-03-28
Update the documentation for v1.4.0

constantinpape:


constantinpape:
> I added the figures now @constantinpape. Let me know how they look!
> 
> PS. I realized I exported the figures in the default DPI, compared to previous screenshots - which probably were in higher DPI. Let me know if the resolution is not okay, I can increase the DPI easily!

Thanks! And the smaller DPI is ok, that should actually make things faster to load.

anwai98:
Looks good to me. Thanks @constantinpape!

PS. I made some tiny typo fixes here and there.

PR #976, opened 2025-03-28
Implement reproducibility functionality for commit file

constantinpape:


constantinpape:
Pending addition of `annotator` to `AnnotatorState` in #943

constantinpape:
Still missing:
- Reproduce commit for 3D segmentation
- Reproduce commit for tracking
- Continue annotation for tracking

The latter two require loading tracking results in the annotator, which is pending #1006 , so I will likely do this in a follow up commit.

Issue #945, opened 2025-03-18
micro_sam.automatic_segmentation: embeddings


aloboa:
I understand that, in micro_sam.automatic_segmentation,
```
  -e EMBEDDING_PATH, --embedding_path EMBEDDING_PATH
                        The path where the embeddings will be saved.
```
if embeddings do not exist, they are are created in that folder, but that if they exist, they are read from that folder

Correct?

constantinpape:
Yes, that is correct.

Note that you don't have to give that argument at all, in that case the embeddings will just be computed internally and will not be saved to a file.

aloboa:
> Note that you don't have to give that argument at all, in that case the embeddings will just be computed internally and will not be saved to a file.

My understanding is that embeddings must be calculated for each image (and model) and then the same embeddings can be used with different values of
`-pred_iou_thresh`,` --stability_score_thresh` and  `--box_nms_thresh`, to speed up the processing, correct?

BTW `-pred_iou_thresh`,` --stability_score_thresh` and  `--box_nms_thresh` are not listed by the -h option:

```
$ micro_sam.automatic_segmentation -h
usage: micro_sam.automatic_segmentation [-h] -i INPUT_PATH -o OUTPUT_PATH [-e EMBEDDING_PATH] [--pattern PATTERN] [-k KEY]
                                        [-m MODEL_TYPE] [-c CHECKPOINT] [--tile_shape TILE_SHAPE [TILE_SHAPE ...]]
                                        [--halo HALO [HALO ...]] [-n NDIM] [--mode MODE] [--annotate] [-d DEVICE] [-v]
```

constantinpape:
> My understanding is that embeddings must be calculated for each image (and model) and then the same embeddings can be used with different values of
> `-pred_iou_thresh`,` --stability_score_thresh` and `--box_nms_thresh`, to speed up the processing, correct?

Yes, that is correct.

> BTW `-pred_iou_thresh`,` --stability_score_thresh` and `--box_nms_thresh` are not listed by the -h option:

Yes, this is because they are part of keyword arguments that depend on whether AMG or AIS is used.

constantinpape:
I think the questions here are all resolved.

aloboa:
> I think the questions here are all resolved.

Yes, but please see https://forum.image.sc/t/specifying-embeddings-in-suite-of-micro-sam-automatic-segmentation-runs/110549. 
> 
[...]

Issue #906, opened 2025-02-28
Tracking integration


constantinpape:
We now have automated tracking support thanks to trackastra integration in #678 . To fully support / further improve this for the next release:
- [ ] Add `trackastra` to citation list in doc for users of automated tracking
- [ ] Get `trackastra` on conda-forge so that we can depend on it, see also https://github.com/weigertlab/trackastra/issues/25
- [ ] Enable filtering short tracks, see https://github.com/computational-cell-analytics/micro-sam/blob/dev/micro_sam/multi_dimensional_segmentation.py#L565
- [ ] Add a tracking layer to the tracking annotator in napari, see #7 

In addition we could:
- Add motile to dependencies if it's on conda-forge, see https://github.com/funkelab/motile/issues/98
- Look into trackastra performance issues for large timeseries (I will open a separate issue on this)

constantinpape:
The main parts are done, but there are still some improvements for the next release, see the unfinished checkmarks above.

PR #979, opened 2025-03-30
Updates to automatic segmentation

constantinpape:
To address #945 and (part of) #906

### DIFF
--- a/micro_sam/automatic_segmentation.py
+++ b/micro_sam/automatic_segmentation.py
@@ -1,8 +1,9 @@
 import os
+from functools import partial
 from glob import glob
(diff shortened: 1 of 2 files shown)


constantinpape:
This adds support for tracking for the automatic segmentation CLI and extends the description of the CLI to explain extra kwargs. Support for two features is still missing, but I would rather address them in a folow-up PR:
- Saving the tracking outputs; I would wait for #1006 and then save them in CTC format.
-  Continuing annotation. This requires initializing the tracking annotator with an existing tracking result, which is not yet implemented. I would also like to wait for #1006 with this, so that we can also read the CTC result here.

PR #989, opened 2025-04-06
Fix model export issues

anwai98:
WIP!


FynnBe:
The warnings are due to changes in pydantic. They are safe to ignore and disappear again with newer pydantic versions that we'll use for coming releases.

constantinpape (approved):

constantinpape:
Thanks for the clarification @FynnBe

PR #990, opened 2025-04-07
Update training logic to computes loss over logits

anwai98:
WIP!

### DIFF
--- a/micro_sam/training/joint_sam_trainer.py
+++ b/micro_sam/training/joint_sam_trainer.py
@@ -5,7 +5,6 @@ from collections import OrderedDict
 
 import torch
 from torch.utils.tensorboard import SummaryWriter
-from torchvision.utils import make_grid
 
 from .sam_trainer import SamTrainer
 
@@ -99,8 +98,7 @@ class JointSamTrainer(SamTrainer):
 
             with forward_context():
                 # 1. train for the interactive segmentation
-                (loss, mask_loss, iou_regression_loss, model_iou,
-                 sampled_binary_y) = self._interactive_train_iteration(x, labels_instances)
+                loss, mask_loss, iou_regression_loss, model_iou = self._interactive_train_iteration(x, labels_instances)
 
             backprop(loss)
 
@@ -114,9 +112,8 @@ class JointSamTrainer(SamTrainer):
 
             if self.logger is not None:
                 lr = [pm["lr"] for pm in self.optimizer.param_groups][0]
-                samples = sampled_binary_y if self._iteration % self.log_image_interval == 0 else None
                 self.logger.log_train(
-                    self._iteration, loss, lr, x, labels_instances, samples,
+                    self._iteration, loss, lr, x, labels_instances,
                     mask_loss, iou_regression_loss, model_iou, unetr_loss
                 )
 
@@ -147,7 +144,7 @@ class JointSamTrainer(SamTrainer):
                 with forward_context():
                     # 1. validate for the interactive segmentation
                     (loss, mask_loss, iou_regression_loss, model_iou,
-                     sampled_binary_y, metric) = self._interactive_val_iteration(x, labels_instances, val_iteration)
+                     metric) = self._interactive_val_iteration(x, labels_instances, val_iteration)
 
                 with forward_context():
                     # 2. validate for the automatic instance segmentation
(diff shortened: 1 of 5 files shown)


anwai98:
Hi @constantinpape,

This should be good for a first review from my side!

The intended idea here is simple: 1) compute loss over logit masks (to make this possible, I convert the labels to match the logits dimensions) and 2) work with iterative prompts for logits

Let me know how it looks!

micro_sam/training/sam_trainer.py
```diff
@@ -240,11 +235,11 @@ def _compute_iterative_loss(self, batched_inputs, y_one_hot, num_subiter, multim
...
         return loss, mask_loss, iou_regression_loss, mean_model_iou
 
-    def _update_prompts(self, batched_inputs, y_one_hot, masks, logits_masks):
+    def _update_prompts(self, batched_inputs, y_one_hot, logits_masks):
         # here, we get the pair-per-batch of predicted and true elements (and also the "batched_inputs")
-        for x1, x2, _inp, logits in zip(masks, y_one_hot, batched_inputs, logits_masks):
+        for y_, _inp, logits in zip(y_one_hot, batched_inputs, logits_masks):
             # here, we get each object in the pairs and do the point choices per-object
-            net_coords, net_labels, _, _ = self.prompt_generator(x2, x1)
+            net_coords, net_labels, _, _ = self.prompt_generator(segmentation=y_, prediction=logits)
```
anwai98, line 242: I observed one thing instantly in one of the training runs: small objects get smaller, hence it throws errors for not being able to find pixels to fetch points during iterative prompting. I will investigate this closely and see what can be done.

Otherwise, we might have to stick to full masks for iterative prompting!

micro_sam/training/joint_sam_trainer.py
```diff
@@ -175,25 +172,22 @@ class JointSamLogger(TorchEmLogger):
...
-        self.log_dir = f"./logs/{trainer.name}" if save_root is None else\
-            os.path.join(save_root, "logs", trainer.name)
+        self.log_dir = f"./logs/{trainer.name}" if save_root is None else os.path.join(save_root, "logs", trainer.name)
         os.makedirs(self.log_dir, exist_ok=True)
 
         self.tb = SummaryWriter(self.log_dir)
         self.log_image_interval = trainer.log_image_interval
 
-    def add_image(self, x, y, samples, name, step):
+    def add_image(self, x, y, name, step):
```
constantinpape, line 181: Was this never used / not used anymore?

micro_sam/training/sam_trainer.py
constantinpape: But fundamentally these must still be there. Because it just does a naive upsampling of the low-res mask in SAM from what I remember, so it cannot add labels that where not there in the low-res mask?!
Maybe we just need to adapt some size threshold?

micro_sam/training/joint_sam_trainer.py
anwai98: It was used previously, I removed it now because the `samples` are logits now. Would you recommend to keep those predictions by upsampling them?

micro_sam/training/sam_trainer.py
anwai98: Okay, even after a size sampler in place, the issue originates from the spot for finding positive point prompts in iterative prompting for a few objects. I will try to investigate what's the core reason behind it to get a better idea!

micro_sam/training/joint_sam_trainer.py
constantinpape: What exactly were the samples? Examples for mask predictions? It would maybe be good to keep them, we can discuss later.

anwai98: Yes, samples were mask predictions!

micro_sam/training/sam_trainer.py
anwai98: This issue is more or less identified and probably is fixed with the latest commit (I will push a test training run anyways). For now, I'll resolve this discussion!

anwai98:
Here's a detailed description on the PR (we see when we can come back to this in future)

Core Idea: Computing loss over predicted logits masks and downsampled ground-truth masks.

The current implementation works, however it does not bring us significant memory advantages (below mentioned are quick try-outs on LIVECell):

```
# # for default settings
# loss over logits: 47.89 GB
# loss over masks: 49.42 GB

# # for resource-efficient setting (i.e. n_objects=5 for 'vit_b')
# loss over logits: 24.01 GB
# loss over masks: 24.38 GB
```

PR #991, opened 2025-04-07
Better handling of input prompts in first iteration

anwai98:
WIP!

### DIFF
--- a/micro_sam/training/sam_trainer.py
+++ b/micro_sam/training/sam_trainer.py
@@ -195,14 +195,7 @@ class SamTrainer(torch_em.trainer.DefaultTrainer):
         masks = (masks > 0.0).float()
         return masks, logits
 
-    def _compute_iterative_loss(self, batched_inputs, y_one_hot, num_subiter, multimask_output):
-        """Compute the loss for several (sub-)iterations of iterative prompting.
-        In each iterations the prompts are updated based on the previous predictions.
-        """
-        image_embeddings, batched_inputs = self.model.image_embeddings_oft(batched_inputs)
-
-        loss, mask_loss, iou_regression_loss, mean_model_iou = 0.0, 0.0, 0.0, 0.0
-
+    def _use_mask_inputs(self, batched_inputs, y_one_hot):
         # Whether to use masks per training top-iteration.
         use_mask_inputs = False  # determines if each sub-iteration will use mask inputs as prompts or not.
         use_zero_mask = False  # determines if the zeroth iteration will use zeros as mask inputs.
@@ -228,7 +221,7 @@ class SamTrainer(torch_em.trainer.DefaultTrainer):
                 use_mask_inputs = bool(use_mask_inputs_tensor.item())
                 use_zero_mask = use_mask_inputs  # provides zeros as mask inputs.
             else:  # training on a single GPU.
-                use_mask_inputs = random.random() > self.mask_prob
+                use_mask_inputs = None
 
         if use_zero_mask:
             # We use zeros as mask inputs for the zeroth iteration.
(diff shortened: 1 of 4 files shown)


anwai98:
Okay, I can reproduce the error with `mask_prob=0.5` now.

```
Parameter at index 190 with name sam.prompt_encoder.mask_downscaling.6.weight has been marked as ready twice. This means that multiple autograd engine  hooks have fired for this particular parameter during this iteration.
```

DDP does not like this for some reason. I will investigate this now!

anwai98:
Hi @constantinpape,

As discussed, this PR is ready as per our discussion. I checked it on multi-GPU setups, and works as expected!

Let me know how this looks. I will run a few training checks with it on LIVECell for now!

micro_sam/training/sam_trainer.py
```diff
@@ -203,6 +203,41 @@ def _compute_iterative_loss(self, batched_inputs, y_one_hot, num_subiter, multim
...
+            if is_data_parallel:  # if training on multiple GPUs.
+                if torch.distributed.get_rank() == 0:  # device with rank 0.
+                    use_mask_inputs_tensor = torch.tensor(
+                        random.random() > self.mask_prob, dtype=torch.uint8, device=self.device,
+                    )
+                else:  # on other devices, we do not need this parameter at this stage.
+                    use_mask_inputs_tensor = torch.tensor(0, dtype=torch.uint8, device=self.device)
+
+                # Broadcast the value to all devices (ranks).
+                torch.distributed.broadcast(use_mask_inputs_tensor, src=0)
```
constantinpape, line 225: Does this imply that we wait here for the variable to be available across iterations?
(I guess you checked / if the tests work then this means that it is correct.)

constantinpape (commented):
Looks good to me! Let's wait to see if the training is working, and then I could merge this.

micro_sam/training/sam_trainer.py
anwai98: Ahha yes I forgot to address this. Yes, it indeed means that the "source" of this variable is from one particular device!

anwai98:
hi @constantinpape,

As discussed yesterday, I have finalized this to keep our old single GPU heuristic for using mask inputs with a probability. GTG from my side!

constantinpape (commented):
I think this is exactly what we discussed, but we need to delineate the two different cases of the mask usage a bit better in the code. It's maybe enough to add a few more comments, but right now I tend to separating it out a bit more. Let's discuss tomorrow.

micro_sam/training/sam_trainer.py
```diff
@@ -195,6 +196,43 @@ def _get_best_masks(self, batched_outputs, batched_iou_predictions):
...
+
+        if self.mask_prob == 1:  # i.e. always use masks.
+            use_mask_inputs = True  # we would like to use mask inputs in all sub-iterations.
+            use_zero_mask = self.is_data_parallel  # we would like to use zeros as mask inputs for zeroth iteration.
+
+        elif self.mask_prob > 0:  # i.e. if we use mask inputs with a probability.
+            if self.is_data_parallel:  # if training on multiple GPUs.
+                if torch.distributed.get_rank() == 0:  # device with rank 0.
+                    use_mask_inputs_tensor = torch.tensor(
+                        random.random() > self.mask_prob, dtype=torch.uint8, device=self.device,
```
constantinpape, line 212: The sign of the comparison is different from line 309. I think the one in 309 is correct, but please double check.

micro_sam/training/sam_trainer.py
```diff
@@ -260,13 +301,21 @@ def _update_prompts(self, batched_inputs, y_one_hot, masks, logits_masks):
...
-                if use_mask_inputs:
-                    _inp["mask_inputs"] = logits
-                else:  # remove  previously existing mask inputs to avoid using them in next sub-iteration
-                    _inp.pop("mask_inputs", None)
+            if self.is_data_parallel:  # multi-GPU training
+                use_mask_inputs_this_iter = use_mask_inputs
+            else:  # single GPU training
+                if self.mask_prob > 0:
+                    # using mask inputs for iterative prompting while training, with a probability
+                    use_mask_inputs = (random.random() < self.mask_prob)
```
constantinpape, line 309: Check previous comment.

constantinpape (commented):
The overall logic should now be correct and is more clear, but there is an inconcistency in the random mask selection logic.

micro_sam/training/sam_trainer.py
anwai98: Yes, you are right. It's actually safe to stay consistent while comparing with the generator in two setups. Thanks for spotting this!

(I have fixed this in a new commit)

PR #1008, opened 2025-04-18
Implement instance segmentation training and update training tests

constantinpape:


constantinpape:
Hi @anwai98 ,
this implements the instance segmentation only training for https://forum.image.sc/t/micro-sam-training-on-negative-data/111248 . I think it's all good, and I also wrote a test and ran it along with the original training test to check that everything works (these tests are still hanging in CI, but they work for me locally). Still, could you check if this looks good from your end?

micro_sam/training/training.py
```diff
@@ -430,6 +603,24 @@ def default_sam_dataset(
...
+                "If 'train_instance_segmentation_only' is True, then 'with_segmentation_decoder' must also be True."
+            )
+        return_instances = False
+
+    # If a sampler is not passed, then we set a MinInstanceSampler, which requires 3 distinct instances per sample.
+    # This is necessary, because training for interactive segmentation does not work on 'empty' images.
+    # However, if we train only the automatic instance segmentation decoder, then this sampler is not required
+    # and we do not set a default sampler.
+    if sampler is None and not train_instance_segmentation_only:
+        sampler = torch_em.data.sampler.MinInstanceSampler(3, min_size=min_size)
```
anwai98, line 622: I would relax this a bit more, to keep the default sampler to work with `min_num_instances=2`

anwai98 (approved):
Thank you so much for the implementation @constantinpape!

I left a minor comment below on relaxing the sampler criterion, otherwise all is pretty good to merge from my side!

PS. I have some ideas on improving the design for passing custom loss functions, etc, to the instance segmentation training pipeline, but let's tackle it in another PR!

micro_sam/training/training.py
constantinpape, line 622:
```suggestion
        sampler = torch_em.data.sampler.MinInstanceSampler(2, min_size=min_size)
```

PR #1016, opened 2025-04-25
Minor updates to quick start notebook

anwai98:
This PR takes care of syncing up the `quick_start.ipynb` with other notebook format and other necessary tool info for users!

GTG from my side!

PS. @haesleinhuepf let us know if all looks good to you as well (I didn't change anything in the `stackview` part. Let me know if you have any suggestions regardless.)


haesleinhuepf:
Hi @anwai98 ,

I'd say this is no longer a quick start notebook. In fact, it is not required to use condacolab oder google colar to use micrso-sam. Hence, I recommend removing the first page of blabla from the quick start notebook again and keep things as simple as possible. Otherwise it's deterrend for beginners.

On the other hand it's your repository. You make the rules here. ;-)

Have a great weekend!

Best,
Robert

anwai98:
Hi @haesleinhuepf,

Thank you so much for the feedback!

I do agree that the first page of the notebook now has a lot of information. The intention for that is to ensure that users have all the flexibility (and hopefully read as much relevant context as possible to get properly started), with the least amount of maintenance effort for me! (and for most new users who don't have much technical experience, I route them all to get started with our notebooks on Kaggle Notebooks)

PS. I do like the proposal of keeping things simple. I now added a python script under the `examples/` folder for users to get jump started ;)

This is finalized from my side now. This awaits a review from @constantinpape, and is GTM from my end!

constantinpape:
Just quickly from my side without checking the changes: is there a way to make the sections specific to the installation on colab / kaggle collapsible? This way we would still have the info present, but it wouldn't overwhelm users in the beginning.

constantinpape:
I had a chance to check this out now, and I really like having the collapsible section! Let's do the same for all the notebooks, this makes it much easier to use them locally if `micro_sam` is installed, but keeps everything consistent and enables usage on Kaggle as well.

anwai98:
Hi @constantinpape,

This is GTG from my side. I suggest we go ahead and merge this.

In case you wanna visualize the drop-down, here's one of our other notebooks:
- https://colab.research.google.com/github/computational-cell-analytics/micro-sam/blob/update-quick-start-nb/notebooks/sam_finetuning.ipynb

PR #1017, opened 2025-04-29
Update automatic documentation deployment script

anwai98:
This PR is inspired by https://github.com/computational-cell-analytics/synapse-net/blob/main/.github/workflows/build_docs.yaml to try whether automatic documentation build works now or not!

@constantinpape what do you think about this?


constantinpape:
I will go ahead and close this --- Claude will take care of it.

PR #1019, opened 2025-05-02
Remove napari pin to allow v6 supports

anwai98:
WIP!

The installation from `conda-forge` seems to create some problems for me locally, but let's see how the tests behave for now!

### DIFF
--- a/environment.yaml
+++ b/environment.yaml
@@ -5,7 +5,7 @@ dependencies:
     - nifty >=1.2.3
     - imagecodecs
     - magicgui
-    - napari >=0.5.0,<0.6.0
+    - napari
     - natsort
     - pip
     - pooch


anwai98:
Okay, seems like the tests are passing on a higher level. We still need to double check all GUI stuff closely to be super sure!

anwai98:
I checked stuff just now. Seems like all works fine for me via GUI!

The installation is slightly tricky with our current pins, but an easy way to get all latest stuff for testing is:

```bash
micromamba create -n test-neu -c conda-forge python=3.11 micro_sam
micromamba install napari=0.6 napari-base=0.6 --no-deps
```

And this should give all expected latest packages (`pyopengl=3.1.9`, `micro_sam=1.4`, `napari=0.6` and latest `pyqt` build)

constantinpape (approved):

psobolewskiPhD:
I think once https://github.com/conda-forge/napari-feedstock/pull/81 is merged your normal installation should work 🤞

PR #1021, opened 2025-05-05
Remove verbosity argument from learning rate schedulers

anwai98:
This PR removes the `verbose` argument from the learning rate schedulers as it is depreciated in `pytorch>=2.7`, and the latest pytorch is up on `conda-forge` since last week!

GTG from my side!

### DIFF
--- a/micro_sam/training/training.py
+++ b/micro_sam/training/training.py
@@ -280,7 +280,7 @@ def train_sam(
         optimizer = optimizer_class(model_params, lr=lr)
 
         if scheduler_kwargs is None:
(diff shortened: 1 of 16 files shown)


micro_sam/training/training.py
anwai98, line 1: I think this is the most important spot to take care of, however I updated all scripts in reach with this argument.

anwai98:

PS. @constantinpape I kept this PR on `master` as users would get the latest `pytorch` now onwards, and the training would break for them, as `pytorch` simply remove the argument from the signature at all.

PR #1023, opened 2025-05-05
Minor update to all docstrings

anwai98:
WIP!


anwai98:
Okay, this looks good now. The reason for doing is because I sometimes struggle with remembering the parameters, what is the default behaviour, what are the expected choices, etc.

This def. helps for now!

I will go ahead and merge this once the tests pass!

PR #1036, opened 2025-05-16
First working prototype for object classifier

constantinpape:


constantinpape:
Hi @anwai98 and @caroteu ,
this is a first working prototype for the object classification with SAM embeddings. The UI can be improved, and we should look at the feature settings etc. based on Caro's findings, but it is ready to be tested.

You can find examples for how to start it in `examples/object_classifier.py`. I have uploaded the sample data for it [here](https://owncloud.gwdg.de/index.php/s/i7ccOmQUpw9tJUJ).

anwai98:
Hi @constantinpape,

Thanks for the prototype. Works pretty good for me so far.

I found a tiny issue though while trying this on histopatho data (you've commented on this already, I'll briefly explain it in a code comment). Besides that, looks good to get started (there are some improvement ideas I have, we can chat about them next week)!

PS. I will push the tiny new changes on top of this via a new PR.

micro_sam/sam_annotator/object_classifier.py
```diff
@@ -0,0 +1,485 @@
...
+
+    # Resize the segmentation and embeddings to be of the same size.
+    # For now we resize the segmentation to the embedding size.
+    # Note: this is more efficient, but we may loose small objects.
+    # Maybe we first resize the embeddings to something intermediate, like 256 x 256?
+
+    # embeddings = resize(embeddings, (256, 256, embeddings.shape[-1]), preserve_range=True).astype(embeddings.dtype)
+    segmentation_rescaled = resize(
+        segmentation_rescaled, embeddings.shape[:2], order=0, anti_aliasing=False, preserve_range=True
+    ).astype(segmentation.dtype)
```
anwai98, line 54: The need to resize the embeddings, otherwise we do loose small objects, which breaks the training pipeline due to computed feature vs label mismatch!

micro_sam/sam_annotator/object_classifier.py
```diff
@@ -115,14 +117,31 @@ def generator():
     return generator, length
 
 
-def _compute_object_features(image_embeddings, segmentation, verbose=True):
+def compute_object_features(
+    image_embeddings: util.ImageEmbeddings,
+    segmentation: np.ndarray,
+    resize_embedding_shape: Tuple[int, int] = (256, 256),
```
constantinpape, line 123: I exposed this as a parameter @anwai98

micro_sam/sam_annotator/object_classifier.py
```diff
@@ -293,7 +312,7 @@ def _require_layers(self, layer_choices: Optional[List[str]] = None):
             if image_scale is not None:
                 self._viewer.layers["annotations"].scale = image_scale
             # Reduce the brush size and set the default mode to "paint" brush mode.
-            annotation_layer.brush_size = 1
+            annotation_layer.brush_size = 3
```
constantinpape, line 315: I think 1 is too aggressive here. With this the annotations are not really visible for larger images. I think 3 is a good default.

constantinpape:
@anwai98 : I worked on this a bit more, and I have added:
- A first version for an image series annotator. This is both for convenience, but also enables training a classifier on multiple images.
- A first implementation for batch prediction.

For both, I have added examples to `examples/object_classifier`; for the former updating your histopatho example.

micro_sam/sam_annotator/object_classifier.py
anwai98: I agree. Makes sense!

anwai98:
Hi @constantinpape,

I briefly checked the implementation. Works as expected (and the setup looks pretty cool). Thanks!

anwai98:
I pushed a few new changes, only to the example scripts (to make the `annotator_devel` function working to train RF model programatically)

I noticed one tiny detail:
- When an image and corresponding labels are added directly via the GUI, the `Segmentation` widget does not consider new label layers!

PR #1037, opened 2025-05-17
Minor updates to object classification prototype

anwai98:
This PR allows resizing of embeddings, and adds example script for histopathology data. GTM from my side!


anwai98:
I also made some new changes just now:
- Reduced the default brush size and set the paint mode to default in `annotation` layer
- Allowing recreation of layers after they are deleted

constantinpape (approved):

PR #1043, opened 2025-05-25
Remove context manager for zarr.open

constantinpape:


anwai98:
Ayyyy I see the green tick. Thank you so much @constantinpape! Should we merge this already?

constantinpape:
Yes!

PR #1047, opened 2025-05-28
Fix: Handle missing seg_ids after resizing

caroteu:
After resizing in `_compute_object_features_impl`, some small instances are dropped. Thus the seg_ids list could become non-contiguous (e.g., [1, 2, 4, 5] because instance 3 was removed). This results in an error when mapping the predictions to the segmentation, since some instances have no corresponding prediction.

I added a quick fix, that simply sets the predictions of these missing instances to 0.

An alternative solution could be to introduce a min_size argument, which would proactively filter out small instances before resizing. This would make the filtering explicit and more transparent to the user.

@constantinpape @anwai98

### DIFF
--- a/micro_sam/object_classification.py
+++ b/micro_sam/object_classification.py
@@ -210,7 +210,14 @@ def project_prediction_to_segmentation(
     """
     assert len(object_prediction) == len(seg_ids)
     prediction = {seg_id: class_pred for seg_id, class_pred in zip(seg_ids, object_prediction)}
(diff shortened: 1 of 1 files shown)


constantinpape:
Good catch @caroteu ! But this triggered my code OCD a bit ;). I added a more pythonic version of the fix in #1047 .
I will merge that one once tests pass. Let me know if you run into other issues with the export.

PR #1060, opened 2025-06-08
Update documentation for BAND

anwai98:
WIP!

Will update the images and some other tiny stuff!


anwai98:
BAND instructions have been updated now. I will leave the data transfer video as it is (via copy-paste method over the virtual desktop).

And this is GTG from my side!

PR #1063, opened 2025-06-13
Allow support for custom label transforms in default_sam_dataset

lufre1:
If custom label_transform is needed it now can be added to the default_sam_dataset without interfering with the existing label transform.

### DIFF
--- a/micro_sam/training/training.py
+++ b/micro_sam/training/training.py
@@ -668,31 +668,25 @@ def default_sam_dataset(
     if raw_transform is None:
         raw_transform = require_8bit
 
+    # Prepare the label transform.
     if with_segmentation_decoder:
-        label_transform_in = kwargs.pop("label_transform", None)
-        if label_transform_in is None:
-            label_transform = torch_em.transform.label.PerObjectDistanceTransform(
-                distances=True,
-                boundary_distances=True,
-                directed_distances=False,
-                foreground=True,
-                instances=return_instances,
-                min_size=min_size,
-            )
-        else:
-            label_transform = torch_em.transform.generic.Compose(
-                label_transform_in,
-                torch_em.transform.label.PerObjectDistanceTransform(
-                    distances=True,
-                    boundary_distances=True,
-                    directed_distances=False,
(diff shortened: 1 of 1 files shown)


micro_sam/training/training.py
```diff
@@ -669,14 +669,28 @@ def default_sam_dataset(
...
+                boundary_distances=True,
+                directed_distances=False,
+                foreground=True,
+                instances=return_instances,
+                min_size=min_size,
+            )
+        else:
+            label_transform = torch_em.transform.generic.Compose(
+                label_transform_in,
+                torch_em.transform.label.PerObjectDistanceTransform(
```
constantinpape, line 685: Don't duplicate all the code for creating the `PerObjectDistanceTransform`. The clean way to do this is:
```python
label_transform = torch_em.transform.label.PerObjectDistanceTransform(...)
label_transform_in = kwargs.pop("label_transform", None)
if label_transform_in is not None:
    label_transform = torch_em.transform.generic.Compose(label_transform_in, label_transform)
```

micro_sam/training/training.py
```diff
@@ -669,14 +669,28 @@ def default_sam_dataset(
...
+                    distances=True,
+                    boundary_distances=True,
+                    directed_distances=False,
+                    foreground=True,
+                    instances=return_instances,
+                    min_size=min_size,
+                )
+            )
     else:
         label_transform = torch_em.transform.label.MinSizeLabelTransform(min_size=min_size)
```
constantinpape, line 695: The same logic as above should be used here. I.e. combine the `label_transform_in` with the MinSizeLabelTransform. Otherwise the behavior becomes inconsistent.

constantinpape (changes requested):
Makes sense, but please improve the implementation, see the comments.

anwai98:
I took care of this, since we are ready in preparation for the next release already.

Let me know how it looks @constantinpape @lufre1

constantinpape (approved):
Thanks @anwai98 ! Now it is consistent and avoids redundancies.

PR #1072, opened 2025-06-22
Minor update for installer release

anwai98:
WIP!


anwai98:
I'll drop this. We have other plans as to how we are gonna approach this in future!

PR #1092, opened 2025-08-06
Add is_multi_tensor argument to default_sam_dataset to split kwargs in loader

lufre1:
Hello again :D
i found a better way to alter param "is_multitensor" in default sam dataset by adding param to the function header. 
This way we utilize split_kwargs function: 
split_kwargs(default_sam_dataset, **kwargs)

and we do not have to use kwargs seperately. 

@anwai98


anwai98 (approved):
Thanks @lufre1!

Looks good to me. I will go ahead and merge this once the tests pass!

PR #1096, opened 2025-08-12
Minor fix to additional arguments for training instance segmentation

anwai98:
Hi @lufre1,

I realized that the argument you spotted doesn't really exist in the function, which is why the issue was happening. Removing that lead to more issues (because it got some SAM-related parameters for interactive segmentation training heuristics). Hence, I make the checks exclusively now for expected additional arguments. 

Could you check if this branch works now?

If yes, I'll merge this then!


anwai98:
Awesome. Thanks for checking it out @lufre1!

Issue #1094, opened 2025-08-11
Update `get_sam_model` design to allow passing decoder weights


anwai98:
I've been seeing this that people want to keep pretrained weights somewhere else (due to system / admin / memory, etc., limitations) and currently our design limits using two checkpoint paths of generalist models to be loaded together!

I'll work on making it modular and open a PR. Let's see if I can keep the design minimalistic! (because there is quite some conditions we need to consider there)

anwai98:
Ah good that the issue is still active.

The primary purpose to allow passing decoder weights is to support older generalist models!

anwai98:
I think we took care of this. This is sorted now! ;)

PR #1101, opened 2025-08-27
Expose decoder path to `get_sam_model`

anwai98:
This PR adds an additional argument `decoder_path` to `get_sam_model`.

I need to test it and see if it works as intended for all parts (i.e. both inference and training). I'll set it up for review once it's ready!

And closes https://github.com/computational-cell-analytics/micro-sam/issues/1094

### DIFF
--- a/micro_sam/util.py
+++ b/micro_sam/util.py
@@ -317,6 +317,7 @@ def get_sam_model(
     peft_kwargs: Optional[Dict] = None,
     flexible_load_checkpoint: bool = False,
     progress_bar_factory: Optional[Callable] = None,
+    decoder_path: Optional[Union[str, os.PathLike]] = None,
     **model_kwargs,
 ) -> SamPredictor:
     r"""Get the Segment Anything Predictor.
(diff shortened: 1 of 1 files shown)


anwai98:
Well, this one is one of those "oops" PRs. Let's discuss this together @constantinpape!

PS. I say this because if we decide to go for it, we may have to change quite some things here and there! (which is fine for me, but we should chat about the relevance of this first)

anwai98:
Okay, I think I am past the "oops" ladder here hehe. I decided to add support for `decoder_path` only via CLI commands (adding it to the tool would be a bit of work to me).

Let me know how it looks @constantinpape

constantinpape (commented):
This makes sense from a technical point of view, but I am not sure if it makes it easier for users to have another argument. Let's discuss.

anwai98:
We decided to opt for a more helpful way to document the ways to access models in a system without internet access, without making several changes in the API. I will go ahead and close this for now!

PR #1126, opened 2025-12-13
Fix training/validation metrics scaling and validation loss aggregation

YonghaoZhao722:
This PR fixes two issues in training and validation metrics logging that affect comparability and correctness.

### 1. Training loss / metric scaled with batch size

In `SamTrainer._compute_loss()`, per-sample `mask_loss` and `iou_regression_loss` were accumulated across the batch without normalization.  
As a result, increasing `batch_size` inflated the reported loss and metric values even when model quality was unchanged, making runs hard to compare.

### 2. Validation `mask_loss` / `iou_loss` were not epoch averages

During validation, `validation/metric` and `validation/loss` were averaged over `len(val_loader)`, but `validation/mask_loss` and `validation/iou_loss` effectively reflected only the **last validation batch** rather than epoch-level averages.

## Changes made

- Normalize `mask_loss` and `iou_regression_loss` by `batch_size` inside `_compute_loss()` so training metrics are batch-size invariant.
- Accumulate and log **epoch-averaged** `validation/mask_loss` and `validation/iou_loss`.
- Apply the same aggregation logic to the joint trainer for consistency.

## Why this matters

- Makes training curves **comparable across different batch sizes**, which is important for scaling experiments.
- Ensures TensorBoard validation scalars accurately represent **epoch-level averages**.

This change does not affect model behavior or optimization—only how metrics are computed and reported.

### DIFF
--- a/workshops/i2k_2024/README.md
+++ b/workshops/i2k_2024/README.md
@@ -38,7 +38,7 @@ You can create a new environment and install it like this:
 ```bash
 conda create -c conda-forge -n micro_sam python=3.11 natsort
 conda activate micro_sam
-conda install -c pytorch -c conda-forge "micro_sam>=1.1" "pytorch>=2.4" "protobuf<5" cpuonly
+conda install -c conda-forge "micro_sam>=1.1" "pytorch>=2.4"
 ```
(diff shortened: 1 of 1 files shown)


anwai98 (approved):
Looks good to me. I'll merge this in a few minutes. Thanks for your contribution!

PR #1127, opened 2025-12-14
Refactor APG interface and instance segmentation tests

constantinpape:


constantinpape:
Thanks for checking this out @anwai98 . I think that the more efficient segmentation is working now. I will quickly check the result tomorrow and merge this if things look as expected.

constantinpape:
The tiling result looks very good:

image

I will merge it despite the test failure, I think that one is a one-off.

PR #1139, opened 2026-01-13
Test claude code ...

constantinpape:
by implementing 'Segment Slice' for SAM2 in the 3D a…nnotator (not yet checked)


constantinpape (commented):
I will clean up some things, merge the current upstream, check if single frame works, check if multi-prompt works, and then merge.

Anwai will work on improving the state initialization with precomputed video embeddings, which is currently not ideal because it recomputes a lot.

PR #1149, opened 2026-01-28
Enable precomputing and loading image embeddings

anwai98:
This PR enables precomputation and loading of image embeddings for the 3d promptable segmentation. There are some `attrs` missing to plug in storing and checking the `zarr` file, but the promptable segmentation experience is significantly better with SAM2 models now!

PS. Will work on aligning the check embeddings step now. Should be good to test without storing it for now!

### DIFF
--- a/micro_sam/sam_annotator/_widgets.py
+++ b/micro_sam/sam_annotator/_widgets.py
@@ -472,7 +472,8 @@ def clear_volume(viewer: "napari.viewer.Viewer", all_slices: bool = True) -> Non
         i = int(viewer.dims.point[0])
         vutil.clear_annotations_slice(viewer, i=i)
 
-    # If it's a SAM2 promptable segmentation workflow, we should reset the prompts after clear annotations has been clicked.
+    # If it's a SAM2 promptable segmentation workflow,
+    # we should reset the prompts after clear annotations has been clicked.
     if state.interactive_segmenter is not None:
         state.interactive_segmenter.reset_predictor()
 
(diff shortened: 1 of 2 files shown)


micro_sam/v2/models/_video_predictor.py
```diff
@@ -118,74 +118,128 @@ class CustomVideoPredictor(SAM2VideoPredictor):
...
     def init_state(
         self,
-        video_path: Optional[str],
-        volume: Optional[np.ndarray] = None,
-        offload_video_to_cpu=False,
-        offload_state_to_cpu=False,
-        async_loading_frames=False,
-        verbosity=True,
+        volume: np.ndarray,
+        volume_embeddings: Optional[np.ndarray] = None,
```
constantinpape: In the old `micro_sam` design we assumed that we always have precomputed embeddings, because they are either computed when the annotator is started or loaded from cached embeddings. This means, `volume_embeddings` would not be optional here. Maybe we want to stick to this design? I think that it makes things more simple.

micro_sam/sam_annotator/_widgets.py
```diff
@@ -1094,8 +1094,7 @@ def segment_slice(viewer: "napari.viewer.Viewer") -> None:
 
         # Create a PromptableSegmentation3D instance with existing inference state
         seg_handler = PromptableSegmentation3D(
-            predictor=state.predictor,
-            volume=volume,
+            predictor=state.predictor, volume=volume, volume_embeddings=state.image_embeddings,
```
constantinpape: It seems redundant to re-create this class every time. It should rather be initialized ones and then be part of either the `state` or the annotator class, whatever is more convenient.

micro_sam/v2/models/_video_predictor.py
anwai98: Ahha that makes my life definitely easier ;) (I'll refactor it then)

micro_sam/v2/models/_video_predictor.py
```diff
@@ -118,74 +118,128 @@ class CustomVideoPredictor(SAM2VideoPredictor):
...
+        device: Optional[Union[str, torch.device]] = None,
+        offload_video_to_cpu: bool = False,
+        offload_state_to_cpu: bool = False,
+        async_loading_frames: bool = False,
+        verbosity: bool = True,
+        ignore_caching_features: bool = False,
```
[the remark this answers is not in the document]
anwai98: This is needed in `precompute_image_embeddings` to warm up the expected parts in `inference_state`, to compute the image embeddings (and other relevant stuff of the backbone) using the `predictor`.

micro_sam/sam_annotator/_widgets.py
anwai98: Yes. I thought of moving it to `state` and call it as the `state.segmenter`. I wanted to decouple that update with the embeddings stuff (which is why I left your comment there still ;))

constantinpape: I would do it right away. Putting it into state is fine, but I think calling it `segmenter` is a bit too vague (since this could also be automatic segmentation). Maybe `interactive_segmenter`?

constantinpape (commented):
You forgot to use `state.instance_segmenter` in one place. Looks good otherwise.

PR #1150, opened 2026-01-29
Add scripts for APG statistical experiments

anwai98:
This PR adds scripts for storing results per image for us to be able to do some statistical tests. It'll have the scripts soon for the stat-tests later. I'll keep this open!

@constantinpape I'll share the per-image quantitative results in a few minutes!


scripts/apg_experiments/run_statistical_comparisons.py
constantinpape: I am not sure it makes sense to have new code for running this. Isn't it better to just use the old scripts and save the results in these? I can concatenate them later.

PR #1186, opened 2026-04-01
Update evaluation scripts

anwai98:
This PR aims to update evaluation scripts to adopt `micro_sam2` backbone. This is gonna be a long WIP, so I'll take care of things one by one here itself! ;)


anwai98:
My quick thought here is to go ahead and merge this. I can do a solid refactor later for other things!

anwai98:
Since the parent working branch (https://github.com/computational-cell-analytics/micro-sam/pull/1196) got merged to `dev`: I think the safest move is to bring everything to `dev` and fix the annotator tool there itself. What do you think? @constantinpape (any objection?)

constantinpape:
Yes, let's go ahead with this. (We should just make sure not to merge anything related to SAM2 onto `main` yet)

anwai98:
Bingo. We should stick to our current `dev` plans! (also take care of keeping `dev` upstream to any `main` changes)

I'll go ahead and merge this then!

PR #1193, opened 2026-04-23
Fix uint8 image interpolation failure on MPS devices in PredictorAdaptor

anwai98:
Minor changes to make this work for macOS (some weird interpolation errors). Let's see if the macOS CIs pass now!

I'll merge this only when the macOS CIs pass!


anwai98:
Yayy the tests pass. I'll go ahead and squash this!

PR #1194, opened 2026-04-23
Fix image loading and saving logic in image_series_annotator

maartenpaul:
Related to this issue: https://github.com/computational-cell-analytics/micro-sam/issues/1174

> When running the image_series_annotator, the first image in the series incorrectly reports that embeddings have not been calculated, even when they have been precomputed for all images. Rerunning the embedding computation resolves it temporarily, but if an embedding_path is set, it throws an error until the embedding path is removed.

> The root cause seem to be in the _load_image helper function, which accepts an image_id parameter but ignores it — instead reading next_image_id from the outer scope.

> A related issue exists in _save_segmentation, which similarly ignores its current_idx parameter in favour of the outer next_image_id.

### DIFF
--- a/micro_sam/sam_annotator/image_series_annotator.py
+++ b/micro_sam/sam_annotator/image_series_annotator.py
@@ -209,10 +209,10 @@ def image_series_annotator(
         return os.path.join(output_folder, fname)
 
     def _load_image(image_id):
-        image = images[next_image_id]
+        image = images[image_id]
         if not have_inputs_as_arrays:
             image = imageio.imread(image)
-        image_embedding_path = embedding_paths[next_image_id]
+        image_embedding_path = embedding_paths[image_id]
         return image, image_embedding_path
 
     # Check which image to load next if we skip segmented images.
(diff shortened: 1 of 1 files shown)


anwai98:
Hi @maartenpaul,

Thanks for the PR. Can you give me a before and after case showing how this fixes #1174?

maartenpaul:
Hi @anwai98 , looking more carefully it seems that although the proposed code changes are probably useful it seems they are not directly related to the issue. 

Surprisingly though I cannot reproduce that issue any more (tried in several ways) so likely it was caused by something else. You can decide whether you want accept this PR or just close it. I guess we can also close the issue until the issue might reappear.

anwai98:
Hi @maartenpaul,

> looking more carefully it seems that although the proposed code changes are probably useful it seems they are not directly related to the issue.

I see!

> Surprisingly though I cannot reproduce that issue any more (tried in several ways) so likely it was caused by something else. 

Okay.

In that case, I'll close this PR for now. If you come across this issue the next time, please let us now and feel free to open a PR. For now, I'll go ahead and close this!

PR #1197, opened 2026-05-06
Fix tiled batched inference with channels

constantinpape:


constantinpape (commented):
Test errors seem unrelated so I will merge this.

PR #1206, opened 2026-05-13
Minor release notes formatting update

anwai98:
This is a bogus PR. My intention is to check the runners, because I suspect something is broken in the bioimageio backbone extraction pipeline. I'll let it run and confirm!


anwai98:
Yayy the tests are broken! 🥳 (I mean nothing to be super happy about, but I know the fix hehe)

anwai98:
The tests are fixed on `master` now! (it was indeed a `bioimageio.core` compatibility issue). Let's see if the tests pass here too. If so, this was a success haha!

PR #1207, opened 2026-05-13
Fix predictor adapter load state dict to accept keyword arguments

anwai98:
`pytorch`'s model loading engine calls has changed and `strict=True` is the hard default for a while (not sure what changed that the tests are breaking right now).

Let's see how the tests work at https://github.com/computational-cell-analytics/micro-sam/pull/1206, this would confirm the suspicion!

### DIFF
--- a/micro_sam/bioimageio/model_export.py
+++ b/micro_sam/bioimageio/model_export.py
@@ -235,10 +235,10 @@ def _check_model(model_description, input_paths, result_paths):
                 "mask_prompts": mask_prompts,
                 "embeddings": embeddings,
             },
-        ).as_single_block()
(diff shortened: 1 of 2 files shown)


anwai98:
Seems like there is a new release around `bioimageio` which broke some functionalities. Lemme have a closer look!

anwai98:
Ayy eureka it works haha (I simply adapted our exports to `bioimageio.core` v0.10.3`. We don't need them actively atm, but we would need them super soon ;)

If the tests pass, I'll simply merge this to fix related broken tests in other (active) branches!

anwai98:
This worked amazingly nice. I'll go ahead and merge this!

PR #1208, opened 2026-05-14
Fix motion model in interactive tracking

anwai98:
Okay meanwhile looking at a user's issue, I realized there are two tiny bugs in interactive tracking:

1. The first one for missing list brackets in `np.array(a, b)` is obvious!  
2. The sign computed for the move was reverted, i.e. both causing the mask shift to drift in the wrong direction from the 3rd tracked frame onwards (I thought about it, and maybe with cells which drift slowly / less is where we didn't notice this)

These are super minor changes, but should do the trick! @constantinpape


constantinpape (approved):
Nice catch!

Issue #1113, opened 2025-09-26
Installing micro-sam headless


ElpadoCan:
Hello, 

First of all, congratulations on the publication, well-deserved! 

Secondly, I was wondering if it's possible to install micro-sam headless, without GUI libraries (such as `qt-main`). 

I'm asking because I'm planning to integrate micro-sam into Cell-ACDC GUI as plugin. However, in Cell-ACDC the GUI libraries are installed via pip and (as we all know) conda does not communicate well with pip.  

I would only need the core functionalities of micro-sam without the GUI. 

Is that possible? I could not find anything in the documentation. I apologise if I missed it. 

If not possible yet, how hard would it be to implement for you? This would also enable installation of micro-sam on HPC clusters where often there is no desktop environment. 

Thanks! 

Francesco

ElpadoCan:
Hi Constantin, 

Thanks for the quick reply! 

Oh yeah, you are right, I'm stupid, I didn't see you also had the `pip install -e .` installation option, I knew I was missing something :D Thanks! 

Speaking of pip, are you planning to release on PyPi? A micro-sam package on PyPI without dependencies (for software like Cell-ACDC that automatically manages environments for users) would be extremely helpful, and it seems like you have everything in place for that. 

Cheers,
Francesco

constantinpape:
> Speaking of pip, are you planning to release on PyPi? A micro-sam package on PyPI without dependencies (for software like Cell-ACDC that automatically manages environments for users) would be extremely helpful, and it seems like you have everything in place for that.

Unfortunately that is not possible right now, because we have (non-gui) dependencies that are conda-only. Removing those would be a major effort. What would be possible in a relative easy way is to have a micro-sam headless version on conda-forge.

constantinpape:
> Headless on conda-forge would also help, true!

I will try to set this up next week; will let you know.

ElpadoCan:
Hi Constantin,

I was trying to set up the headless environment; however, by installing the packages one by one, I found out that when I install the package `torch_em`, other GUI pacakges get installed, such as `magicgui` and `superqt`. Should I remove `torch_em` too and install it with `pip install -e .`? 

Thanks!

ElpadoCan:
Actually, also [elf](https://github.com/constantinpape/elf) is installing `qtpy` and `superqt` even if I install it with `pip`, so I guess I would have to go with `--no-deps` there?

constantinpape:
[...]

PR #1213, opened 2026-05-20
Enable pip install by deferring conda-only imports

keejkrej:
## Summary

This PR makes the core micro-sam Python API importable and usable via pip without requiring conda-forge-only packages (`nifty`, `vigra`, `python-elf`).

Changes:
- Defer imports of `nifty`, `vigra`, and `python-elf` to the functions that use them
- Add pip-installable `install_requires` to `setup.cfg` (PyPI-available deps only)

Relates to #1113.

## Motivation

Several dependencies used by micro-sam are only available on conda-forge. Importing `micro_sam.util` currently pulls them in at module load time, which breaks pip-only installs — even when the caller only needs the core SAM workflow (load model → compute embeddings → segment from prompts).

This matters for:
- Integration into applications that manage their own pip environments (e.g. Cell-ACDC)
- HPC / headless setups without a full conda stack

## Why this doesn't affect the core neural network workflow

The SAM inference path (`get_sam_model`, `precompute_image_embeddings`, `segment_from_points`) depends on **torch**, **segment-anything**, and standard pip packages. It does not use nifty, vigra, or python-elf.

The deferred packages serve **orthogonal infrastructure**, not the model itself:

| Package | Used for | When needed |
|---------|----------|-------------|
| `nifty` | Block tiling for large images | Tiled 2D/3D inference only |
| `vigra` | Eccentricity-based centroids (`mode="v"`) | Optional prompt generation |
| `python-elf` | H5/N5/Zarr I/O, parallel connected components | Keyed file loading, mask post-processing |

None of these participate in encoder/decoder forward passes. Users on pip who don't need tiling, vigra centroids, or elf I/O can install and run the core workflow without conda. Users who need those features can still `conda install -c conda-forge nifty vigra python-elf` as today.

## Test plan

- [ ] `from micro_sam.util import get_sam_model` succeeds without nifty/vigra/python-elf installed
- [ ] `from micro_sam.prompt_based_segmentation import segment_from_points` succeeds without nifty installed
- [ ] Existing conda CI/tests pass unchanged
- [ ] Tiled inference still works when nifty is installed
- [ ] `load_image_data(..., key=...)` still works when python-elf is installed

Made with [Cursor](https://cursor.com)


constantinpape:
Hi @keejkrej ,

thank you for your interest in improving the usability of `micro_sam`. We are currently working on another solution for this issue though and are creating a dedicated C++ library with all functionality that is needed and that can be installed via pip:
https://github.com/computational-cell-analytics/bioimage-cpp

This has the advantage that all features of `micro_sam`  will be available to pip users. We are already in the process of migrating `elf`, see https://github.com/constantinpape/elf/pull/141 and will migrate `torch_em` and then `micro_sam` next. I expect this to be done in 2-3 weeks.

In general, contributions to `micro_sam` (or other libraries developed by us) are very welcome, but I recommend to first open an issue for any larger contribution to see if working on this makes sense.

keejkrej:
Thanks! Very interested in your new lib, for now I'll just install from my fork.

PR #1224, opened 2026-06-08
Add support scripts for C. albicans data finetuning

### LINKED ISSUE #1214: Finetuning microsam AIS (continued)
Hello, continuing from this issue #1211, our data also contain some mammalian cells in addition to the hyphal cells. We grow them in a coculture. The mammalian cells are imaged with a fluorescence cytoplasmic channel and a DAPI nuclear channel, while the hyphal cells are best visualized with DIC. When in focus, the mammalian cells can be segmented quite well by AIS, but is a little problematic when a part of a cell goes out of focus while a part of it is still in focus along the z stack. So I also want to fine-tune the model on the mammalian cell z-stack images. 

Since the mammalian cell images contain 2 channels. How should I process the channels for finetuning, as the `microsam.train._check_loader` only accepts 1 or 3 channel inputs? The `microsam.util._to_image` function just adds a third channel of 0s, but in your paper, you described: "averaging the two channels to obtain a single channel that is then duplicated three times works better ... but is detrimental compared to using both channels independently." I want to align with your training protocol as close as possible to minimize the amount of new data distribution the model needs to learn.

anwai98:


anwai98:
I scratched up a debug script. I'll come back to this in the coming days!

anwai98:
https://github.com/computational-cell-analytics/micro-sam/issues/1214

Issue #1227, opened 2026-06-09
Renaming breaks napari


guiwitz:
The recent renaming of the package in 5033b0a5c2543f44ba2e65ea771cc1aa044385b8 unfortunately breaks the mechanism used by napari to discover plugins. Installing in a fresh environment using `napari==0.6.6` and `python==3.13`, I'm getting the following when launching napari:
```
ERROR : MainThread : napari.manifest -> 'micro_sam' could not be imported: The name field in the manifest ('micro-sam') must match the package name ('micro_sam')
```
Also when checking if the napari manifest is ok using `npe2 validate micro_sam`, I'm getting: 
```
🅇 Unexpected error in 'micro_sam'.
ValueError: Could not find manifest for 'micro_sam' as either a package name or a file..
```
The fix is to rename the plugin in the `napari.yml` manifest as well.

constantinpape:
Closed by #1228

PR #1228, opened 2026-06-09
rename using new package name

guiwitz:
This PR offers a fix to the bug reported in #1227 by adjusting the package name in the napari manifest file. I tested that a local install worked with this solution.


constantinpape (approved):
Thank you for identifying the error and fixing it @guiwitz !!

(I was a bit confused by name resolution in PyPI which lead to this change)

PR #1234, opened 2026-06-11
Make the v1 training choose precision from the hardware

chuyaowang:
The PRs mentioned in the issue #1214. Work with [this PR for torch-em](https://github.com/constantinpape/torch-em/pull/718)

fp16 mixed precision causes NaN in Conv2d(512, 512, 3x3) inside the UNETR decoder on GPUs that lack Tensor Cores (e.g. GTX 1660 Ti). Without Tensor Cores, cuDNN accumulates the 4608-element dot product in fp16, overflowing its ceiling of ~65504. bfloat16 avoids this by sharing fp32's exponent range. Confirmed by forward hook diagnostic and deterministic mode test ruling out Winograd as the specific cause.

- Add `_get_mixed_precision_dtype(device)` helper that returns torch.float16 for GPUs with Tensor Cores and torch.bfloat16 otherwise
- Detect Tensor Core presence via compute capability and device name: Ampere+ (sm_8.x+) always fp16; Volta/Turing (sm_7.x) fp16 except GTX 16xx series (TU116/TU117); Pascal and below bfloat16
- Avoid torch.cuda.is_bf16_supported() as it reflects software-level dtype support and returns True on sm_7.5 with CUDA 12.x regardless of Tensor Core presence
- Wire dtype through train_sam (JointSamTrainer and SamTrainer) and train_instance_segmentation (DefaultTrainer) via the new mixed_precision_dtype parameter added to torch_em DefaultTrainer

### DIFF
--- a/micro_sam/util.py
+++ b/micro_sam/util.py
@@ -232,6 +232,29 @@ def get_device(device: Optional[Union[str, torch.device]] = None) -> Union[str, 
     return device
 
 
+# The compute capability that runs bfloat16 natively. Gated on this and not on
+# 'torch.cuda.is_bf16_supported', whose default counts emulation and reports bf16 on Volta.
+BF16_MIN_CAPABILITY = (8, 0)
+
+
+def training_autocast_dtype(device: Optional[Union[str, torch.device]] = None) -> Optional[torch.dtype]:
+    """The dtype that training runs in on a device.
+
+    Training runs in bfloat16 or fp32, never float16: float16 needs loss scaling and still overflows
+    in the decoder convolutions on GPUs that accumulate it in half precision.
+
+    Args:
+        device: The device the forward pass runs on. Defaults to the best available one.
+
+    Returns:
+        bfloat16 where the device runs it natively, or None where training stays in fp32.
+    """
+    device = torch.device(get_device() if device is None else device)
+    if device.type != "cuda" or not torch.cuda.is_available():
+        return None
+    return torch.bfloat16 if torch.cuda.get_device_capability(device) >= BF16_MIN_CAPABILITY else None
+
+
 def get_embedding_function(model_type: str) -> callable:
     """Get the precompute-embeddings function for the model family of `model_type`.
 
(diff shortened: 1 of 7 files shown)


anwai98:
Hi @chuyaowang,

Sorry for the delayed response (the team has been quite busy throughout the summer).

I thought about this in the past days, and I kinda agree that we need a precision switching parameter for autocast (funny enough, stumbled along this problem last week -- patched it for me internally, and then remember that there's someone who proposed a better solution haha) 😉 

How about I take a look at your PR over the weekend and come back to you?

chuyaowang:
no problem. let me know how it goes!

anwai98:
Thanks a lot for digging into this, the fp16 accumulation overflow diagnosis was spot on.

I have reworked the fix on this branch and retargeted the PR to `dev`, to line it up with what we did for the SAM2 training in #1336:

- Training now runs in bfloat16 where the hardware runs it natively, and in fp32 everywhere else (never in fp16 automatically). Both training stacks follow the same rule now.
- That removes the need to detect tensor cores: the compute capability alone is the gate, so the device name match for the GTX 16xx series is gone.
- `training_autocast_dtype` moved from `micro_sam/v2/util.py` to `micro_sam/util.py`, so v1 and v2 share one helper. `torch-em` disables the GradScaler for bfloat16 by itself (constantinpape/torch-em#718, constantinpape/torch-em#758), so nothing else is needed here.

One consequence worth flagging: Volta and Turing cards that do have tensor cores (V100, T4, RTX 20xx) now train in fp32 rather than fp16. That is slower and uses more activation memory, but it takes fp16 out of training entirely, which is the trade we also accepted for SAM2 (although if users really wanna use `fp16` -- they should change it manually under `mixed_precision_dtype`)

The force push replaced your commit, so I kept you as co-author on the new one. The PR description still describes the original approach, feel free to update it.

anwai98:
PS. I'll merge this once the CIs pass!

chuyaowang:
It's nice that you found a more general solution. Please go ahead with the changes.

anwai98:
Thanks for the contribution @chuyaowang 🥳

PR #1235, opened 2026-06-16
Fix uSAM2 annotator prototype

anwai98:
Trying to fix and test it. I'll keep this open!


anwai98:
Okay this now atleast lets the tool open haha. Will check other things out after lunch!

anwai98:
This makes the base annotator work for now. I'll work on the next steps in another PR!

PR #1244, opened 2026-06-19
Add 3d toggle for selective z-propagation and early stopping in interactive segmentation

anwai98:
So the tricks here are mostly applied to aim for speeding up CPU-based segmentation

1. Early stopping for missing masks after n slices
2. Selective propagation for any "n (continuous) slices"

Both are currently optional, and works like a charm tbh.

I think it's a good addditional to the segmentation tool, and will merge it for now. If it feels too much, we can remove it in future.

PS. It's hidden down inside "Segmentation Settings" dropdown, and shows up only when "Apply to Volume" is selected for volumetric interactive segmentation! ;)


anwai98:
Also, precomputation is a big bottleneck - if you precompute once, it's gonna make things so much faster, it's crazy! (but again, relatively fast - not like blazing fast)

PR #1246, opened 2026-06-21
Add support for AMG with SAM2 models

anwai98:
Would be good for some use-cases. I'll work on GUI compat in some -- need to debug a few details first!


anwai98:
Nice. Works fine. I'll go ahead and merge this!

PR #1247, opened 2026-06-22
Add support for tiled 3d interactive segmentation

anwai98:
Okay this is an ambitious attempt to see whether I can enable true tiling in 3d interactive segmentation. I'll quickly try it out and call it a day!


anwai98:
Okay tile grouping was missing before in `micro-sam` as well -- I have an idea ;)

anwai98:
Looks pretty smooth to me! ;) 

I'll merge it once the tests pass!

PR #1248, opened 2026-06-22
Add better handling for singletons and improper channels for raw input data

anwai98:
Had this idea for a while too. Lemme check stuff out and see how things work there! 

TLDR for the changes here would be to handle singletons and multi-channel situations better!


anwai98:
Works as expected! ;)

PR #1249, opened 2026-06-22
Add pixel classification prototype

anwai98:
Okay the first prototype is up and running. For my first try-outs, it works great. Here's a summary:

1. Works for 2d and 3d images
2. Implemented tiling-based pixel classification for larger image/volumetric data.
3. Touched up some technical details for both classification tools (object and pixel)

I wanna add a few more features (I wanna keep the widget as minimalistic as possible), and then ask for review from the team!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -207,3 +207,5 @@ data/
 embeddings/
 *.zarr
 
+# Trained classifier exports.
+*.joblib
--- a/environment.yaml
+++ b/environment.yaml
@@ -33,3 +33,4 @@ dependencies:
     - pip:
         - git+https://github.com/ChaoningZhang/MobileSAM.git
         - git+https://github.com/facebookresearch/sam2.git
+        - git+https://github.com/wimmerth/anyup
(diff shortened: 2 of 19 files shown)


anwai98:
Okay the tool is actually ready to play with ;) It currently supports scribble annotations (just like object classification tool). It also supports tiling, and would tile by default for large images. In addition, the tool now supports both loading and saving trained models.

It's good to test already (should work just fine)

cc: @caroteu @constantinpape

anwai98:
Aaaaaaaaand now it's sorted hehe (made annotation a bit more handy by adding some keybindings)

anwai98:
Important note:

1. Add metadata for trained random forest model spec (eg. tiling / with or without PCA / with or without AnyUp / model name, etc.)
    - Lemme see how this works with the `joblib` stored model pipeline!

anwai98:
I think we are at a good point here. To keep all things in the same place, I'll go ahead and merge this!

PS. We'll work on future feedback via individual PRs one after the other!

PR #1250, opened 2026-06-22
More features for classification tools

anwai98:
Well I came up with another improvement on top of https://github.com/computational-cell-analytics/micro-sam/pull/1249 (might be cool to discuss and see if we wanna keep one or both of the features)

The idea was to integrate:
1. PCA over all interpolated features (in some cases, I noticed via the tool that there's a slight advantage)
2. Add AnyUp upsampling

(1+2 are currently checkboxes, if not applied, all features with interpolation based upsampling works! either one or both can be applied as the user likes)

PS. AnyUp does not really look good for pixel classification at all, but PCA does look much more crisp in terms of segmentation quality.

What do we think? @caroteu @constantinpape

### DIFF
--- a/environment.yaml
+++ b/environment.yaml
@@ -33,3 +33,5 @@ dependencies:
     - pip:
         - git+https://github.com/ChaoningZhang/MobileSAM.git
         - git+https://github.com/facebookresearch/sam2.git
+        # Optional AnyUp feature upsampler for the pixel and object classifiers.
+        - git+https://github.com/wimmerth/anyup
--- a/environment.yaml
+++ b/environment.yaml
@@ -33,5 +33,4 @@ dependencies:
     - pip:
         - git+https://github.com/ChaoningZhang/MobileSAM.git
         - git+https://github.com/facebookresearch/sam2.git
-        # Optional AnyUp feature upsampler for the pixel and object classifiers.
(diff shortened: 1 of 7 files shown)


constantinpape:
I tried with the PCA, this works much better for the SAM2 natural images model. It was basically unusable without PCA. I did not try with AnyUp yet, would need to install it.

A few things to keep in mind:
- we should store relevant infos in the classifier (which embeddings was it trained on, PCA, etc.) so that we can check it's applied in the correct manner
- we should keep this aligned across pixel- and object-classification.

anwai98:
Thanks @constantinpape for checking it out.

Quick re:

> we should store relevant infos in the classifier (which embeddings was it trained on, PCA, etc.) so that we can check it's applied in the correct manner

Oh good point. I'll plug this to the export model section!

> we should keep this aligned across pixel- and object-classification.

Yes, that's something I'm meticulously taking care of (all features work atm for both pixel and object classifiers)

anwai98:
I briefly talked to @constantinpape about this. Since we agree these two would be a good addition to our classification tool suite, I'll go ahead and merge them to the other working branch (we should track the changes at https://github.com/computational-cell-analytics/micro-sam/pull/1249)

PR #1251, opened 2026-06-23
Refactor tracking annotator

anwai98:
I started working on it. I will continue on this once a few more segmentation stuff is sorted out (eg. model integration, etc).

For now, added a few low hanging plans I've had in the past!


anwai98:
Well another thought: it would be great to support GEFF exports (I worked on CTC exports for automatic and interactive already via tool here -- will expand it to make it modular to GEFF / trackmate xml as well)

anwai98:
Okay another idea:
1. For tracking support: I'll add selective install_requires, eg. `pip install micro_sam[tracking]` -- this will ensure how motile and trackastra should be installed in the package, otherwise the installation would be crazy slow!

anwai98:
This is sorted now. I'll go ahead and merge this!

PR #1252, opened 2026-06-23
Add support for preliminary micro-sam2 models

anwai98:
Need to change the model suffix, and take a closer look at resize and normalize consistency.

Lemme take care of them and come back to finalize this (I'll leave this as a draft)


anwai98:
The mac tests failing are unrelated. I'll check later if they break on `main` too -- and then fix them in this direction. Besides, this is good to go from my side!

I'll merge this once the linux and windows tests pass!

anwai98:
This part right here works fine now. I'll go ahead and finally merge this (was way more work for alignment than expected haha ;))

PR #1255, opened 2026-06-24
Start fixing tiling based segmentation in annotator

anwai98:
This should sort out the tiling behaviour. Can you double check this @constantinpape?

### DIFF
--- a/micro_sam/v2/util.py
+++ b/micro_sam/v2/util.py
@@ -55,6 +55,11 @@ DEFAULT_TILING_THRESHOLD = 768
 DEFAULT_TILE_SHAPE = (512, 512)
 DEFAULT_HALO = (128, 128)
 
+# Default z block / halo for volumetric (3d) tiling, matching the UniSAM2 training crop. Set the
+# z tile >= the slice count to disable z-tiling (the whole volume becomes one block along z).
+DEFAULT_TILE_Z = 4
+DEFAULT_HALO_Z = 2
+
 
 def needs_default_tiling(shape):
     """Whether default in-plane tiling should be enabled for a given image shape.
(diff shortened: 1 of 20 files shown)


constantinpape (dismissed):
This should be centrally refactored and not be part of a private widget implementation.

anwai98:
Okie. Would make sense to put it somewhere at `v2/util.py` -- I'll take care of it!

anwai98:
What do you think now? @constantinpape

micro_sam/v2/util.py
```diff
@@ -48,6 +48,31 @@
...
 
+# Default in-plane tiling for large images. Tiling is enabled when an in-plane axis exceeds
+# DEFAULT_TILING_THRESHOLD; the SAM input patch per axis is then DEFAULT_TILE_SHAPE + 2 * DEFAULT_HALO,
+# which is kept equal to the threshold (512 + 2 * 128 = 768).
+DEFAULT_TILING_THRESHOLD = 768
+DEFAULT_TILE_SHAPE = (512, 512)
+DEFAULT_HALO = (128, 128)
+
+
+def needs_default_tiling(shape):
```
constantinpape, line 66: What happens for multi-channel data?

constantinpape:
I also think you have to update how the 3D segmentation does tiling accordingly.

micro_sam/v2/util.py
anwai98: Need to double check it, in particular what do we really wanna do in terms of multi-channel handling (backend has some support for it, reinvestigate)

anwai98: Idea: add ndim parameter -- max 3 min 2

anwai98: This is sorted now ;)

anwai98:
I'll convert this to a draft since there are a good few things to take care of here. I'll open it to review once I've made some progress!

PR #1256, opened 2026-06-26
Improve precomputed embedding injestion in decoder for volumetric data

anwai98:
@constantinpape tiny fix to make per-slice volumetric segmentation faster!


anwai98:
I'll merge this since all works as expectedhere!

PR #1266, opened 2026-06-30
Remove global event emitter error suppression filter

anwai98:
The latest napari version does not like this, paired with refreshing structure (creates a huge chaos because it filters out some valid errors -- and that's dangerous).

It's best that we keep the event emitter signal errors around -- the annotators just have to be a bit patient and slow hehe!


anwai98:
Poof, that fixed CI stuff (that's good)! I'll go ahead and merge this!

PR #1267, opened 2026-06-30
Enable switching automatic tracking model and solver

anwai98:
Super minor changes, but makes automatic tracking more flexible for users! (eg. I needed to switch around the pretrained model and solver choice to improve tracking).

This would hit `main` directly since it's still part of v1. What do you think? @constantinpape


anwai98:
I'll look into the broken pip tests in a sec, not sure what's going on there!

anwai98:
Okay the tests pass now. Can you give it a quick review @constantinpape?

constantinpape (approved):

anwai98:
Nice, thanks Constantin! :D

PR #1271, opened 2026-07-01
Fix SAM2 automatic segmentation on MPS and drop hard vigra dependency

anwai98:
There were two hickups:
1. Existing vigra functions in automatic segmentation, the global import broke things -- and looking at this, I decided to port all v2 remaining things to `bioimage-cpp` as well. I did some checks, looks good to me.
2. The mechanism how arrays are converted around for MPS device (this broke automatic segmentation before for MPS)

Should work now!

@Marei33 can you retest this please on this branch?

cc: @constantinpape


constantinpape (approved):
We should avoid any nifty or vigra dependencies --- so looks good to me!

anwai98:
Cool. Then I'll go ahead and merge this (and ping in the group for Marei and others)

PR #1273, opened 2026-07-01
Keep batch annotator open when annotation is complete

anwai98:
Lemme test this in several combinations first how it works!


anwai98:
Nice. This is good now. Should work as expected!

PR #1274, opened 2026-07-01
Migrate to z5py backend for storing and loading image embeddings

anwai98:
Well this was a tiny bit more tricker than I anticipated hehe, but it worked out! (crazy speed ups even in my shallow benchmarks haha)

One caveat: zarr is still kept around for in-memory embeddings (i.e. tiled with no save_path). z5py does not have in-memory stores, and that path does no disk IO to accelerate anyway.

What do you think about this? @constantinpape

### DIFF
--- a/environment.yaml
+++ b/environment.yaml
@@ -29,8 +29,8 @@ dependencies:
     - timm
     - trackastra >=0.5.3
     - xarray
-    - z5py
-    - zarr
+    - z5py >=3
+    - zarr >=3
     - pip:
         - git+https://github.com/ChaoningZhang/MobileSAM.git
         - git+https://github.com/facebookresearch/sam2.git
--- a/environment.yaml
+++ b/environment.yaml
@@ -29,6 +29,7 @@ dependencies:
     - timm
     - trackastra >=0.5.3
     - xarray
+    - z5py
     - zarr
     - pip:
         - git+https://github.com/ChaoningZhang/MobileSAM.git
--- a/test/test_util.py
+++ b/test/test_util.py
@@ -14,8 +14,10 @@ except ImportError:
 
 from skimage.data import binary_blobs
 from skimage.measure import label
-from micro_sam.util import VIT_T_SUPPORT, SamPredictor, get_cache_directory
+from micro_sam.util import VIT_T_SUPPORT, SamPredictor, get_cache_directory, _open_embeddings
 from micro_sam.v1.util import get_sam_model, set_precomputed
+
+ZARR_MAJOR = int(zarr.__version__.split(".")[0])
 
 
 class TestUtil(unittest.TestCase):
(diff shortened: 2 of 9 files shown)


test/test_util.py
```diff
@@ -7,11 +7,18 @@
 import torch
 import zarr
 
+try:
+    import z5py
+except ImportError:
+    z5py = None
```
constantinpape: Do we need the import guard here? You don't have it in `[‎micro_sam/util.py](https://github.com/computational-cell-analytics/micro-sam/pull/1274/changes#diff-69cb491559d563c5e6db71cdecb33d2a2dff8a1ae3502814c904a645b8e4b5bd)`. So using `micro_sam` without `z5py` would now fail anyways.

test/test_util.py
```diff
@@ -385,5 +392,92 @@ def check_slices(embeddings):
         check_slices(embeddings)
 
 
+@unittest.skipUnless(z5py, "The z5py embedding backend requires z5py.")
```
constantinpape: I would make `z5py` a strict dependency and then not skip this, see also comment above.

environment.yaml
```diff
@@ -29,6 +29,7 @@ dependencies:
     - timm
     - trackastra >=0.5.3
     - xarray
+    - z5py
```
constantinpape: Maybe we should pin both `z5py` and `zarr` to versions `>=3`? Same in `setup.cfg`.

constantinpape:
Looks very good overall!

> One caveat: zarr is still kept around for in-memory embeddings (i.e. tiled with no save_path). z5py does not have in-memory stores, and that path does no disk IO to accelerate anyway.

This is fine. `zarr` is a fairly lightweight dependency so there's no harm in keeping it around. I would assume that the in-memory is so fast that we don't care about the speed for this anyways.

constantinpape:
Regarding the test failures on Mac: It does not properly install `z5py`. We have seen similar issues in CochleaNet, but I thought that this was due to some version issues. Could you maybe try it on your test macbook? If there is some issue with the conda package I can investigate.

environment.yaml
anwai98: Good point actually. Lemme take care of it!

test/test_util.py
anwai98: +1 Yep, good idea!

test/test_util.py
anwai98: Not needed actually. I'll make it pinned global import. On it!

anwai98:
> I would assume that the in-memory is so fast that we don't care about the speed for this anyways.

Exactly. It's insanely fast since it's just a proxy container that stays in memory all the time!

anwai98:
This would still fail, but let's see if new build try-out a) passes in the first place, and then b) if it helps out here!

constantinpape:
I think your PR in https://github.com/conda-forge/z5py-feedstock/pull/90 should then fix the issue. I set it to automerge, so it will get merged if all tests pass.

anwai98:
Nice ;) Thanks @constantinpape (fingers-crossed haha)

PR #1276, opened 2026-07-02
Refactor CLI for micro-sam backend support

anwai98:
WIP!


anwai98:
Okay this is sorted now! @constantinpape 

Some details on how things would work:
1. Needs a `pip install -e .` in the current branch
2. Check out some simple integrations, eg. `micro_sam --help` / `micro_sam --version` / `micro_sam info`
3. And the other ports:
    a. `micro_sam.annotator` -> `micro_sam segmentation_annotator`
    b. `micro_sam.annotator_tracking` -> `micro_sam tracking_annotator`
    c. `micro_sam.automatic_segmentation` (SAM1) -> `micro_sam v1 automatic_segmentation`
    d. `micro_sam.train` (SAM1) -> `micro_sam v1 train`
    e. `micro_sam.evaluate` / `micro_sam.benchmark_sam` -> `micro_sam v1 evaluate` / `micro_sam v1 benchmark_sam`
    f. `micro_sam.precompute_embeddings` / `micro_sam.info` -> `micro_sam precompute_embeddings` / `micro_sam info`
    g. _...and all other CLIs too_

micro_sam/util.py
```diff
@@ -138,6 +138,30 @@ def get_device(device: Optional[Union[str, torch.device]] = None) -> Union[str,
     return device
 
 
+def get_embedding_function(model_type: str) -> callable:
```
constantinpape, line 141: This doesn't raise an error for an invalid `model_type`. I think it's better to raise here if `model_type` is invalid rather than relying on `v1.util.precompute_image_embeddings` to raise later.

constantinpape (commented):
I will test this properly tomorrow morning. Noticed one minor thing that could be improved in the code review.

constantinpape:
Here's some more feedback:

General:
- Shorten the first part of the doc string so that it is not cut from the general help message (see screenshot below)
- It would be good to have a short usage example in the help message of each command (except `info` and `v1`).
- Unify naming `"-i"` -> `"--input"` (currently varies between `input`, `input_path`, `input_folder`); `"-o"` -> `"--outtput"`
- Rename `halo` -> `overlap` (mirror change in GUI)

image

Specific:
- `automatic_segmentation`: 
   - add an optional `--view` to display results in napari?
   - segmentation in a folder does not work: `micro_sam automatic_segmentation --input_path /home/pape/.cache/micro_sam/sample_data/image-series.zip.unzip/series/ -o test-out-seg --pattern *.tif` fails with `  File "/home/pape/Work/my_projects/micro-sam/micro_sam/cli.py", line 324, in automatic_segmentation
    assert len(input_paths) > 0, "'micro-sam' could not extract any image data internally."`; folder contains 3 tif images
  - Note: it actually works when passing `"*.tif"`, so probably just an issue in argument parsing
- `batch_annotator`: opens the segmentation annotator but should support all of the annotators

I haven't checked the rest in detail

micro_sam/pixel_classification.py
```diff
@@ -539,20 +540,24 @@ def run_prediction_with_pixel_classifier(
         upsampler: An optional AnyUp model (see `get_anyup_upsampler`) to upsample the embeddings
             with the image as guidance instead of plain interpolation. Use the same setting that
             the classifier was trained with.
+        model_type: The model family of `predictor`, used to pick the matching embedding function
+            (SAM1 / SAM2 / VFM). If not given, the SAM1 embedding function is used.
```
constantinpape, line 544: I think we should use the `SAM2` embedding function by default --- and use our uSAM2 model as the default model.

anwai98: I see. Then I'll wrap this out around it like that!

constantinpape:
Should we merge this @anwai98 ?

anwai98:
Oh I kind of was waiting for your approval here haha. If you assessment is positive here, let's go for it. It is GTM from my side!

constantinpape:
Yeah let's merge it.

micro_sam/pixel_classification.py
constantinpape: Let's address this separately.

anwai98:
Nice! ;)

PR #1278, opened 2026-07-03
Update default parameters for automatic segmentation

anwai98:
Still need to add the one for sparse mode, but should be a couple more minutes!


anwai98:
I tested the parameters -- things look much better now! (for both LM and EM)

I'll go ahead and merge this! @constantinpape

PR #1280, opened 2026-07-11
Some more bug fixes in annotator GUI

anwai98:
WIP. Need to double check if things work as expected!


anwai98:
Nice, this worked out quite nice actually.

Okay so I fixed three things here:
1. Improving the progressbar semantics
2. When images are reloaded, the embedding widget should be restored back to expected defaults
3. Improve interactive segmentation quality for iterative prompting in 3d

All works pretty nice now. I'll go ahead and merge this!

anwai98:
Nah, I'll wait for the CIs to pass!

PR #1281, opened 2026-07-11
Add unification class for resizing across training and inference

anwai98:
WIP. I realized it's better to unify resize transforms in one place. Lemme do some tests before I merge this!


anwai98:
Okay, this works like a charm -- was important to take care of this for me. I'll go ahead and merge this right away!

PR #1282, opened 2026-07-13
added examples to clarify how to submit args

lufre1:
encountered occasions when users were confused how to submit tiling args. Added example to clarify how to use it

### DIFF
--- a/micro_sam/automatic_segmentation.py
+++ b/micro_sam/automatic_segmentation.py
@@ -391,10 +391,12 @@ def main():
         "-c", "--checkpoint", default=None, type=str, help="Checkpoint from which the SAM model will be loaded."
     )
     parser.add_argument(
-        "--tile_shape", nargs="+", type=int, help="The tile shape for using tiled prediction.", default=None
+        "--tile_shape", nargs="+", type=int,
+        help="The tile shape for using tiled prediction. Example: --tile_shape 1024 1024.", default=None
     )
     parser.add_argument(
-        "--halo", nargs="+", type=int, help="The halo for using tiled prediction.", default=None
+        "--halo", nargs="+", type=int,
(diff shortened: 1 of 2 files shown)


micro_sam/automatic_segmentation.py
```diff
@@ -399,10 +391,12 @@ def main():
         "-c", "--checkpoint", default=None, type=str, help="Checkpoint from which the SAM model will be loaded."
     )
     parser.add_argument(
-        "--tile_shape", nargs="+", type=int, help="The tile shape for using tiled prediction.", default=None
+        "--tile_shape", nargs="+", type=int,
+        help="The tile shape for using tiled prediction. Example: --tile_shape 1024 1024.", default=None
```
anwai98:
```suggestion
        help="The tile shape for using tiled prediction. Example: --tile_shape 384 384.", default=None
```

micro_sam/automatic_segmentation.py
```diff
@@ -399,10 +391,12 @@ def main():
...
     )
     parser.add_argument(
-        "--halo", nargs="+", type=int, help="The halo for using tiled prediction.", default=None
+        "--halo", nargs="+", type=int,
+        help="The halo for using tiled prediction. Example: --tile_shape 128 128.", default=None
```
anwai98:
```suggestion
        help="The halo for using tiled prediction. Example: --tile_shape 64 64.", default=None
```

anwai98 (approved):
Hi @lufre1,

This is a great idea, I like it. Thanks for taking care of it!

PS. I'll check why the `pip` tests are broken and merge this once it works out!
PPS. There are two minor suggestions below -- I'll also commit them!

anwai98:
Okay it was indeed a new sync miss in latest `bioimage-cpp`. Should fix multi-dimensional segmentation and the `pip` tests too. Let's see!

PR #1283, opened 2026-07-13
Unify input normalization

anwai98:
I decided to consolidate all of these to be the same -- there's just too many quotients around, and I would like to centralize them all around percentile normalization. Lemme do a few final tests, then I'll merge this once the tests pass!


constantinpape:
Good idea!

micro_sam/v2/normalization.py
```diff
@@ -0,0 +1,48 @@
...
+UINT8_RANGE = (0.0, 255.0)
+SAM2_INPUT_RANGE = (0.0, 1.0)
+
+
+def normalize_raw(
+    raw: np.ndarray,
+    axis: Optional[Union[int, Tuple[int, ...]]] = None,
+    output_range: Tuple[float, float] = SAM2_INPUT_RANGE,
+    dtype: Union[str, np.dtype] = "float32",
+) -> np.ndarray:
```
constantinpape, line 17: Do we always want to do percentile? Make the percentiles params?
(Let's quickly discuss this in person before merging)

anwai98: Okie. Let's chat about this tomorrow then!

micro_sam/v2/normalization.py
```diff
@@ -0,0 +1,48 @@
...
+    """Percentile-normalize raw image data to the requested input range.
+
+    The 1st and 99th percentiles are mapped to zero and one and values outside
+    this range are clipped.
+    """
+    dtype = np.dtype(dtype)
+    if dtype != np.dtype("uint8") and not np.issubdtype(dtype, np.floating):
+        raise ValueError(
+            f"Invalid normalization dtype '{dtype}'. Expect a floating dtype or uint8."
```
anwai98: Don't reject dtypes. Both int8, uint16 and int16 are very common datatypes in microscopy! 

You can reject 32 and 64 bit integer types.

anwai98: Write tests for all the supported dtypes and sensible input ranges

micro_sam/v2/normalization.py
anwai98: Let's add `lower_percentile` and `upper_percentile` as parameters, with the defaults as the hard-coded values you have right now. We don't make it settable anywhere else for now.

anwai98 (commented):
See comments below

micro_sam/v2/normalization.py
anwai98: Sorted!

micro_sam/v2/normalization.py
anwai98: Sorted!

anwai98:
This should be sorted now. What do you think? @constantinpape

micro_sam/v2/normalization.py
```diff
@@ -0,0 +1,51 @@
...
+RAW_NORMALIZATION = "percentile_2_98"
+UINT8_RANGE = (0.0, 255.0)
+SAM2_INPUT_RANGE = (0.0, 1.0)
+
+
+def normalize_raw(
+    raw: np.ndarray,
+    axis: Optional[Union[int, Tuple[int, ...]]] = None,
+    output_range: Tuple[float, float] = SAM2_INPUT_RANGE,
+    dtype: Union[str, np.dtype] = "float32",
```
constantinpape: Why do we pass the `dtype`  as an extra argument here? Shouldn't this just be derived from `raw` as `raw.dtype` internally?

micro_sam/v2/normalization.py
```diff
@@ -0,0 +1,51 @@
...
+    lower_percentile: float = 2.0,
+    upper_percentile: float = 98.0,
+) -> np.ndarray:
+    """Percentile-normalize raw image data to the requested input range.
+
+    The lower and upper percentiles (2nd and 98th by default) are mapped to zero
+    and one and values outside this range are clipped.
```
constantinpape: Please extend the docstring so that the arguments are explained.

micro_sam/v2/normalization.py
constantinpape: Or is this the output dtype only? Then we should rename it.

constantinpape (commented):
I think this looks good now, only some cosmetic changes needed, see comments.

constantinpape:
I thought about it again and `(output_)dtype` and `output_range` are closely linked but currently there are no checks validating that they match. E.g. an int type with [0, 1] output range.

micro_sam/v2/normalization.py
anwai98: Sorted!

micro_sam/v2/normalization.py
anwai98: It's output dtype. Lemme rename it for now, let's finalize this when we meet later!

micro_sam/v2/normalization.py
anwai98, line 1: Revisit this!

anwai98 (commented):
FIX IT

anwai98:
This is sorted now ;) I'll go ahead and merge this once the CIs pass!

PR #1287, opened 2026-07-14
Add support for scribble prompts

anwai98:
I thought of this idea a while ago and decided to implement this now. The cool part -- it helps with neuron segmentation and weird structures by like a freaking lot (which is very cool improvement to observe instead of putting tons of points around).

I need to make a few final tests before I merge this!

Could you give this a quick shot? @constantinpape 

PS. Use the "Add path" (or line / poyline) tool in layer controls for scribbles, a "T" keybinding to switch between positive and negative scribbles!

### DIFF
--- a/micro_sam/sam_annotator/_annotator.py
+++ b/micro_sam/sam_annotator/_annotator.py
@@ -89,15 +89,40 @@ class _AnnotatorBase(QtWidgets.QScrollArea):
             )
             self._point_prompt_layer.border_color_mode = "cycle"
 
-        if "prompts" not in self._viewer.layers:
+        if "prompts" in self._viewer.layers:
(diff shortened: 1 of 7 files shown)


anwai98:
Oh I forgot to explain -- it's a simple design to have a "chain of prompts" derived in a gap subjected to the curvature and a few other parameters of the scribble drawn!

anwai98:
This is good to go from my side!

anwai98:
Okay, we have a thumbs up from CP here ;) I'll test this on Mac quickly before merging this!

anwai98:
Just checked over Mac. Works quite nice over there as well ;) I'll go ahead and merge this!

PR #1289, opened 2026-07-14
Fix SAM2 3d interactive segmentation on MPS

anwai98:
Somehow there was a weird in-memory embeddings forwarding issue which created patchy NaN masks -- which is the segmentation artefact @Marei33 reported!

This happens due to async transfer across CPU and MPS while recasting inputs around like a swing (because several `pytorch`-related stuff is not compiled in `mps`). I just offload the lightweight components to CPU and works fine in MPS since it's unified memory!

Will merge this once the tests pass!


anwai98:
I'll go ahead and merge this now!

PR #1290, opened 2026-07-14
Fix 3d embedding computation on lazy arrays in napari

anwai98:
@lufre1 this should fix the issue you reported by dragging in `s3` resolution from `zarr` directly into `napari`.

TLDR: I decided to handle dask arrays better now (i.e. how `napari` adopts the image layer when container ffs are dragged in).


anwai98:
Okay this is also sorted. I'll merge this!

PR #1291, opened 2026-07-14
Add disk-backed lazy embeddings for eager annotation workflow

anwai98:
Okay this is an ambitious attempt. I'll pair this with two use-cases: 1) one with @lufre1's, and 2) on my Macbook on larger volumes!

The eager loading of embeddings should work fairly well (tried and it seems to do the trick for me).

I'll leave this as a draft for now!

### DIFF
--- a/micro_sam/sam_annotator/_state.py
+++ b/micro_sam/sam_annotator/_state.py
@@ -90,6 +90,9 @@ class AnnotatorState(metaclass=Singleton):
     ndim: Optional[int] = None
     image_name: Optional[str] = None
     embedding_path: Optional[str] = None
+    # Path to an ephemeral on-disk embedding cache created when no save path is given (SAM2 volumes /
+    # tiled images), so embeddings stream from disk instead of filling RAM. Removed on reset / exit.
+    embedding_tmpdir: Optional[str] = None
     data_signature: Optional[str] = None
     skip_recomputing_embeddings: Optional[bool] = None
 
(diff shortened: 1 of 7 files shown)


anwai98:
I'll check tomorrow what's up with the ubuntu CI (reads like something weird to me)

anwai98:
I tried in on my Macbook now, works like a charm now (with a quick patch fix on streaming over tiled 3d embeddings -- needed to nest the streaming there). I think I'll simply go ahead and merge this once the tests pass, since this would help so much with trying the scribble prompt PR on Macbook! (otherwise it OOMs on the Macbook on whole Lucchi without this PR)

PS: @lufre1 I can confirm it works pretty neat. I would be very curious if it works on JupyterHPC. Could you try it once you have sometime? (I'll merge this latest today)

lufre1:
@anwai98 
tested it now on "dev" works now like a charm. Computing embeddings takes quite a few hours so testing is slow.

constantinpape:
> Computing embeddings takes quite a few hours so testing is slow

So we keep track of peformance bottlenecks: is this on CPU or GPU? How big is the volume and how many tiles are there in total? If it takes several hours on the GPU that would be a good use-case to optimize (we actually have some room for improvements by queuing input and output as in recent `torch-em` changes).

anwai98:
> tested it now on "dev" works now like a charm. Computing embeddings takes quite a few hours so testing is slow.

Yeah for the scale you are playing on -- over several slices with tiling, this is indeed a tricky one!

> So we keep track of peformance bottlenecks: is this on CPU or GPU?

Taking a good guess of what Luca was doing, it's RTX 5000 on JupyterHPC. 

> If it takes several hours on the GPU that would be a good use-case to optimize (we actually have some room for improvements by queuing input and output as in recent torch-em changes).

Yep, batching tiles, using multiple GPUs and using takeaways from `torch-em` are the next steps for me hehe! (will spawn an issue to work on it later)

lufre1:
> @anwai98 tested it now on "dev" works now like a charm. Computing embeddings takes quite a few hours so testing is slow.

found the reason why it's that slow... tmp/embeddings were stored in home directory, clogging it. i'll put it somewhere else and it should resolve the issue

i used a volume with approx size: 700z 600y 600x

anwai98:
> found the reason why it's that slow... tmp/embeddings were stored in home directory, clogging it. i'll put it somewhere else and it should resolve the issue

Ah yeah that's the on-the-fly workflow. If you change the filepath location to something where you have storage, this'll sort itself out ;)

lufre1:
now it works and for the volume i used it computes embeddings for approx 15min on that RTX5000. 
However sometimes the "batched" checkbox vanishes without any message in the terminal: 
image

anwai98:
> now it works and for the volume i used it computes embeddings for approx 15min on that RTX5000.
However sometimes the "batched" checkbox vanishes without any message in the terminal:

Yeah batched interactive segmentation is not supported with tiling!

constantinpape:
> Yep, batching tiles, using multiple GPUs and using takeaways from `torch-em` are the next steps for me hehe! (will spawn an issue to work on it later)

Great! 

> now it works and for the volume i used it computes embeddings for approx 15min on that RTX5000.

Ok, that sounds more realistic ;).

constantinpape:
> Ah yeah that's the on-the-fly workflow. If you change the filepath location to something where you have storage, this'll sort itself out ;)

Are they cleaned up after the viewer is closed, @anwai98 ?

anwai98:
> Are they cleaned up after the viewer is closed, @anwai98 ?

Yes!

PR #1292, opened 2026-07-15
Fix pixel and object classification geometry for SAM2 embeddings

anwai98:
Yeah I broke this while my unification attempt for segmentation and tracking annotators. Should be working now! ;)

I'll make a few final tests with models besides SAM2 for classification, and merge this!


anwai98:
Nice, works fine for others. Will merge this once the CIs pass!

PS. Thanks for the report @caroteu ;)

PR #1294, opened 2026-07-15
Add info window popup for micro-sam annotator on CPU

anwai98:
Straight-forward addition of information window for CPU users. Will merge this once the tests pass!

### DIFF
--- a/micro_sam/sam_annotator/_widgets.py
+++ b/micro_sam/sam_annotator/_widgets.py
@@ -2040,6 +2040,20 @@ class EmbeddingWidget(_WidgetBase):
             "info", "Embeddings have already been precomputed. Press OK to recompute the embeddings."
         )
 
+    def _maybe_warn_cpu(self):
+        """Show a one-time-per-session info popup when the annotator runs on the CPU."""
+        state = AnnotatorState()
+        if state.cpu_info_shown:
+            return
+        if str(util.get_device(self.device)) != "cpu":
+            return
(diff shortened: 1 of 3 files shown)


micro_sam/sam_annotator/_widgets.py
```diff
@@ -2102,6 +2116,9 @@ def __call__(self, skip_validate=False):
         )
         image_data = image.data
 
+        # Warn CPU users once per session that processing can be slow.
+        self._maybe_warn_cpu()
```
constantinpape: I would not show this warning if we just process 2d data, only for 3d embeddings or 2d data with a lot of tiles.

anwai98: I see. Lemme tag it to tiling and 3d then!

constantinpape: > I see. Lemme tag it to tiling and 3d then!

For tiling I would also make in condition on > 64 tiles or so. Anything less should still be fine on a CPU (or some other number you think makes more sense, but something like 4 - 20 tiles is totally fine on CPU)

anwai98: Good point. Lemme see if there's a tile counter I can tag this to!

anwai98: Sorted! ;)

anwai98:
Sorted. I'll go ahead and merge this!

PR #1297, opened 2026-07-16
Supports for napari 0.8 and bump version to 1.8.6

anwai98:
Just wanna align this with the latest napari release, i.e. stick to >=py311 and make PyQt6 our default Qt backend choice (since PyQt5 is EOL end of this year).


anwai98:
Lemme quickly try py313 and py314. The current builds are limited to py311-py312 and that's it.

anwai98:
Okay I did some local tests in my Macbook. They work for all python builds (py311-py314). I would endorse to go ahead with this! 

What do we think about this? @constantinpape 

PS. I'll also patch the feedstock later to build for py313 and py314.

anwai98:
The tests locally also finished from my side on Ubuntu -- looks good!

constantinpape (approved):

PR #1298, opened 2026-07-16
Test broader python range of 311-314 for CIs

anwai98:
Let's see how this goes. Not gonna merge this unless I am super confident!

### DIFF
--- a/.github/workflows/test-pip-installation.yaml
+++ b/.github/workflows/test-pip-installation.yaml
@@ -18,7 +18,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.12"]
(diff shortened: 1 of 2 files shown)


constantinpape:
I wouldn't expand the matrix in macOS. These tests take quite long and are brittle (I assume because the very limited RAM of the VM running it.)

anwai98:
This is just a dummy test -- to broaden the feedstock and pypi wheel builds. I'm probably not gonna merge it. Will come back to this in the evening!

anwai98:
The good thing here is -- my interests suffice here for now. I'll go ahead and close this (and open up builds for py311 -py314), and double test it locally on my Ubuntu and Mac machines!

PR #1299, opened 2026-07-17
Add random percentile normalization augmentation

anwai98:
WIP!


anwai98:
The tests will break but should be fine (will wait for the next `torch-em` release I'll trigger in some)

anwai98:
Sorted out. I'll go ahead and merge this!

PR #1303, opened 2026-07-21
Add batched and pipelined inference for SAM2

anwai98:
Lemme take a closer look and report the improved numbers. Still gonna be a WIP before we hit a proper review!


anwai98:
~Few more ideas for self:~
1. ~batch size spin box should not show up for CPU resources~
2. ~Call it "batch size" and that's it~

^ sorted!

anwai98:
Hi @constantinpape,

Sorry for the long wait. This is now ready for a review!

I did extensive checks over Mac, JupyterHPC GPUs and my own Linux. This works pretty neat now! ;) (and even for single hardware single batch size condition, the throughput is fairly fast(er))

micro_sam/v2/batched_inference.py
anwai98, line 1: Here's the main brain for this PR (the core functions for batching and pipelining stuff)

micro_sam/sam_annotator/_widgets.py
```diff
@@ -2179,7 +2210,9 @@ def __call__(self, skip_validate=False):
             show_info("Set an embeddings save path to cache the automatic segmentation state.")
 
         # Set up progress bar and signals for using it within a threadworker.
-        pbar, pbar_signals = _create_pbar_for_threadworker()
+        # Model and decoder preparation happens before the backend knows the tile / slice count, so
+        # start with an indeterminate status instead of leaving napari's activity display empty.
+        pbar, pbar_signals = _create_pbar_for_threadworker("Preparing image embeddings")
```
constantinpape: I would rather print `"Computing image embeddings"`

micro_sam/sam_annotator/_widgets.py
```diff
@@ -1788,6 +1793,23 @@ def _create_settings_widget(self):
         )
         setting_values.layout().addLayout(layout)
 
+        # Use a conservative default and let users opt into larger per-GPU batches when their
+        # workload and available memory benefit from them. Batching only helps on a GPU, so the
+        # control is hidden whenever the effective device is the CPU (the default of 1 is still
+        # used then). Visibility tracks the device dropdown, so it also updates when the user
+        # switches devices or when 'auto' resolves to the CPU.
+        self.batch_size = 1
```
constantinpape, line 1805: Right now the default is always set to 1. We earlier discussed to set it to a safe default derived from the available VRAM (if the user has a GPU). Did you decide against that?

micro_sam/v2/automatic_segmentation.py
```diff
@@ -118,6 +122,11 @@ def automatic_instance_segmentation(
         mode: The AIS post-processing mode, 'sparse' (flow) or 'dense' (multicut). Ignored for AMG.
         device: The device to run inference on.
         verbose: Whether to print progress.
+        batch_size: Explicit tile or slice batch size. Defaults to one; pass None for
```
constantinpape: I think "tile or slice batch size" is a bit misleading. We can either be more verbose: "The batch size used when running inference for multiple slices and/or batches" or less verbose and just say the batch size.

micro_sam/v2/automatic_segmentation.py
```diff
@@ -147,11 +156,28 @@ def automatic_instance_segmentation(
...
-                emb_predictor, raw, save_path=embedding_path, ndim=ndim,
-                tile_shape=tile_shape, halo=halo, verbose=verbose, lazy_loading=(ndim == 3),
+                emb_predictor,
+                raw,
+                save_path=embedding_path,
+                ndim=ndim,
+                tile_shape=tile_shape,
+                halo=halo,
+                verbose=verbose,
+                lazy_loading=(ndim == 3),
```
constantinpape: This is not directly related to the changes in this PR, but it would make sense to also enable lazy loading for large tile shapes, even in 2D. We would have the same memory issues there otherwise.

micro_sam/v2/batched_inference.py
constantinpape, line 1: Please add full doc strings to all public functionality here. You can also consider making less things public. See detailed comments below

micro_sam/v2/batched_inference.py
```diff
@@ -0,0 +1,1253 @@
...
+
+from micro_sam.util import _create_dataset_with_data, _create_dataset_without_data
+from .normalization import to_image
+from .util import Device, Devices
+
+
+STOP = object()
+
+
+class PipelineAborted(Exception):
```
constantinpape: I think these classes can all be private, i.e. `_PipelineAborted`, `_AtomicCounter`, etc.

They are internal implementation details and don't need to be part of the documented interface.

micro_sam/v2/batched_inference.py
```diff
@@ -0,0 +1,1253 @@
...
+
+
+def _model_device(model: torch.nn.Module) -> torch.device:
+    try:
+        return _normalize_device(next(model.parameters()).device)
+    except (AttributeError, StopIteration):
+        return torch.device("cpu")
+
+
+def resolve_devices(model: torch.nn.Module, devices: Devices = None) -> List[torch.device]:
```
constantinpape: Add full doc-string (including Args and Return) or make private.

micro_sam/v2/batched_inference.py
```diff
@@ -0,0 +1,1253 @@
...
+    if len(resolved) == 0:
+        raise ValueError("At least one inference device is required.")
+    if len(set(resolved)) != len(resolved):
+        raise ValueError(f"Inference devices must be unique, got {resolved}.")
+    if any(device.type == "cuda" for device in resolved) and not torch.cuda.is_available():
+        raise RuntimeError("CUDA inference was requested, but PyTorch CUDA support is unavailable.")
+    return resolved
+
+
+def prepare_models(
```
constantinpape: Make private?

micro_sam/v2/batched_inference.py
```diff
@@ -0,0 +1,1253 @@
...
+    models = []
+    for device in devices:
+        replica = model if device == source_device else deepcopy(model).to(device)
+        if hasattr(replica, "eval"):
+            replica.eval()
+        models.append((replica, device))
+    return models
+
+
+def release_model_replicas(model_devices: List[Tuple[torch.nn.Module, torch.device]]) -> None:
```
constantinpape: Make private?

micro_sam/v2/batched_inference.py
```diff
@@ -0,0 +1,1253 @@
...
+    if not measurements:
+        raise ValueError("At least one batch-throughput measurement is required.")
+    selected_batch, selected_throughput = measurements[0]
+    for batch_size, throughput in measurements[1:]:
+        if throughput >= selected_throughput * (1.0 + min_relative_improvement):
+            selected_batch, selected_throughput = batch_size, throughput
+    return int(selected_batch)
+
+
+def compute_auto_batch_sizes(
```
constantinpape: Add Args and Return

micro_sam/v2/batched_inference.py
```diff
@@ -0,0 +1,1253 @@
...
+        stop_event.set()
+        for thread in threads:
+            thread.join()
+
+    if error_box:
+        raise error_box[0]
+
+
+IMAGE_MEAN = (0.485, 0.456, 0.406)
+IMAGE_STD = (0.229, 0.224, 0.225)
```
constantinpape: This doesn't look like it belongs here.

micro_sam/v2/batched_inference.py
```diff
@@ -0,0 +1,1253 @@
...
+        f"Batch size {len(items)} ran out of memory on {device}; retrying with smaller batches.",
+        RuntimeWarning,
+        stacklevel=2,
+    )
+    left, left_size = _predict_with_oom_backoff(model, items[:split], device, predict_fn)
+    right, right_size = _predict_with_oom_backoff(model, items[split:], device, predict_fn)
+    return left + right, min(left_size, right_size)
+
+
+def run_batched_pipeline(
```
constantinpape: This should probably be private.

micro_sam/v2/batched_inference.py
```diff
@@ -0,0 +1,1253 @@
...
+    jobs: Iterable[Any],
+    model_devices: Sequence[Tuple[torch.nn.Module, torch.device]],
+    batch_sizes: Sequence[int],
+    load_fn: Callable[[Any], Any],
+    predict_fn: Callable[[torch.nn.Module, List[Any], torch.device], List[Any]],
+    write_fn: Callable[[Any, Any], None],
+    num_prefetch_workers: int = 4,
+    update_progress: Optional[Callable[[int], None]] = None,
+    progress_increment: Optional[Callable[[Any], int]] = None,
```
constantinpape, line 363: Shouldn't we also have `num_write_workers` here?

micro_sam/v2/batched_inference.py
```diff
@@ -0,0 +1,1253 @@
...
+        {
+            "features": features[-1][index:index + 1],
+            "high_res_feats": [feature[index:index + 1] for feature in features[:-1]],
+            "original_size": items[index]["original_size"],
+        }
+        for index in range(batch_size)
+    ]
+
+
+def compute_tiled_2d(
```
constantinpape: Either private or fully documented for all functions following too (I think private fits better since the user would call something more high-level)

micro_sam/v2/batched_inference.py
constantinpape: Also, I did not review everything in detail. And: I think more of the functions should expose `n_writer_threads`?

micro_sam/sam_annotator/_widgets.py
anwai98: Done!

micro_sam/sam_annotator/_widgets.py
anwai98: Yeah. I thought of doing this in a follow-up PR by benchmarking across GPUs.

micro_sam/v2/automatic_segmentation.py
anwai98: Good idea. I'd choose the verbose way (otherwise `batch_size` is a keyword which is used in several contexts around)

micro_sam/v2/automatic_segmentation.py
anwai98: On it!

constantinpape:
Since this is quite big and central I asked codex to review. Here's its findings (I didn't validate this myself, but it sounds reasonable)

## Findings

  1. [P1] The configured inference device is ignored by the new batched paths.
     micro_sam/v2/instance_segmentation.py:1002 resolves devices only from the new per-call devices argument. When that argument is omitted, it ignores
     self._device, which is the documented device passed to the segmenter constructor. The decoder helpers have the same issue: for example, micro_sam/
     v2/batched_inference.py:1079 accepts device but forwards only devices. Consequently, an explicit device="cpu" or device="cuda:1" can silently run
     on the model’s current device—or all visible GPUs—in full, 3D, and tiled inference. Use the configured single device as the fallback when devices
     is None.

  2. [P2] Invalid tiled z-block settings can silently return an empty prediction.
     micro_sam/v2/batched_inference.py:1181 converts z_block and z_halo but does not validate them before creating jobs. A negative z_block makes the
     range empty, so the function returns an all-zero prediction without running the decoder. A negative halo can also construct invalid context slices.
     The non-tiled implementation already rejects these values at line 1053; the tiled path should apply the same validation.

  3. [P2] Tile-ID partitioning can leave most selected GPUs idle.
     micro_sam/v2/prompt_based_segmentation.py:395 assigns work with tile_id % len(predictor_devices). Sparse active tiles commonly share the same
     residue—for example, tile IDs 0, 2, 4, 6 with two devices—so all work runs serially on one GPU. The 3D scheduler repeats this pattern at line 1104.
     For 2D, distribute the active job list round-robin. For persistent 3D tile states, assign each tile round-robin when it is first activated and
     retain that affinity.

micro_sam/sam_annotator/_widgets.py
constantinpape: Ok, I agree! It's better to do this in a follow up rather in this already big PR.

micro_sam/v2/batched_inference.py
anwai98: On it!

micro_sam/v2/batched_inference.py
anwai98: Yes and yes. On it!

micro_sam/v2/batched_inference.py
anwai98: Done (went private)

micro_sam/v2/batched_inference.py
anwai98: Done!

micro_sam/v2/batched_inference.py
anwai98: Done!

micro_sam/v2/batched_inference.py
anwai98: Done!

micro_sam/v2/batched_inference.py
anwai98: There are other traces of it. I'll use them instead. Thanks for spotting!

micro_sam/v2/batched_inference.py
anwai98: Done!

micro_sam/v2/batched_inference.py
anwai98: I'd go for private too. Done!

micro_sam/v2/batched_inference.py
anwai98: Yep, will plug it and expose it!

anwai98:
> Since this is quite big and central I asked codex to review. Here's its findings (I didn't validate this myself, but it sounds reasonable)

Nice, thanks @constantinpape

I'll give it another round of closer look and review once more before pinging you for the next round!

anwai98:
Okay all the changes considered are pushed now. I'll test them on my Macbook and JupyterHPC tomorrow!

anwai98:
Okay we are ready for the next round of review. Works pretty great! (across my three demo systems) @constantinpape

micro_sam/sam_annotator/_state.py
```diff
@@ -252,11 +252,11 @@ def initialize_predictor(
...
                 self.embedding_tmpdir = save_path
 
-            # For SAM2 volumes, load the embeddings lazily from the zarr so the high-resolution
-            # per-slice features stay on disk and are streamed one slice at a time during tracking.
+            # For SAM2 volumes and tiled images, load the embeddings lazily from the zarr so the
+            # high-resolution features stay on disk and are streamed one slice / tile at a time.
             # This keeps memory bounded for large volumes (materialising all slices costs
             # ~200 MB/slice and OOMs); it only applies when the embeddings are cached on disk.
-            lazy_loading = self.is_sam2 and ndim == 3 and isinstance(save_path, str)
+            lazy_loading = needs_disk_cache and isinstance(save_path, str)
```
constantinpape: Is the second part of this check necessary? I think `isinstnance(save_path, str)` is always `True` here because you set it to `self.embedding_tmpdir` (which I assume is a str) above.

anwai98: Good catch -- and yes that's redundant! (fixed it)

anwai98:
Okay, we are ready here. What do you think? @constantinpape

constantinpape (approved):

PR #1304, opened 2026-07-23
Treat cleared checkpoint path as None in annotator widgets

anwai98:
Reported by @ericvillard-paris at https://github.com/computational-cell-analytics/micro-sam/commit/7a77b410303683bcf9faa6a483a3d614c9889982#commitcomment-193484308 (thanks for spotting 😁)

TL&DR on the core issue: when a provided custom weights path is cleared, it then registers it as `""` instead of `None`, so `get_sam_model` goes crazy about it and tries to load an empty checkpoint path and crashes (instead of falling back to the default model). I went ahead and made it modular!

PS. I'll also add @ericvillard-paris as a co-author to my commit in this PR (since you diagnosed and found this issue -- thanks once again 🥳)

I'll merge this once the tests pass, and probably bump the version in a follow-up PR!

### DIFF
--- a/micro_sam/sam_annotator/_widgets.py
+++ b/micro_sam/sam_annotator/_widgets.py
@@ -85,10 +85,10 @@ class _WidgetBase(QtWidgets.QWidget):
             label.setToolTip(tooltip)
         layout.addWidget(label)
         param = QtWidgets.QLineEdit()
-        param.setText(value)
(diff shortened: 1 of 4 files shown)


anwai98:
Nice, GTM once the tests pass!

anwai98:
Awesome. Lemme go ahead and merge this!

PR #1307, opened 2026-07-24
Fix scribble negative toggle and add tracking scribble support

anwai98:
I missed a corner case. Lemme check it out on the Macbook!


anwai98:
Poof, managed to fix it (freaked out there for a sec haha). Will merge it once the CI pass!

PR #1312, opened 2026-07-27
Allow batch size based on available VRAM and stop storing redundant embeddings

anwai98:
Works pretty good. I'll check over CPU and Macbook and then merge this!

### DIFF
--- a/micro_sam/v2/normalization.py
+++ b/micro_sam/v2/normalization.py
@@ -4,14 +4,43 @@ from typing import Optional, Tuple, Union
 
 import numpy as np
 
-from torch_em.transform.raw import normalize_percentile
-
 # Persist the preprocessing (normalization + resize) policy in embedding caches so incompatible
 # features are not silently reused. The 2d image path uses per-channel min-max (via `to_image`); the
 # 3d / video path uses percentile normalization with the tensor resize used in training. The video
 # resize suffix invalidates embeddings created by the former skimage path, which did not match it.
 IMAGE_PREPROCESSING = "minmax_per_channel"
-VIDEO_PREPROCESSING = "percentile_2_98_per_channel_torch_resize_v1"
+# v2 stores one shared positional encoding per volume / tile instead of one per slice.
+VIDEO_PREPROCESSING = "percentile_2_98_per_channel_torch_resize_v2"
+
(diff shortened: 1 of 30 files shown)


anwai98:
Okay CPU and JupyterHPC works quite nice! ;) I need to check for MPS tomorrow and confirm what's up there, and is the VRAM pairing respected there!

micro_sam/v2/normalization.py
```diff
@@ -4,14 +4,43 @@
...
 # features are not silently reused. The 2d image path uses per-channel min-max (via `to_image`); the
 # 3d / video path uses percentile normalization with the tensor resize used in training. The video
 # resize suffix invalidates embeddings created by the former skimage path, which did not match it.
 IMAGE_PREPROCESSING = "minmax_per_channel"
-VIDEO_PREPROCESSING = "percentile_2_98_per_channel_torch_resize_v1"
+# v2 stores one shared positional encoding per volume / tile instead of one per slice.
+VIDEO_PREPROCESSING = "percentile_2_98_per_channel_torch_resize_v2"
+
+
+def _normalize_percentile(
```
constantinpape, line 16: Is there a reason to remove the `torch_em` import and reimplement?

anwai98: Yes, to avoid the import overhead (I timed it and realized it takes a couple of seconds) -- in the direction to speed up tooling experience!

anwai98:
Hmm I did some experiments on Macbook -- seems like it's memory-bandwidth limited :( 

i.e. it's better to use `batch_size=1` than bump it up -- reflects in runtime speed!

anwai98:
The tests are passing (awesome stuff ;). Thanks @constantinpape for the detailed review forward! 🥳 

I'll quickly merge this once the Mac CIs pass and then test it around machines to confirm one final time if all works as expected!

micro_sam/v2/normalization.py
constantinpape: Makes sense. In general, best to make an explicit comment in cases like this.

anwai98: Good point. I'll document this in next round around the function for us to remember!

PR #1313, opened 2026-07-30
Support tracking with existing segmentations in tracking annotator

anwai98:
Gonna check first of all how this works and then come back with an answer / updates!


anwai98:
I know where the tests are broken. Let's not worry much about it since it's more of a prototype for now!

Issue #768, opened 2024-10-29
Improve irregular masks generated via automatic / manual / pre-loaded segmentation


anwai98:
I thought about this at I2K. Let's say there is a situation where the automatic segmentation does an okay job to segment objects, and there are some objects where I think it could do a better job.

Can we think of extending interactive segmentation in a way that the model takes the incomplete segmentations for an object paired with point / box prompts and considering both as input prompts (masks and additional input prompts), and returns a segmentation?

anwai98:
Another +1 on this we observed was from recent workflows @Marei33 tried out on her data. I'll put a discussion label on this to think if there's an elegant way (if at all) to consider this pipeline!

Issue #731, opened 2024-10-10
Add GUI for extending 2d instances to 3d?


constantinpape:
We may consider adding a GUI element for prompting volumetric segmentation with the results of 2d automatic segmentation.
 
Currently not implemented in the UI but we have the underlying functionality for it. The function [segment_mask_in_volume](https://github.com/computational-cell-analytics/micro-sam/blob/master/micro_sam/multi_dimensional_segmentation.py#L86) takes a (2D) mask as input and then uses it as prompt for 3D segmentation, so what you want is quite easy to implement (just outlining the idea):
```python
from micro_sam.multi_dimensional_segmentation import segment_mask_in_volume

# assuming we have the array 'segmentation' and unique ids 'seg_ids' 
for seg_id in seg_ids:
  mask = segmentation == seg_id
  mask_3d = segment_mask_in_volume(
    segmentation=mask,
    predictor=predictor,  # this is the sam model
    image_embeddings=image_embeddings,  # these are the precomputed 3d image embeddings - important to make this efficcient
    segmented_slices=[z],  # the slice id for which you ran auto segmentation, i.e. which contains the mask
   stop_lower=False, stop__upper=False,  # these both have to be set to False for this use-case
   iou_threshold=0.5,  # when to stop segmentation depending on IOU of adjacent slices
   projection="single_point",  # different projection options might affect the result, single_point is the simplest and often works well. 
  )
  ...  # do something to keep track of the 3d masks
```
I am also not quite sure what would be the best way to integrate this into the GUI to avoid more options and complexity for the user.

cc @psobolewskiPhD

Issue #383, opened 2024-02-06
Refining masks with micro-sam.


constantinpape:
There are different ways for refining existing masks with `micro_sam`.

The easiest option would be to derive point prompts from the centers of the masks and then prompt the model with these points. 

The function [batched_inference](https://github.com/computational-cell-analytics/micro-sam/blob/master/micro_sam/inference.py#L17) can be used for this.

Here is some (non-tested!) code for this, using skimage to derive the point prompts.

```python
import numpy as np
from micro_sam.util import get_sam_model
from micro_sam.inference import batched_inference
from skimage.measure import regionprops

image = ...  # <- this is the 2D! image
initial_segmentation = ...   # <- this is the segmentation you want to refine

props = regionprops(initial_segmentation)
points = np.array([prop.centroid for prop in props])[:, ::-1]   # The coordinates of the centroids need to be reversed to match the convention of SAM.
point_labels = np.ones(len(points), dtype="int")  # <- All prompts are positive.

predictor = get_sam_model(model_type="vit_b_lm")  # <- you can control which model is used with the model type argument. 
# See the function signature of get_sam_model for details.

refined_segmentation = batched_inference(
  predictor, image,
  batch_size=32,  # This controls how many points are processed at once, lower it if you get memory issues 
  points=points,
  point_labels=point_labels,
  return_instance_segmentation=True
)
```

Another possible strategies is to derive bounding boxes from the segmented objects and use these for prompts instead.
This could be done by passing the [boxes](https://github.com/computational-cell-analytics/micro-sam/blob/master/micro_sam/inference.py#L21) argument.

Note that this code will only for 2D. It is possible to extend this to 3D, but I would suggest to start in 2D first and once this is working well I can give hints for how to extend it to 3D.

cc @Nal44

Nal44:
Great !!! thanks a lot :) 

I will try to make it "napari compatible" using viewer.layers as well :)

Thanks !!!

Nal44:
my attempt  on a 2d image : 

import numpy as np
from micro_sam.util import get_sam_model
from micro_sam.inference import batched_inference
from skimage.measure import regionprops

image = viewer.layers[0]
initial_segmentation = viewer.layers[1]

# Get the labels from the napari Labels layer
labels = initial_segmentation.data

# Convert labels to a numpy array
labeled_image = np.array(labels)

# Get the image data as a NumPy array
image_data = viewer.layers[1].data

# Convert to a NumPy array
[...]

PR #1314, opened 2026-07-31
Add existing Labels layers as SAM mask inputs

arthursw:
## Summary

- Add compatible napari Labels layers as mask inputs for interactive segmentation.
- Treat every nonzero label ID as an independent SAM mask prompt and preserve IDs in the merged result.
- Assign point, scribble and shape corrections per input object, so different objects can receive different corrections in one Segment Object operation.
- Let users commit satisfactory input masks unchanged, without running SAM or computing embeddings.

## Motivation

This enables segmentations created by Cellpose or other napari tools to become first-class inputs to micro-sam.
Users can combine an existing mask with positive or negative corrections when refinement is useful, while masks that are already satisfactory can bypass SAM entirely.

Addresses #768.
Related to #383 and #731.

## User-facing behavior

The annotator includes an **Existing Segmentation Inputs** checklist with **Select all** and **Unselect all** controls.
Internal `current_object`, `auto_segmentation` and `committed_objects` layers are excluded so they cannot be selected accidentally; users can duplicate such a layer in napari if they intentionally want to reuse it.

When at least one Labels layer is selected, **Segment Object [S]** uses every selected object as a separate SAM mask prompt and writes the merged result to `current_object`.
With no selected mask input, the action retains the existing point and shape prompt behavior.
This replaces an earlier separate refinement button because manual UX testing showed that two overlapping segmentation actions were confusing.

The **Target object ID** control is populated from the IDs in the selected Labels layers.
Each point or shape stores its own target ID; selecting prompts updates the control, changing the control updates all selected prompts, and mixed selections are shown explicitly.
If the inputs contain only one ID, all current and future corrections are assigned to it automatically.
Every correction must target an ID present in the selected mask inputs.

The point prompt selector is now grouped under **Point and scribble prompts** and is labeled **Prompt type**, with `Positive (include)` and `Negative (exclude)` choices.
It updates selected or future points and scribbles.
Rectangles, ellipses and polygons remain positive because SAM box and dense shape prompts do not have negative semantics.

**Commit selected masks unchanged** copies the selected masks directly into `committed_objects` without computing image embeddings.
This is useful when an external segmentation is already correct or when SAM mask-prompt downsampling would remove fine structures.

## Label and geometry semantics

- Every nonzero label value is an object ID.
- Equal IDs from multiple selected layers are unioned.
- Overlapping different IDs are rejected with an actionable error.
- Source Labels layers are never modified.
- Input and image transforms must match; masks are not resampled.
- A trailing one-pixel surplus is cropped per spatial axis to accommodate the size mismatch produced by newly created napari Labels layers.
- Other shape mismatches remain errors.
- Direct and refined commits preserve sparse IDs and reject collisions atomically.

## 2D and 3D behavior

Each object is sent to SAM independently as its own binary mask prompt, together with only the point, scribble and closed-shape corrections assigned to that object.
This retains object identity and prevents one object from becoming a foreground hint for another.

For 3D inputs, **Refine all occupied slices** performs 2D mask refinement on each occupied slice and keeps the existing z extent.
**Propagate from seed slices** selects the largest cross-section for each object, refines that seed plus any assigned correction anchors, and propagates it through the volume.
SAM-v1 uses slice-to-slice projection, while SAM2 conditions its video predictor with the refined seeds.
Tiled prompt paths are supported for both model generations.

The prompt-based segmentation helpers route mask prompts through SAM-v1, tiled SAM-v1, image-style SAM2 and SAM2 video propagation.
Prompt collection can filter every prompt type by object ID and, for 3D data, by slice.

## Limitation

SAM converts each 2D mask prompt to its 256 × 256 low-resolution representation.
Fine structures may therefore be lost during refinement.
The UI warns users before refinement and the unchanged-commit action provides a lossless alternative.

## Development progression

1. Added Labels-layer collection, validation, per-ID mask extraction and direct unchanged commits.
2. Unified mask refinement with the existing Segment Object action, excluded micro-sam output layers, and added the safe one-pixel crop for new napari Labels layers.
3. Reorganized the controls and replaced the single global correction target with per-prompt object IDs after manual workflow testing and review.
4. Added selection synchronization and regression coverage for mixed targets, cross-layer defaults, prompt polarity preservation and 3D routing.

## Documentation and tests

The annotation documentation explains layer selection, ID merging, geometry validation, per-prompt correction routing, unchanged commits, the 256 × 256 warning, and both 3D modes.

Tests cover input collection and validation, deterministic overlap handling, atomic commits, checklist behavior, per-object 2D and 3D prompt routing, mixed prompt selections, napari trailing-pixel cropping, tiled SAM-v1 prompts, SAM2 mask conditioning and 3D seed propagation.

Local validation:

- Consolidated focused suite: 78 tests passed.
- Additional 3D per-object routing regression: passed.
- Python compilation: passed.
- Ruff checks, excluding the repository's existing long-line style: passed.
- `git diff --check`: passed.
- Manual napari smoke test of the complete mask-input workflow: passed.

The GitHub test matrix requires a maintainer to approve workflows from the fork.

### DIFF
--- a/doc/annotation_tools.md
+++ b/doc/annotation_tools.md
@@ -56,6 +56,23 @@ When `commit_path` is given the results will automatically be saved there.
 
 Point prompts and box prompts can be combined. When you're using point prompts you can only segment one object at a time, unless the `batched` mode is activated. With box prompts you can segment several objects at once, both in the normal and `batched` mode.
 
+### Using existing segmentations
+
+The `Existing Segmentation Inputs` section accepts any compatible napari Labels layer, including segmentations created by Cellpose or another napari plugin.
+Use the checkboxes or the `Select all` and `Unselect all` buttons to choose inputs.
+All nonzero label IDs are treated as objects: equal IDs from multiple selected layers are unioned, while overlapping different IDs must be resolved in napari before continuing.
+The selected layers must have the same spatial shape and data-to-world transform as the selected image; micro-sam does not resample them.
+
+`Refine with SAM` processes every selected object independently and writes the merged result to `current_object`.
+If point or shape corrections are present and the inputs contain several objects, choose the single object ID they apply to.
+The normal `Commit` action then replaces those IDs in `committed_objects` without renumbering them.
+SAM converts each 2D mask prompt to its 256 by 256 low-resolution mask representation, so fine structures may be lost during refinement.
+The source Labels layers are never changed.
+
+`Commit input masks unchanged` bypasses SAM and the embedding computation.
+Use it when an input mask is already satisfactory.
+It preserves the input IDs and atomically unions equal IDs into `committed_objects`; a collision with a different committed ID is rejected.
+
 Check out [the video tutorial](https://youtu.be/9xjJBg_Bfuc) for an in-depth explanation on how to use this tool.
 
 
@@ -81,6 +98,12 @@ Most elements are the same as in [the 2d annotator](#annotator-2d):
 8. The menu for clearing the current annotations. If `all slices` is set all annotations will be cleared, otherwise they are only cleared for the current slice.
 
 You can only segment one object at a time using the interactive segmentation functionality with this tool.
+
+Existing 3D Labels layers can be refined in two ways.
+`Refine all occupied slices` runs 2D mask refinement on each slice where an object exists and preserves the input z-extent.
+`Propagate from seed slices` chooses each object's largest cross-section as its seed (the lowest z wins ties), refines the seed, and propagates it through the selected z-range.
+For SAM-v1 this uses slice-to-slice mask projection; for SAM2 it conditions the video predictor with the already refined seed.
+A 3D point or shape correction is routed to the chosen object ID; in occupied-slice mode it must lie on an occupied slice, while propagation mode can use a positive cue on an empty slice as another anchor.
 
 Check out [the video tutorial](https://youtu.be/nqpyNQSyu74) for an in-depth explanation on how to use this tool.
 
--- a/doc/annotation_tools.md
+++ b/doc/annotation_tools.md
@@ -65,14 +65,18 @@ All nonzero label IDs are treated as objects: equal IDs from multiple selected l
 The selected layers must have the same data-to-world transform as the selected image.
 An input that is exactly one trailing pixel larger on one or more axes is cropped to the image extent, which accommodates new Labels layers created by napari; other shape mismatches are rejected and micro-sam does not resample them.
 
-When mask inputs are selected, `Segment Object` (or `S`) processes every selected object independently and writes the merged result to `current_object`.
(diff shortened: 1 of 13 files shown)


arthursw:
Note: this was fully generated with GPT 5.6 Sol with high level of reasoning. If this does not comply with your code of conduct ; I can delete the PR.
Also, I will soon make an update of the PR ; to allow better control on points/shape corrections.

anwai98:
Hi @arthursw,

First of all, thanks for your interest in `micro-sam`! 🥳 

Quick reply on something important: currently, `dev` is heavily under active development (working towards our next generation tool and models for `micro-sam`)

I'm happy to take a PR, but I'd like to go with a route of opening an issue first.

That doesn't mean you should close this PR (yet), but let's do it this way:
1. Open a relevant issue stating clearly what is the feature you have a request for and what's your expectation.
2. Define if it is related to `micro-sam` / `sam` or `micro-sam2` / `sam2` (or both) -- this would define where should we base the changes to (although I'd prefer to stick to supporting changes only for `micro-sam` -- since we are a few weeks aways from an alpha release for `micro-sam2`)
3. Once the issue describes the issue, the PR body text (in minimal words) should explain what files are changes and what's the intention for any major changes?

If you are up for it, I'd be happy to take a look at the issue you created and continue the discussion there 😉

Lemme know once it's there!

arthursw:
Hi @anwai98,

Thanks for your quick answer and interest.
I just created [a quick issue](https://github.com/computational-cell-analytics/micro-sam/issues/1315) describing the feature request.

I didn't know about micro-sam2 ; and I think the request is relevant for both SAM and SAM2 ; so I used general terms.

Here is a screenshot of the current version of the PR:

image

You can see the "Existing Segmentation Inputs" section enabling to select Label layers ; to commit them directly or use them as SAM inputs.

anwai98:
> I just created https://github.com/computational-cell-analytics/micro-sam/issues/1315 describing the feature request.

Amazing, thanks @arthursw 🥳 

> I didn't know about micro-sam2 

Haha it's still under sincere active development, so would take a couple more weeks before an announcement, but we're definitely making a formal release this summer!

> I think the request is relevant for both SAM and SAM2 ; so I used general terms.

Nice, good to know!

And re: screenshot: thanks, that helps for sure!

I'll come back to your issue and look at things carefully over the weekend (I need to finish some other things now before I go afk for the evening hehe)!

arthursw:
I just updated the PR ; here is the new GUI:

image


I updated the "promp: positive/negative" and the "Target object ID: N" inputs ; to make them more explicit.

It is now possible (and desired) to affect one ID to each point prompt.

arthursw:
Nice that you are working on a new release ;-)

And it's nice if you can have a look at this PR ; this weekend, or in the coming days.

Thanks ;-)

PR #1316, opened 2026-08-04
Update evaluation pipeline for quantitative results

anwai98:
Mostly clean-up and aligning stuff with GUI. WIP!


anwai98:
Nice, this is sorted out to my liking now, and all works. There are some tiny things here and there that I will improve later on!

PR #1321, opened 2026-08-09
Several training improvements

anwai98:
Okay this was a good clean-up and fixing round. Looks good to me. I'll let Codex review another round and see if anything comes out of it!


anwai98:
I tested with a quick single GPU run -- I noticed clear improvement even for a super short run. I'll merge this so that all is in `dev` and I can get everything in one place!

PR #1327, opened 2026-08-13
Set the UNETR decoder to train and eval mode explicitly

### LINKED ISSUE #1325: Nitpick: explicitly set unetr to train and eval for correct batchnorm behaviour.
I think the UNETR decoder currently stays in its default train mode, even when running validation. It uses InstanceNorm everywhere without tracking stats, so its not a problem. But it might be an idea to future proof this a bit by explicitly setting train and eval modes, here: 

https://github.com/computational-cell-analytics/micro-sam/blob/3628539c0ab6dc9864a23e93b3c912c4de708b83/micro_sam/training/joint_sam_trainer.py#L86

and 

https://github.com/computational-cell-analytics/micro-sam/blob/3628539c0ab6dc9864a23e93b3c912c4de708b83/micro_sam/training/joint_sam_trainer.py#L133

anwai98:
Closes https://github.com/computational-cell-analytics/micro-sam/issues/1325

Good catch @mdiephuis 😉

### DIFF
--- a/micro_sam/training/joint_sam_trainer.py
+++ b/micro_sam/training/joint_sam_trainer.py
@@ -84,6 +84,7 @@ class JointSamTrainer(SamTrainer):
 
     def _train_epoch_impl(self, progress, forward_context, backprop):
         self.model.train()
+        self.unetr.train()
 
         input_check_done = False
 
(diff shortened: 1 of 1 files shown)


anwai98:
I'll merge this once the tests pass!

anwai98:
Also, I added you here as co-committer to the changes in this PR (I hope that's alright) 😁 @mdiephuis

PR #1328, opened 2026-08-13
Bump version to 1.8.9 and pin napari<0.9

anwai98:
Gonna bump a new version with changes in logging and validation part of trainer, and also pin down napari (since the new version is coming soon and I wanna check it first and avoid unknown broken issues)

Will merge this once the CIs pass!


ovalerio:
Thank you @anwai98 for sharing your thoughts on pinning down the *napari* version. 🙏

At first when I read about the pin I was wondering if there was some breaking update introduced in *napari* that I need to be aware. Then I came here and realise is just you being extra nice and avoiding everyone else a bad day. 👏💪

anwai98:
Thanks for your kind words @ovalerio 😄 
It's just me trying to keep up and stay on my toes to avoid any future troubles for many! (thanks for noticing)

PR #1334, opened 2026-08-19
Update hvit_t_cells to new model version

anwai98:
The new model is so freaking much better in a lot of ways. I'll go ahead and merge this right away so that it's the default new model we are gonna profit from (will remove the old models jlt).

Will merge this once the CIs pass! (and pull this inside other important PRs)


anwai98:
Something's going on with the Qt library setup action. I'm not gonna wait for it -- I wanna test things right away (and Windows passes anyways)

+nothing should break CIs here essentially. I'll merge this straight up!

PR #1335, opened 2026-08-19
APG Code Review

constantinpape:
This contains changes from a code review form CODEX. The points are all rather minor but in my opinion most make sense to integrate, some are legit small bug fixes (removing duplicate normalization, applying size filter for cellpose data, fixing the progbar for 3D APG). I am not sure if we want the checkpoint logic in eval from the first commit --- can be removed if you dislike it @anwai98 .


anwai98 (approved):
Nice, thanks @constantinpape. I skimmed through it, and some are good redundancy handles.

PR #1337, opened 2026-08-20
Add automatic instance segmentation to the bioimageio model export

nilsmechtel:
### Motivation

The models exported with `micro_sam.bioimageio.export_sam_model` currently only support prompt-based segmentation. The instance segmentation decoder is attached to the exported model as a separate file, but it is not part of the model's forward pass. As a result, tools that consume the bioimageio format, including the model test-run on bioimage.io, cannot run automatic instance segmentation (AIS) with these models, and calling the model without prompts fails.

### What this PR does

- `PredictorAdaptor` can now load a combined checkpoint in the micro_sam finetuning-checkpoint format (`{"model_state": ..., "decoder_state": ...}`). If a decoder state is present, the UNETR decoder is created via `micro_sam.instance_segmentation.get_decoder`. Plain SAM state dicts still load as before, so already published models keep working with the updated adaptor.
- Calling the adaptor **without any prompt inputs** now runs AIS via `InstanceSegmentationWithDecoder` (decoder prediction and seeded watershed with default parameters) instead of failing. The instances are returned as a stack of binary object masks, so the output signature (`masks`, `scores`, `embeddings`) and therefore the model spec are unchanged. Prompt-based segmentation behaves exactly as before.
- `export_sam_model` now:
  - fetches the corresponding decoder from the model registry when exporting a registry model (e.g. `model_type="vit_b_lm"`), or uses the `decoder_state` of a finetuning checkpoint as before,
  - saves the SAM state and the decoder state in a single combined weight file (the decoder is additionally kept as a separate attachment, since `micro_sam`'s registry downloads the decoder from that file),
  - maps derived model types such as `vit_b_lm` to the base architecture type (`vit_b`) for the adaptor kwargs, so registry models can be exported directly,
  - writes `micro_sam` into the exported `environment.yaml` when the model contains a decoder, since the bundled architecture uses `micro_sam` for AIS,
  - checks the AIS path (prediction without prompts) as part of `_check_model`.
- A new test `test_model_export_with_decoder` covers exporting a registry model with its decoder.

### Validation

- The new unit test `test_model_export_with_decoder` passes (it exports `vit_b_lm` from the model registry, and `_check_model` includes the new AIS check).
- The package exported for `vit_b_lm` was validated with `bioimageio.core.test_model`: status `passed`.
- Prediction with only the image input (no prompts and no precomputed embeddings) runs AIS end to end and returned 195 instances on the synthetic test image, covering 17 percent of the image (the synthetic data is generated with a volume fraction of 0.15).
- The exported package was additionally uploaded as a staged draft model on bioimage.io and tested there with the BioEngine model runner inside the model's own conda environment built from the exported `environment.yaml` (i.e. with `micro_sam` installed): all checks passed, including reproducing the test outputs.

### Notes

- AIS runs with the default parameters of `InstanceSegmentationWithDecoder.generate()`. Exposing the post-processing parameters (thresholds, smoothing, min_size) through the exported model could be a follow-up, e.g. via architecture kwargs.
- Since the output signature is unchanged, updating the published bioimage.io models to support AIS only requires re-exporting them with this code and publishing new versions.

### DIFF
--- a/micro_sam/bioimageio/model_export.py
+++ b/micro_sam/bioimageio/model_export.py
@@ -1,4 +1,5 @@
 import os
+import shutil
 import tempfile
 from pathlib import Path
 from typing import Optional, Union
@@ -41,6 +42,12 @@ DEFAULTS = {
 ARBITRARY_SIZE = spec.ParameterizedSize(min=1, step=1)
 
 
+def _get_architecture_model_type(model_type):
+    # Finetuned models like 'vit_b_lm' use the architecture of the base model type,
+    # which is encoded in the first five characters, e.g. 'vit_b'.
+    return model_type[:5]
+
+
 def _create_test_inputs_and_outputs(image, labels, model_type, checkpoint_path, tmp_dir):
 
     # For now we just generate a single box prompt here, but we could also generate more input prompts.
@@ -64,8 +71,8 @@ def _create_test_inputs_and_outputs(image, labels, model_type, checkpoint_path, 
         [_compute_logits_from_mask(labels == 1), _compute_logits_from_mask(labels == 2)]
     )[None]
 
-    predictor = PredictorAdaptor(model_type=model_type)
-    predictor.load_state_dict(torch.load(checkpoint_path))
+    predictor = PredictorAdaptor(model_type=_get_architecture_model_type(model_type))
+    predictor.load_state_dict(torch.load(checkpoint_path, map_location="cpu", weights_only=True))
 
     input_ = util._to_image(image).transpose(2, 0, 1)[None]
     image_path = os.path.join(tmp_dir, "input.npy")
@@ -131,7 +138,10 @@ def _get_checkpoint(model_type, checkpoint_path, tmp_dir):
     if checkpoint_path is None:
         model_registry = util.models()
         checkpoint_path = model_registry.fetch(model_type)
-        return checkpoint_path, None
+        # If the registry also has an instance segmentation decoder for this model we fetch it too.
+        decoder_name = f"{model_type}_decoder"
+        decoder_path = model_registry.fetch(decoder_name) if decoder_name in model_registry.registry else None
+        return checkpoint_path, decoder_path
 
     # Otherwise we have to load the checkpoint to see if it is the state dict of an encoder,
     # or the checkpoint for a custom SAM model.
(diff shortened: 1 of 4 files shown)


anwai98 (approved):
Hi @nilsmechtel,

Thanks for the PR. Looks good to me. I’ll merge it once the tests pass! 😁

PS. I also added regression tests and a validation script using some example data.
PS2. If you’d like a new release with the BioImage.IO export support, let me know and I’ll make one right away!

nilsmechtel:
A note on the runtime dependencies: the watershed postprocessing that the exported models use for AIS is now also proposed as a utility in bioimageio.core, see bioimage-io/core-bioimage-io-python#501. Once that lands in a release, the exported models no longer need the `micro_sam` package at runtime. The bundled architecture can build the UNETR decoder directly with `torch_em` (inlining the small `get_unetr` / `DecoderAdapter` logic) and call `bioimageio.core.microsam_watershed` for the postprocessing. The exported `environment.yaml` for decoder models then shrinks to `segment-anything`, `torch_em`, `bioimageio.core` and `scikit-image`, which avoids installing the napari / Qt stack that the `micro_sam` conda package pulls in.

I have prepared that variant on top of this branch: [`ais-export-lite-deps`](https://github.com/nilsmechtel/micro-sam/compare/add-ais-bioimageio-export...nilsmechtel:micro-sam:ais-export-lite-deps) (commit nilsmechtel/micro-sam@57d191e). It passes the same validation: both unit tests, `bioimageio.core.test_model` on the exported `vit_b_lm` package, and the prompt-free prediction returns the identical 195 instances on the synthetic test image.

Happy to switch this PR to the lighter variant, either directly or as a follow-up once the bioimageio.core release containing the watershed is out. Let me know which you prefer.

anwai98:
Hi @nilsmechtel,

> Happy to switch this PR to the lighter variant, either directly or as a follow-up once the bioimageio.core release containing the watershed is out. Let me know which you prefer.

Let's do it as a follow-up once the watershed PR is settled!

For now, I'd go ahead and merge this! (unless you have more things to push here?)

nilsmechtel:
And a new release with the BioImage.IO export support would be fantastic! 👌🏼

Do you need any help with updating the 6 micro SAM models in the BioImage Model Zoo?

anwai98:
> And a new release with the BioImage.IO export support would be fantastic! 👌🏼

Awesome, I'll make a release right away!

> Do you need any help with updating the 6 micro SAM models in the BioImage Model Zoo?

If you want, I'd love to have a helping hand with that! 😁

PR #1340, opened 2026-08-21
Generalize and extend refinement round for APG

constantinpape:
This adds a generalized refinement stage after the APG, which supports different modes like box refinement (which was already there), or points derived from suppressed points and other instance centers used as negatives (which can be combined with a box).
 
The refinement yields some overall gains, yet not consistent across datasets and at a significant runtime increase.
The two most promising settings are `point + boxes` with 1 positive and 6 negative points and a mask consistency insertion gate, which yields large gains for some datasets but small regressions for others, and `boxes` (with the same insertion gate) which yields consistent modest gains.

It makes sense to adopt this, but for now we shouldn't make it the default.

So far only for 2d (non-tiled); extension to tiled and 3d are in progress.

p.s.: this was forked from `[improve-automatic-segmentation](https://github.com/computational-cell-analytics/micro-sam/tree/improve-automatic-segmentation)` at some point, which has now been deleted, but it's unclear where it got merged into `dev` exactly.


constantinpape:
CODEX found some big optimzation knobs:

image

I am looking into this a bit further.

constantinpape:
2D optimization is done, see final results below --- it will need some clean up in the end though. I am launching the 3d campaign, which should result in similar gains by learning selection and scoring rather than relying on the IOU prediction.

image

constantinpape:
This is done for now from my end: the 2d optimization turned out nicely, but transferring to 3d was more difficult then expected so I decided to keep the 3d version as is.

It makes sense to merge after you review @anwai98 ; we also need to resolve the merge conflicts.

anwai98:
> This is done for now from my end: the 2d optimization turned out nicely, but transferring to 3d was more difficult then expected so I decided to keep the 3d version as is.

Okie. Thanks Constantin!

> It makes sense to merge after you review @anwai98 ; we also need to resolve the merge conflicts.

Yep, I'll rebase it to make the diff realistic and short, solve the merge conflicts and review it only then.

I'll do it all after lunch. Thanks for this! 😁

anwai98 (approved):
This looks in a great shape I'd say. There are a few more integration tests I'd do, but for now I'll go ahead and merge this right away! (will plug a few more things later)

PR #1347, opened 2026-08-23
Invalidate cached image embeddings when forward() receives a different image

nilsmechtel:
The `PredictorAdaptor` computes image embeddings on the first prompt-free `forward()` call and then never recomputes them: `is_image_set` stays true, so every later call with a *different* image silently reuses the first image's embeddings and returns masks in the first image's geometry. The bug is masked whenever the `embeddings` input is passed (that branch overwrites the cache on every call), which is why `bioimageio.core.test_model` does not catch it for exports that declare the prompt/embeddings inputs. It does affect any consumer that keeps one model instance alive across images, e.g. `bioimageio.core` prediction pipelines and model servers, and it makes prompt-free (AIS-only) exports fail `test_model`'s parametrized-size probes, which reuse a single pipeline for the n=0, 1, 2 input sizes.

This change keys the embedding cache on the actual input: the adaptor records the image tensor the current embeddings belong to and reuses them only when the incoming image is equal to it. Passing `embeddings` explicitly still overrides the cache as before, and repeated calls with the same image still skip recomputation, so interactive prompt iteration on one image is unaffected.

Validated on the six bioimage.io zoo SAM models: `bioimageio.core.test_model` passes with the patched adaptor for image-only (AIS) packages, including the parametrized-size probes at three different input sizes, both locally and in clean conda environments created from the exported `environment.yaml`. Sequential `forward()` calls with different image sizes now return correctly sized outputs where they previously returned the first call's geometry.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/micro_sam/bioimageio/predictor_adaptor.py
+++ b/micro_sam/bioimageio/predictor_adaptor.py
@@ -39,6 +39,10 @@ class PredictorAdaptor(nn.Module):
         self.sam_model = sam_model_registry[model_type]()
         self.sam = SamPredictor(self.sam_model)
         self.decoder = None
+        # The image the current self.sam embeddings were computed for (or that
+        # externally passed embeddings were declared for). Used to invalidate
+        # the embedding cache when forward() is called with a different image.
+        self._input_image = None
 
     def load_state_dict(self, state, **kwargs):
         # Finetuning checkpoints store SAM and decoder weights separately.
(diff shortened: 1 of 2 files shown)


anwai98 (approved):
Nice, thanks for the PR fix @nilsmechtel!

I'll go ahead and merge this once the CIs pass!

PS. I switched the cache to SAM's resized inputs so that (with a corner case found by Codex) larger images don't leave full-size copies sitting in memory with several `clone()`s around

PR #1348, opened 2026-08-23
Test exported decoder models via automatic instance segmentation

nilsmechtel:
Builds on #1347 (the first commit here is that PR's cache fix; this change needs it and should be merged after it).

`export_sam_model` currently declares the image plus five optional prompt/embeddings inputs, each with a test tensor. The bioimageio spec feeds every declared test tensor during testing, so `bioimageio.core.test_model` always exercises the promptable head and the automatic instance segmentation path of decoder exports is never tested. On top of that, an image-only self-test cannot pass with the current declarations: the parametrized-size probes resize the test image to the declared y/x minimum of 1 pixel, where AIS finds no objects and violates the mask output's data dependent minimum of one object.

With this change, models exported **with** an instance segmentation decoder declare only the image input, and the test procedure validates AIS end to end, including the image encoder:

- The reference outputs (`mask.npy`, `scores.npy`, `embeddings.npy`) are generated from a prompt-free call, so "reproduce test outputs" runs automatic instance segmentation. An export whose test image yields zero instances fails immediately with a clear error instead of producing a package that fails downstream.
- The image y/x axes use `ParameterizedSize(min=256, step=1)` for decoder exports. Below roughly 200 pixels the resized probe images no longer contain segmentable objects (measured across the six bioimage.io zoo models: 0-1 instances below 192 px, 3-10 for the EM models and 46-61 for the LM models at 256-258 px). Every size of at least 256 remains valid.
- Reference generation runs in eval mode: `bioimageio.core` executes exported models in eval mode, and in train mode the batch norm layers of the image encoder use batch statistics, which shifts AIS results by whole instances. The prompted reference outputs of existing exports were also generated in train mode, but thresholded prompted masks happened to tolerate the difference.
- The reference outputs are regenerated through a prediction pipeline (same call sequence and device selection as the test procedure) before packaging, and the model description declares a reproducibility tolerance of 1000 mismatched elements per million for the embeddings output. Without these, per-element float drift between the direct call and the pipeline, and between different hardware, fails the strict reproducibility check (measured 255 ppm between CPU and CUDA for vit_t_lm) while having no effect on the segmentation outputs.
- Cover images for decoder exports show the outline of all AIS instances instead of a box prompt with a single mask.

Models exported **without** a decoder are unchanged: they keep the full prompt interface, the arbitrary image size, and the existing prompted test procedure. The promptable head of decoder exports also remains fully functional through micro_sam itself; it is only no longer part of the RDF interface.

Validated with the updated unit tests (the with-decoder test now asserts the image-only interface and the raised size minimum) and with a full export of vit_t_lm on a real microscopy image followed by `bioimageio.core.test_model`, which passes on CUDA and on CPU, including the parametrized probes at 256, 257 and 258 px. The same image-only design is already staged on the six micro-sam models on bioimage.io and passes the BioEngine test runner there.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


anwai98 (approved):
Thanks for another PR @nilsmechtel 🥳 

PS. macOS CIs weren't happy because the synthetic images generated weird stuff everytime, and BioimageIO was not happy about it. I patched it with a small example image! (hopefully the CIs now pass hehe)

I'll merge this very soon!

anwai98:
Oh also btw two things: 1) give this merged PR a quick local try and lemme know if all works good, otherwise feel free to open another patch PR, and 2) once all looks good, lemme know if you'd like a new release -- I'll spawn one anytime!

Thanks once again, @nilsmechtel 🥳

anwai98:
Okay this works now. I'll go ahead and merge this!

nilsmechtel:
@anwai98 thanks for fixing the CIs! 🙏🏼 I also tested it in BioEngine and the model passes 🥳

A new release on top of the merged PR would be really great, thank you!

anwai98:
Awesome, making one right away. Thanks for confirming! 😁

PR #1353, opened 2026-08-26
Wire per-model-type defaults for AIS and APG postprocessing parameters

anwai98:
Gonna quickly test this before I consider merging it.


anwai98:
This is in a good shape. I know why the CIs are broken and will fix them right away in a few minutes!

anwai98:
Nice, this is ready to be merged once the CIs pass!

PR #1357, opened 2026-08-27
Enable IB comms for multi-node joint training

anwai98:
Okay IB comms works fine now (poof). The weird mixed precision engineered tactics I was using was fixed by me in `torch-em` already haha, so this worked as smooth as butter!

Gonna make a super quick test and then merge this right away!


anwai98:
This shouldn't break any CIs. I'll go ahead and merge this right away!

PR #1358, opened 2026-08-28
Add candidate pruning to tiled 3d APG propagation

anwai98:
Aimed towards making 3d APG with tiling faster!


anwai98:
This is like very very fast right now, and ironically gives a slight bump in scores too ;) I love it. I'll do a quick test on the pruning schema, benchmark it over 1 and 4 GPUs each, and then merge it once I'm like happy happy hehe!

PR #1361, opened 2026-09-01
Fix segment anything installation from pip

### LINKED ISSUE #1360: Installing micro-sam with pip silently squashes images out of shape before training
If you install micro-sam with pip rather than conda, it quietly picks up a 2023 copy of the Segment Anything library that carries a known geometry bug. Every image handed to the trainer is squashed to the wrong aspect ratio before the model sees it — a square 512×512 patch is stretched into a 341×1024 sliver, then padded. Nothing raises an error: training loss still descends, checkpoints still save, and the bioimageio export self-test still passes. The model just learns from distorted pictures.

Installs from conda-forge are unaffected, because that build takes the library from source, where the bug was fixed nine days after the broken copy was published. The broken copy is the only one ever released to PyPI, so every pip install of micro-sam gets it.

I found this while running micro-SAM automatic-instance-segmentation models through `bioimageio.core` in a pip-built environment: 56 instances instead of 67, with no error anywhere. Localising it took a per-stage numerical diff between a conda environment and a pip one.

---

## The defect

`segment_anything/utils/transforms.py` in the PyPI `1.0` wheel:

```python
# line 30 — apply_image, numpy HWC: [0], [1] really are height, width   correct
target_size = self.get_preprocess_shape(image.shape[0], image.shape[1], self.target_length)

# line 62 — apply_image_torch, torch BCHW: [0], [1] are batch, channels  WRONG
target_size = self.get_preprocess_shape(image.shape[0], image.shape[1], self.target_length)
```

Line 62 should read `image.shape[2], image.shape[3]`. Upstream fixed exactly this in [`567662b`](https://github.com/facebookresearch/segment-anything/commit/567662b0fd33ca4b022d94d3b8de896628cd32dd), *"Fix incorrect shape in ResizeLongestSide.apply_image_torch."*

`get_preprocess_shape` scales by the longest of the two values it is given, so `get_preprocess_shape(1, 3, 1024)` yields `scale = 1024/3`: a `(1, 3, 520, 704)` input comes back as `(1, 3, 341, 1024)` — 341/1024 = 1/3, the channels-to-batch ratio — and is then padded to 1024×1024.

## Why pinning is the only route

| | |
|---|---|
| PyPI `segment-anything==1.0` uploaded | 2023-04-06 18:04 UTC |
| upstream fix commit `567662b` | 2023-04-15 |
| releases on PyPI, ever | 1 |
| last code push to `facebookresearch/segment-anything` | 2024-09-18 |
| PyPI project owner | `JoOkuma` (not Meta) |

The wheel predates its own fix by nine days and was never re-cut, and the PyPI project is not owned by facebookresearch — so waiting for a corrected release is not a realistic path. Any pip resolution of the bare name `segment-anything` gets the broken copy, and micro-sam's `requires_dist` lists it bare and unpinned.

## Where micro-sam calls the broken method

`ResizeLongestSide` is imported in four modules; `apply_image_torch` is called in three (paths from 1.6.2, unchanged in 1.8.x):

- `micro_sam/training/trainable_sam.py:36` — on **every training forward pass**, immediately before normalize + pad
- `micro_sam/bioimageio/predictor_adaptor.py:74` and `:79`
- `micro_sam/prompt_based_segmentation.py:103,109` (on logits)

The training call is the consequential one. I have verified the mechanism and the dependency resolution, **not** a training-quality regression, so I make no claim about a specific accuracy impact.

## Reproduce

```bash
pip install micro-sam
python -c "import inspect, segment_anything.utils.transforms as t; print(inspect.getsource(t.ResizeLongestSide.apply_image_torch))"
```

If the body reads `image.shape[0], image.shape[1]`, the install is affected. The same command in a conda-forge `micro-sam` env prints `image.shape[2], image.shape[3]`.

A direct check of the consequence:

```python
import torch
from segment_anything.utils.transforms import ResizeLongestSide
print(ResizeLongestSide(1024).apply_image_torch(torch.zeros(1, 3, 520, 704)).shape)
# broken: torch.Size([1, 3, 341, 1024])
# fixed:  torch.Size([1, 3, 756, 1024])
```

## Suggested fix

Pin the upstream commit as a direct reference, so pip cannot resolve the stale wheel:

```
segment-anything @ git+https://github.com/facebookresearch/segment-anything.git@dca509fe793f601edb92606367a655c15ac00fdf
```

A direct URL in published metadata does prevent the package itself from being uploaded to PyPI, so this may need to be an install-doc instruction rather than a `requires_dist` entry. Two alternatives if so:

1. an import-time assertion that `apply_image_torch` handles BCHW correctly, failing loudly with the git-pin instruction rather than silently mis-resizing;
2. documenting pip installs as unsupported, with the git pin stated explicitly in the install instructions.

Happy to open a PR for whichever you prefer.

## Environment

micro-sam 1.6.2 (conda-forge, correct) and 1.8.11 (pip, affected). Verified across four environments on one machine: three pip-built envs all show the broken line 62; the single conda-forge env is correct.

anwai98:
Okay over `micro_sam-feedstock` -- we use the `segment-anything` package from this feedstock: https://github.com/conda-forge/segment-anything-feedstock.

And the same author has pypi wheels at https://pypi.org/project/segment-anything-py/. And they've had the parallel versions of adopting the fixes around. So I think it's better to go with theirs, knowing the issue reported by @nilsmechtel (thanks once again for reporting 🥳)

I'll go ahead, merge this and then make a new release to fix it!

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -55,7 +55,7 @@ install_requires =
     python-elf>=0.9
     scikit-image
     scikit-learn
-    segment-anything
+    segment-anything-py>=1.0.1
     superqt
     timm
     torch>=2.5
--- a/test/test_dependencies.py
+++ b/test/test_dependencies.py
@@ -1,0 +1,15 @@
+import unittest
+
+import torch
+
+from segment_anything.utils.transforms import ResizeLongestSide
+
+
+class TestDependencies(unittest.TestCase):
+    def test_resize_longest_side_torch(self):
+        transform = ResizeLongestSide(16)
+        image = torch.zeros((1, 3, 5, 7))
+
+        resized = transform.apply_image_torch(image)
+
+        self.assertEqual(tuple(resized.shape), (1, 3, 11, 16))


anwai98:
cc: @constantinpape

PR #1364, opened 2026-09-02
Improve joint training speed

anwai98:
This increases the training speed and finally makes the best of IB comms speed at it's best!

We are down to ca 1.5s/it from ca. >2.5s/it over multiple nodes at scale!

I'm gonna keep this open since it needs a new release for changes from here: https://github.com/constantinpape/torch-em/pull/777. 

The CIs might break, but let's not panic leute bitte!


anwai98:
If the CIs now work too, I'm gonna merge this right away!

PR #1365, opened 2026-09-04
Exclude the Platynereis neuropil from the loss with an ignore label

anwai98:
Tagged with https://github.com/constantinpape/torch-em/pull/796. What do you think @constantinpape? Could you please give it a quick review?

### DIFF
--- a/micro_sam/v2/transforms/labels.py
+++ b/micro_sam/v2/transforms/labels.py
@@ -9,6 +9,11 @@ from bioimage_cpp.distance import distance_transform, geodesic_distance_field, v
 from bioimage_cpp.segmentation import label as connected_components, relabel_sequential
 
 
+# Sentinel in the foreground channel for voxels with unknown ground truth, so the loss can skip them.
+# The target reaches the loss as float32, where the uint32 label-space sentinel is not representable.
+IGNORE_FOREGROUND = 255
+
+
 def _instance_labels(labels):
     """Relabel each connected region as a unique integer instance.
 
@@ -31,7 +36,9 @@ def _axondeepseg_pre_label_transform(y):
     return connected_components(y == 2).astype("uint32")
 
 
-def _em_cell_label_trafo(y, label_trafo):
+def _em_cell_label_trafo(y, label_trafo, ignore_label=None):
+    # Take the ignore mask before label_trafo, which replaces the instance ids with distances.
+    ignore = None if ignore_label is None else np.asarray(y) == ignore_label
     y = label_trafo(y)
 
     # Prepare the true background.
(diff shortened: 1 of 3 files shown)


anwai98:
The CIs are broken because we'd  need a new `torch-em` with the new changes. Let's not worry about those details for now!

anwai98:
Any thoughts on this? @constantinpape

micro_sam/v2/transforms/labels.py
```diff
@@ -9,6 +9,11 @@
 from bioimage_cpp.segmentation import label as connected_components, relabel_sequential
 
 
+# Sentinel in the foreground channel for voxels with unknown ground truth, so the loss can skip them.
+# The target reaches the loss as float32, where the uint32 label-space sentinel is not representable.
+IGNORE_FOREGROUND = 255
```
constantinpape: Really 255? That seems dangerous, we have no guarantee that it doesn't conflict with a label in the volume; 255 distinct IDs in a 3d volume is very likely for some of our datasets. I'd rather go for a large integer that is still representable in float 32. We could also choose 65536 (2 ** 16).

constantinpape (commented):
Idea is good, but I find 255 very dangerous as ignore id.

micro_sam/v2/transforms/labels.py
anwai98: I didn't put it to test but I think your'e right. Larger integer would be a better choice. Lemme check upon this -- we don't wanna hurt densely annotated structures!

anwai98: Sorted!

micro_sam/v2/transforms/labels.py
```diff
@@ -9,6 +9,10 @@
 from bioimage_cpp.segmentation import label as connected_components, relabel_sequential
 
 
+# Written into the foreground target channel, which otherwise holds 0 or 1, for voxels without ground truth.
+FOREGROUND_IGNORE_VALUE = -1
```
anwai98, line 13: Now the ignore value lives only in the foreground target channel: we find the `ignore_label` voxels (needs to be provided by the user) and map them to -1 there, and the loss skips them from then on.

constantinpape: That makes sense.

anwai98: Neeeeat. Thanks! 😁

anwai98:
Okay this is gonna stay open because I need a few more moving parts figured out in `torch-em`. Then I'll make a new release there, merge this, and then we are good to fly. Thanks @constantinpape!

anwai98:
Gonna merge this. The CIs would break, but I don't wanna wait for `torch-em` release (because I'm still cross checking all datasets one final time)

PR #1368, opened 2026-09-06
Apg clean up

constantinpape:
This PR cleans up a lot of APG methodology that turned out to not bring any advantages. @anwai98 could you quickly (let it be) review(ed) to check if nothing critical gets removed?

Also contains the extensive notes on the optimization campaigns run and refuted.


constantinpape:
(and we should def. squash this when merging)

anwai98 (approved):

constantinpape:
Thanks for the review. I think the test failures are now all because torch-em is not up-to-date. Let's wait till this is resolved to make sure these changes don't break anything.

anwai98:
This should be sorted (new releasing coming soon for `torch-em`: https://github.com/conda-forge/torch_em-feedstock/pull/41)

PR #1382, opened 2026-09-21
Some minor training routine updates

anwai98:


anwai98:
This needs to stay open until the last batch of trainings are sorted! (which are ongoing as we speak ;))

anwai98:
cc: @caroteu

